PR #19, opened 2020-09-03
fix go-mod

henderiw:
go-mod was wrongly initialised


henderiw:
approved

henderiw:
approved confirmed

PR #29, opened 2020-09-04
deployCmd flags

karimra:
Added deploy command flags:
`--prefix | p`: overwrites the Prefix var in topo file
`--bridge | b`: overwrites the DockerInfo.bridge var in topo file
`--ipv4-subnet | 4`: overwrites the DockerInfo.ipv4_subnet var in topo file
`--ipv6-subnet | 6`: overwrites the DockerInfo.ipv6_subnet var in topo file

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -7,7 +7,7 @@ import (
 // var debug bool
 
 type cLab struct {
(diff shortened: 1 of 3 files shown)


henderiw:
approved

PR #30, opened 2020-09-04
ixrd2 srl type typo

steiler:
ixrd2 type had a typo in the key. 
This was not visible in the panic output because it was not generated from the map, which is changed in this commit.


hellt:
Thanks @steiler 

do you mind

1. changing the variable to camelCase instead of snake_case 
2. and making the slice for keys created with a known lenght like `keys := make([]string, 0, len(srl_types)`

henderiw (approved):

PR #32, opened 2020-09-04
wrong filename; README syntax highlight

steiler:
The config.tpl needed to be renamed to srlconfig.tpl.
README.md yaml config documentation enriched with code marks for better readability.


henderiw:
approved

PR #37, opened 2020-09-07
Exec command

karimra:
added exec command,
it can be used to run a command on multiple containers of a fabric, filtered by labels, then outputs stdout and stderr on the terminal
E.g:
`containerlab exec --prefix wan --label kind=srl "sr_cli -r show interface"` 
runs cmd="sr_cli -r show interface" on all srl containers in fabric having prefix wan

`containerlab exec --prefix wan --label kind=srl --label group=spines "sr_cli -r show interface"` 
runs cmd="sr_cli -r show interface" on all srl containers belonging to "group spines" in fabric having prefix wan

`containerlab exec --prefix wan --label kind=alpine ip addr show` 
runs cmd="ip addr show" on all alpine containers in fabric having prefix wan

notes: 
- if the command contains flags (starting with - or --) the whole command has to be between quotes
- the command output is printed once the command execution finishes, so it's not currently possible to run commands that require stdin input to terminate (like ping without a -c)

### DIFF
--- a/clab/docker.go
+++ b/clab/docker.go
@@ -288,6 +288,7 @@ func (c *cLab) Exec(ctx context.Context, id string, cmd []string) ([]byte, []byt
 	})
 	if err != nil {
 		log.Errorf("failed to create exec in container %s: %v", cont.Name, err)
+		return nil, nil, err
 	}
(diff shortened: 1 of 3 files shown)


henderiw:
approved

PR #43, opened 2020-09-09
Using go-graphviz to generate not just the .dot but also a png file.

steiler:
Also fixing, that a topology would have to be deployed before a graph could be created, since the folder structure did not already exist.


henderiw:
approved

PR #44, opened 2020-09-09
link-labeling

henderiw:
added labeling to links in the config file for the links such that later on the configuration can become easier


henderiw:
approved

PR #70, opened 2020-09-29
pull docker images which are not already present

steiler:
Docker images are not pulled by default. If e.g. the henderiw/client-alpine:1.0.0 image is not present the deploy task will fail.
This branch will try downloading the container images which are missing on the system.
This would later on also help when SRL is available via a registry. But if SRL is now already loaded everything is fine.

### DIFF
--- a/clab/docker.go
+++ b/clab/docker.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -144,6 +146,12 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
 	if node.Group != "" {
 		labels["group"] = node.Group
 	}
+
+	err = c.PullImageIfRequired(nctx, node)
+	if err != nil {
+		return err
+	}
+
 	cont, err := c.DockerClient.ContainerCreate(nctx,
 		&container.Config{
 			Image:        node.Image,
@@ -177,6 +185,55 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
 		return err
 	}
 	return createContainerNS(node.Pid, node.LongName)
+}
+
+func (c *cLab) PullImageIfRequired(ctx context.Context, node *Node) (err error) {
+	filter := filters.NewArgs()
+	filter.Add("reference", node.Image)
+
+	ilo := types.ImageListOptions{
+		All:     false,
+		Filters: filter,
+	}
+
+	log.Debugf("Looking up %s Docker image", node.Image)
+
+	images, err := c.DockerClient.ImageList(ctx, ilo)
+	if err != nil {
+		return err
+	}
+
+	// If Image doesn't exist, we need to pull it
+	if len(images) > 0 {
+		log.Debugf("Image %s present, skip pulling", node.Image)
+		return nil
+	}
+
+	// might need canonical name e.g.
+	//    -> alpine == docker.io/library/alpine
+	//    -> foo/bar == docker.io/foo/bar
+	//    -> docker.elastic.co/elasticsearch/elasticsearch == docker.elastic.co/elasticsearch/elasticsearch
+	canonicalImageName := node.Image
+	slashCount := strings.Count(node.Image, "/")
+
+	switch slashCount {
+	case 0:
+		canonicalImageName = "docker.io/library/" + node.Image
+	case 1:
+		canonicalImageName = "docker.io/" + node.Image
+	}
+
+	log.Infof("Pulling %s Docker image", node.Image)
+	reader, err := c.DockerClient.ImagePull(ctx, canonicalImageName, types.ImagePullOptions{})
+	if err != nil {
+		return err
+	}
+	defer reader.Close()
+	// must read from reader, otherwise image is not properly pulled
+	io.Copy(ioutil.Discard, reader)
+	log.Infof("Done pulling %s", canonicalImageName)
+
+	return nil
 }
 
 // StartContainer starts a docker container
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -5,7 +5,7 @@ import (
 	"sync"
 	"time"
 
-	docker "github.com/docker/engine/client"
+	docker "github.com/docker/docker/client"
 )
 
 // var debug bool
--- a/docs/install.md
+++ b/docs/install.md
@@ -3,7 +3,7 @@ Containerlab is distributed as Linux package and can be installed on any Debian-
 ### Pre-requisites
 * Have `sudo` rights on the system: `containerlab` sets some parameters in the linux system to support the various options the containers need
 * [Install docker](https://docs.docker.com/engine/install/): this is used to manage the containers
-* Pull the container images which you intend to run to the Linux system
+* Import or Pull container images (e.g. Nokia SR-Linux, Arista cEOS) which are not online available for download from a registry. Containerlab will try to pull images if they do not exist on execution time.
 
 ### Package installation
 Containerlab package can be installed using the [installer script](https://github.com/srl-wim/container-lab/blob/master/get.sh) which detects the operating system type and installs the relevant package:


clab/docker.go
```diff
@@ -179,6 +187,45 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+		} else if strings.Count(node.Image, "/") == 1 {
+			canonicalImageName = "docker.io/" + node.Image
+		}
+		log.Infof("Pulling %s Docker image", node.Image)
+		reader, err := c.DockerClient.ImagePull(ctx, canonicalImageName, types.ImagePullOptions{})
+		// must read from reader, otherwise image is not properly pulled
+		io.Copy(ioutil.Discard, reader)
+		log.Infof("Done pulling %s", canonicalImageName)
+		if err != nil {
+			panic(err)
```
hellt, line 221: lets return an error here instead of panicing

clab/docker.go
```diff
@@ -179,6 +187,45 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+	images, err := c.DockerClient.ImageList(ctx, ilo)
+	_ = images
+	// Image doesn't exist, need to pull it
+	if len(images) <= 0 {
+		// might need canonical name e.g.
+		//    -> alpine == docker.io/library/alpine
+		//    -> foo/bar == docker.io/foo/bar
+		//    -> docker.elastic.co/elasticsearch/elasticsearch == docker.elastic.co/elasticsearch/elasticsearch
+		canonicalImageName := node.Image
+		if strings.Count(node.Image, "/") == 0 {
```
hellt, line 210: lets rewrite `if/else if` to a more canonical `switch/case`?

clab/docker.go
```diff
@@ -179,6 +187,45 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+
+	ilo := types.ImageListOptions{
+		All:     false,
+		Filters: filter,
+	}
+
+	log.Debugf("Looking up %s Docker image", node.Image)
+
+	images, err := c.DockerClient.ImageList(ctx, ilo)
+	_ = images
```
hellt, line 202: why is this needed?

clab/docker.go
```diff
@@ -179,6 +187,45 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+	// Image doesn't exist, need to pull it
+	if len(images) <= 0 {
```
hellt, line 204: can length over images return a negative?

hellt (commented):
thanks @steiler 
I have a few questions/suggestions

clab/docker.go
steiler: will do

clab/docker.go
steiler: will do

clab/docker.go
steiler: L202 is not needed any more, you're right I'm now using that value.

clab/docker.go
steiler: will adjust

clab/docker.go
```diff
@@ -179,6 +187,52 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+		//    -> docker.elastic.co/elasticsearch/elasticsearch == docker.elastic.co/elasticsearch/elasticsearch
+		canonicalImageName := node.Image
+		slashCount := strings.Count(node.Image, "/")
+
+		switch slashCount {
+		case 0:
+			canonicalImageName = "docker.io/library/" + node.Image
+		case 1:
+			canonicalImageName = "docker.io/" + node.Image
+		}
```
hellt, line 220: when I looked into SDK examples I saw that they don't use the canonical name https://docs.docker.com/engine/api/sdk/examples/#pull-an-image

can you test if that is needed? My assumption is that if you specify simply `alpine` it will still do the pull

clab/docker.go
```diff
@@ -179,6 +187,52 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+
+	log.Debugf("Looking up %s Docker image", node.Image)
+
+	images, err := c.DockerClient.ImageList(ctx, ilo)
+	if err != nil {
+		return err
+	}
+
+	// Image doesn't exist, need to pull it
+	if len(images) == 0 {
```
hellt, line 207: lets keep the "happy path" on the first indentation level, meaning that the structure of the code should be revered like that:

```go
if len(images) > 0:
    log.Debugf("Image %s present, skip pulling", node.Image)
    return nil

// here goes the pulling logic on the same indentation level
```

clab/docker.go
```diff
@@ -179,6 +187,52 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+
+		switch slashCount {
+		case 0:
+			canonicalImageName = "docker.io/library/" + node.Image
+		case 1:
+			canonicalImageName = "docker.io/" + node.Image
+		}
+
+		log.Infof("Pulling %s Docker image", node.Image)
+		reader, err := c.DockerClient.ImagePull(ctx, canonicalImageName, types.ImagePullOptions{})
```
hellt, line 223: add `defer reader.Close()` after you created a readcloser on L223

clab/docker.go
```diff
@@ -179,6 +187,52 @@ func (c *cLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
+		// must read from reader, otherwise image is not properly pulled
+		io.Copy(ioutil.Discard, reader)
```
hellt, line 225: I also think maybe you should read to `os.Stdout`? I think that way you will be able to show the progress of image pulling to a user during the lab instantiation

hellt:
@steiler to finish this off, can you change this line https://github.com/srl-wim/container-lab/blame/master/docs/install.md#L6 to mention that only images which can't be pulled from public registries have to be pulled upfront?

clab/docker.go
steiler: I did and I was shocked myself.

```
Start - alpine
repository name must be canonical
Done - alpine

Start - library/alpine
repository name must be canonical
Done - library/alpine

Start - docker.io/library/alpine
{"status":"Pulling from library/alpine","id":"latest"}
...
{"status":"Pull complete","progressDetail":{},"id":"df20fa9351a1"}
Done - docker.io/library/alpine
```

clab/docker.go
steiler: I dislike this approach.
The happy path in my view is when the last statement of a function is reached without any early exiting via any return other then the last line of a function... (exceptions exist of course)
However I'll change order and have the image exists atop.

clab/docker.go
steiler: Thanks, will do!

clab/docker.go
hellt: Interesting, I see that we use https://github.com/srl-wim/container-lab/blob/f11586e6916cb0bf0600633073e1bb001a6f4e4b/clab/clab.go#L8 as a docker api package, although this is a fork from `moby/moby`
the official Go docs actually say to use the `github.com/docker/docker/client` package

I wonder, can you swap it and give it a try? Maybe the upstream Go SDK is a more polished version that we should rather use?

/cc @karimra

clab/docker.go
steiler: Agree, but after having seen the output I've decided to discard.

And that's just the output for the tiny alpine image...
```
Start -  docker.io/library/alpine
{"status":"Pulling from library/alpine","id":"latest"}

{"status":"Pulling fs layer","progressDetail":{},"id":"df20fa9351a1"}

{"status":"Downloading","progressDetail":{"current":29170,"total":2797541},"progress":"[\u003e                                                  ]  29.17kB/2.798MB","id":"df20fa9351a1"}

{"status":"Downloading","progressDetail":{"current":425216,"total":2797541},"progress":"[=======\u003e                                           ]  425.2kB/2.798MB","id":"df20fa9351a1"}

{"status":"Downloading","progressDetail":{"current":781568,"total":2797541},"progress":"[=============\u003e                                     ]  781.6kB/2.798MB","id":"df20fa9351a1"}

{"status":"Downloading","progressDetail":{"current":1207552,"total":2797541},"progress":"[=====================\u003e                             ]  1.208MB/2.798MB","id":"df20fa9351a1"}

{"status":"Downloading","progressDetail":{"current":1604061,"total":2797541},"progress":"[============================\u003e                      ]  1.604MB/2.798MB","id":"df20fa9351a1"}

{"status":"Downloading","progressDetail":{"current":2071005,"total":2797541},"progress":"[=====================================\u003e             ]  2.071MB/2.798MB","id":"df20fa9351a1"}

{"status":"Downloading","progressDetail":{"current":2464221,"total":2797541},"progress":"[============================================\u003e      ]  2.464MB/2.798MB","id":"df20fa9351a1"}

{"status":"Verifying Checksum","progressDetail":{},"id":"df20fa9351a1"}

{"status":"Download complete","progressDetail":{},"id":"df20fa9351a1"}

{"status":"Extracting","progressDetail":{"current":32768,"total":2797541},"progress":"[\u003e                                                  ]  32.77kB/2.798MB","id":"df20fa9351a1"}

{"status":"Extracting","progressDetail":{"current":2064384,"total":2797541},"progress":"[====================================\u003e              ]  2.064MB/2.798MB","id":"df20fa9351a1"}

{"status":"Extracting","progressDetail":{"current":2797541,"total":2797541},"progress":"[==================================================\u003e]  2.798MB/2.798MB","id":"df20fa9351a1"}

{"status":"Pull complete","progressDetail":{},"id":"df20fa9351a1"}

{"status":"Digest: sha256:185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321"}

{"status":"Status: Downloaded newer image for alpine:latest"}

Done -  docker.io/library/alpine
```

clab/docker.go
hellt: that is what is considered an idiomatic Go (left aligned happy path with early returns), so its good to adhere to what community adopted, that will make code look consistent

clab/docker.go
hellt: valid point. especially considering that we will pull in parallel quite soon.

clab/docker.go
karimra: This was changed because of this issue #23

hellt: yes, I remember #23, it just maybe it was not the right pointer =)

clab/docker.go
karimra: Talking about idiomatic go: https://go-proverbs.github.io/
Last one is "don't panic" includes both panic statement and log.Fatalx

steiler: > that is what is considered an idiomatic Go (left aligned happy path with early returns), so its good to adhere to what community adopted, that will make code look consistent

Well if that's the case ... who am I to complain, right?
It's changed!

clab/docker.go
steiler: The official `github.com/docker/docker/client` works fine as well and doesn't break anything, but the canonical name logic is still required.

hellt:
lgtm

PR #71, opened 2020-09-29
changed sysctl modifications via writing to a file

### LINKED ISSUE #52: duplicated sysctl command
The sysctl commands https://github.com/srl-wim/container-lab/blob/master/clab/docker.go#L99 and  https://github.com/srl-wim/container-lab/blob/master/clab/docker.go#L108 are identical, shouldn't the second one be: 
`b, err = exec.Command("sudo", "sysctl", "-w", "net.ipv4.conf.default.rp_filter=0").CombinedOutput()` ?

hellt:
This PR changes the way we interact with sysctl by swapping subshelling to bash with a write to a respective file

Closes #52 

Additionally, the same approach is used to add LLDP capability to a linux bridge.

### DIFF
--- a/clab/docker.go
+++ b/clab/docker.go
@@ -87,7 +87,11 @@ func (c *cLab) CreateBridge(ctx context.Context) (err error) {
 	log.Debug("Disable RPF check on the docker host")
 	err = setSysctl("net/ipv4/conf/all/rp_filter", 0)
 	if err != nil {
-		return fmt.Errorf("failed to disable RP filter on docker host: %v", err)
+		return fmt.Errorf("failed to disable RP filter on docker host for the 'all' scope: %v", err)
+	}
+	err = setSysctl("net/ipv4/conf/default/rp_filter", 0)
+	if err != nil {
+		return fmt.Errorf("failed to disable RP filter on docker host for the 'default' scope: %v", err)
 	}
 
 	log.Debugf("Enable LLDP on the docker bridge %s", bridgeName)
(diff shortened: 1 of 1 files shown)


clab/docker.go
```diff
@@ -79,33 +83,23 @@ func (c *cLab) CreateBridge(ctx context.Context) (err error) {
...
-	var b []byte
-	b, err = exec.Command("sudo", "sysctl", "-w", "net.ipv4.conf.all.rp_filter=0").CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("failed to disable Checksum Offloading on docker bridge: %v", err)
-	}
-	log.Debugf("%s", string(b))
-	//if err = sysctl.Set("net.ipv4.conf.default.rp_filter", "0"); err != nil {
-	//	return err
-	//}
-	log.Debug("Disable RPF check on the docker host part2")
```
hellt, line 92: @henderiw what was the reason to disable RPF check twice (part1 and part2)? I removed the part2 as it seems unnecessary, unless it is needed for the reasons unknown to me.

karimra: check #52

hellt: Ah, sorry, my bad, I didn't notice it had different scopes in the commands (ipv4 vs default)

Both are needed to disable rpf on existing and new interfaces. I will adjust the code accordingly

hellt: fixed in 7a6585e

hellt:
tested locally

PR #72, opened 2020-09-30
added lldp config for def srl config template

hellt:
This PR enables LLDP globally for SRL nodes.


hellt:
@karimra am I right that we don't use `templates/config.json` and `templates/config.yml.tpl` anymore?
it seems its safe to delete them

karimra:
yes, they can be deleted.

PR #77, opened 2020-09-30
Use netlink and netns package for interfaces/netns deletion

hellt:
This PR removes some dependancies on iproute2 util in favor of netns and netlink packages that are used by container engines.

### DIFF
--- a/clab/netlink.go
+++ b/clab/netlink.go
@@ -3,6 +3,7 @@ package clab
 import (
 	"fmt"
 	"net"
+	"os"
 	"os/exec"
 	"strings"
 	"sync"
@@ -10,6 +11,7 @@ import (
 	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
 )
 
 func (c *cLab) InitVirtualWiring() {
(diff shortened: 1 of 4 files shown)


hellt:
The same set of packages can be used in the wiring functions that currently rely on iproute2

clab/netlink.go
```diff
@@ -150,28 +155,16 @@ func (c *cLab) createvethToBridge(l *Link) error {
 	return nil
 }
 
-// DeleteVirtualWiring deletes the virtual wiring
+// DeleteVirtualWiring deletes the virtual wiring by deleting network namespaces
 func (c *cLab) DeleteVirtualWiring(id int, link *Link) (err error) {
```
karimra, line 159: I think we can get rid of the `id int` in this method signature, it's not used anywhere.

karimra: in case one of the endpoints is a bridge, shouldn't we delete the interface that was added under that bridge? since containerlab created it during deploy

karimra (changes requested):

clab/netlink.go
hellt: removed `id` in 3f29ac3

for deletion of the interfaces of a bridge I need to check additionally, likely in a separate PR, as this functionality is not currently there

karimra: Lgtm

PR #79, opened 2020-10-03
removed prefixing slash from containers name returned by "inspect" cmd

hellt:
Docker inspect cmd uses FQDN for container names which start with "/"
unfortunately, this is inconsistent with the rest of the CLI commands, which take a non-slashed names as their arguments.

In this PR I removed the "/" from the container name for the `containerlab inspect` command:
```
❯ ../container-lab/container-lab inspect -t 1_srl.yml                                                                                                                                                                                                                      ─╯
INFO[0000] Getting topology information ...             
+------------------------+--------------------+------+-------+---------+----------------+----------------------+
|          Name          |       Image        | Kind | Group |  State  |  IPv4 Address  |     IPv6 Address     |
+------------------------+--------------------+------+-------+---------+----------------+----------------------+
| containerlab-1srl-srl1 | srlinux:20.6.1-286 | srl  |       | running | 172.19.20.2/24 | 2001:172:19:20::2/80 |
+------------------------+--------------------+------+-------+---------+----------------+----------------------+
```

Additionally, I removed the default topo file, which I think is a bit misleading

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/olekukonko/tablewriter"
 	log "github.com/sirupsen/logrus"
@@ -33,7 +34,7 @@ var inspectCmd = &cobra.Command{
 
 	Run: func(cmd *cobra.Command, args []string) {
 		if prefix == "" && topo == "" {
-			fmt.Println("provide either lab prefix (--prefix) or topology file path (--topo)")
+			fmt.Println("provide either a lab prefix (--prefix) or a topology file path (--topo)")
 			return
 		}
 		c := clab.NewContainerLab(debug)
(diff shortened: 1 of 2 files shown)


karimra:
Removing the prefix is fine, but it has some meaning. it means the container is under the local docker daemon.
BTW which other commands you are referring to that take container names ?

hellt:
@karimra actually I was wrong that leading slash prevents some docker cli from working. I read on an issue in docker repo which referenced this https://github.com/moby/moby/issues/6705#issuecomment-134183456
but I didn't notice this issue was quite old, and recent docker cli can take leading slashes with no problems.

At the same time, I think its still safe to remove the leading slash, as with the remote docker engine the path will still be recognizable if we remove the leading slash?

karimra:
Yes it's safe to remove, it looks better without "/". and we dont support remote deployment/inspect/destroy/...

PR #80, opened 2020-10-04
decouple container creation and container details retrieval

karimra:
This PR decouples containers creation/starting and container inspect to get the IPs and PID.
This allows to run the creation and starting functions concurrently.

This PR also rewrites the hosts file generation function, as well as moving the inspect cmd table output to a separate function to allow its reuse in deploy command.

Got rid of a few log.Fatal as well, replaced by returning an error on the cobra command level

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -126,9 +126,7 @@ var deployCmd = &cobra.Command{
 		}
 
 		// show topology output
-		if err = c.CreateLabOutput(); err != nil {
-			log.Error(err)
-		}
+		c.CreateHostsFile()
 		return nil
 	},
 }
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -60,11 +60,10 @@ var deployCmd = &cobra.Command{
 		if err != nil {
 			return fmt.Errorf("failed to generate rootCa: %v", err)
 		}
-		if debug {
-			log.Debugf("root CSR: %s", string(rootCerts.Csr))
-			log.Debugf("root Cert: %s", string(rootCerts.Cert))
-			log.Debugf("root Key: %s", string(rootCerts.Key))
-		}
+
+		log.Debugf("root CSR: %s", string(rootCerts.Csr))
+		log.Debugf("root Cert: %s", string(rootCerts.Cert))
+		log.Debugf("root Key: %s", string(rootCerts.Key))
 
 		// create bridge
 		if err = c.CreateBridge(ctx); err != nil {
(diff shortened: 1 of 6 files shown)


karimra:
@hellt I will add the "/" trimming to this PR after the other PR is merged.

hellt:
@karimra in a615950 I added a condition to create symlinks only if links are present

links section is absent when we want to create a node with just a management interface

karimra:
Creating the link doesn't hurt, it just creates a network namespace with the docker name. 

In the future we can add a connect command that creates a link between 2 already existing docker containers.

hellt:
I conditioned this func, which creates a symlink https://github.com/srl-wim/container-lab/blob/a615950fe4de0b846e53ad8c30bdae610985894c/clab/docker.go#L339
I dont think its ok to create a symlink in case no links are needed, because the `destroy` command then doesn't clean it up (since no c.Links is present in the topo)

cmd/deploy.go
```diff
@@ -52,26 +57,25 @@ var deployCmd = &cobra.Command{
...
 		tpl, err := template.ParseFiles(rootCaCsrTemplate)
 		if err != nil {
-			log.Fatalf("failed to parse rootCACsrTemplate: %v", err)
+			return fmt.Errorf("failed to parse rootCACsrTemplate: %v", err)
 		}
 		rootCerts, err := c.GenerateRootCa(tpl, clab.CaRootInput{Prefix: c.Conf.Prefix})
 		if err != nil {
-			log.Fatalf("failed to generate rootCa: %v", err)
-		}
-		if debug {
```
hellt, line 61: so we want to display cert content even when debug is not on?

hellt:
LGTM

karimra:
I still think that change in cLab.CreateContainer is not right, it checks if there are links in the topology to create a symlink to the container PID, not if that specific container has link or not.
if a topology has both containers with and without links, those without links will get their symlink created but never deleted because no link ( in c.Links ) references them.

What you are trying to fix here is better fixed in cLab.DeleteVirtualWiring, not during the container creation.

cmd/deploy.go
karimra: The log.Debugf takes care of that. the debug level is already set in rootCmd.PreRun command.
the if debug here is redundant.

hellt:
reverted the commit
is it ok to merge the PR from your pov?

karimra:
Should be ok now, the destroy behavior can be resolved in another PR

PR #81, opened 2020-10-04
change image pull function signature, only node image name is needed

karimra:
Change image pull function signature, only node image name is needed within the function.
Also removed the named return var, the error is redeclared before first use.


hellt:
lgtm

Issue #85, opened 2020-10-07
merge generated hosts file with the system one


karimra:
containerlab deploy should merge the generated hosts file with the system's hosts file.
containerlab destroy should remove the entries specific to the deleted fabric

hellt:
closed via #89

PR #89, opened 2020-10-19
Hosts file

karimra:
#85

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -196,6 +196,7 @@ func createHostsFile(containers []types.Container, bridgeName string) error {
 	return nil
 }
 
+// hostEntries builds an /etc/hosts compliant text blob (as []byte]) for containers ipv4/6 address<->name pairs
 func hostsEntries(containers []types.Container, bridgeName string) []byte {
 	buff := bytes.Buffer{}
 	for _, cont := range containers {
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"path"
+	"strings"
 	"sync"
 	"text/template"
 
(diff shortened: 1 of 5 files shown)


cmd/deploy.go
```diff
@@ -170,10 +171,31 @@ func setFlags(conf *clab.Conf) {
...
+	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_WRONLY, os.ModeAppend)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	_, err = f.WriteString("\n")
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(data)
```
hellt, line 191: do we want to check if the entry we about to add was already in the /etc/hosts?
can be a leftover from a failed destroy or a misconfig/clash of names.

cmd/deploy.go
```diff
@@ -170,10 +171,31 @@ func setFlags(conf *clab.Conf) {
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func hostsEntries(containers []types.Container, bridgeName string) []byte {
```
hellt, line 198: how about to put a comment to this func, smth like:

hostEntries builds an /etc/hosts compliant text blob (returned as a byte slice) by adding containers IPv4/v6 addresses and mapping them to container names.

hellt:
@karimra I had issue with using container name in the stop/remove funcs of the deletion sequence:

image

to alleviate this issue, I added the container ID to the func interface in b4e77d6. What do you think?

karimra:
In the destroy.go file, you just need to remove the `/` prefix. no need to change the method signature.
Doesn't make sense to pass both name and ID.

cmd/deploy.go
karimra: `data` here is all the entries as a single []byte.
adding entries one by one might lead to discrepancies. It's more simple to add all entries at once.
We might need a cleanup function at deploy. a function that deletes entries with the same prefix as the fabric to be deployed

cmd/destroy.go
```diff
@@ -16,54 +21,51 @@ var destroyCmd = &cobra.Command{
...
+	RunE: func(cmd *cobra.Command, args []string) error {
 		c := clab.NewContainerLab(debug)
 		err := c.Init(timeout)
 		if err != nil {
-			log.Fatal(err)
+			return err
 		}
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		if prefix != "" {
```
hellt, line 27: you removed the prefix based removal
does it mean that you see that destroy should act on the topo file only?
In that case we need to remove the `-p/--prefix` flag from cmd

karimra: yes, it doesn't work properly, I was hoping to fix it in another PR.
The flag can be removed

hellt:
added a few log messages in #93

Issue #99, opened 2020-11-04
do not try to remove docker bridge if it has active endpoints


hellt:
Currently, the lab topologies reuse the management network provided by docker network `containerlab`.
When we destroy the topo, we try to delete this network and its entities, but that errors, due to other containers from other labs might be still attached:

```
INFO[0001] Deleting docker bridge ...
ERRO[0001] Error response from daemon: error while removing network: network containerlab id 1a050256ae16f2f7e73d2c29a336b04b133698a17fc77230fdea34b29bc5a2cd has active endpoints
```

the proposal is to first check if the network has active endpoints attached and if it doesn't - try and delete it

PR #101, opened 2020-11-05
do not delete network if it has endpoints

karimra:
#99


karimra:
will re add this in another PR

PR #103, opened 2020-11-05
check if network has active endpoints before deleting it

### LINKED ISSUE #99: do not try to remove docker bridge if it has active endpoints

karimra:
close #99

### DIFF
--- a/clab/docker.go
+++ b/clab/docker.go
@@ -126,8 +126,10 @@ func (c *cLab) DeleteBridge(ctx context.Context) (err error) {
 	numEndpoints := len(nres.Containers)
 	if numEndpoints > 0 {
 		log.Warnf("network '%s' has %d active endpoints", c.Conf.DockerInfo.Bridge, numEndpoints)
-		for _, endp := range nres.Containers {
-			log.Debugf("'%s' is connected to %s", endp.Name, c.Conf.DockerInfo.Bridge)
+		if c.debug {
+			for _, endp := range nres.Containers {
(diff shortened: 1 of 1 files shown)


clab/docker.go
```diff
@@ -119,6 +119,18 @@ func (c *cLab) CreateBridge(ctx context.Context) (err error) {
...
 	nctx, cancel := context.WithTimeout(ctx, c.timeout)
 	defer cancel()
+	nres, err := c.DockerClient.NetworkInspect(ctx, c.Conf.DockerInfo.Bridge)
+	if err != nil {
+		return err
+	}
+	numEndpoints := len(nres.Containers)
+	if numEndpoints > 0 {
+		log.Warnf("network '%s' has %d active endpoints", c.Conf.DockerInfo.Bridge, numEndpoints)
+		for _, endp := range nres.Containers {
```
hellt, line 129: do we want to enter this loop only if the debug is set?

karimra: we could, but I don't see much gain

Issue #75, opened 2020-09-30
docker network name should be optional


hellt:
as discussed with @karimra the proposal is to make docker network name optional (auto-generated based on lab prefix)

hellt:
it already is, the defaults for docker network related values are corrected in #112

PR #112, opened 2020-11-06
implement short clab prefix instead of a longer `containerlab` one

### LINKED ISSUE #104: use a shorter `clab` prefix for container/if names where possible
container names are not start with `containerlab` prefix. I think its a bit long, the `clab` makes it way shorter -> faster to type, less space on the screen

hellt:
close #104, #75

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -10,11 +10,6 @@ import (
 
 // var debug bool
 
-const (
-	prefix    string = "clab"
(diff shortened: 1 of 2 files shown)


hellt:
@karimra PTAL, changes are short, but it seems they suffice the needs of #104

karimra:
LGTM

PR #115, opened 2020-11-09
Topology definition rework

### LINKED ISSUE #97: rework topology definition file
This issue will track the design decisions on the new "topology definition" format clab should adopt.

The recent proposal was:

```yaml
name: wan-topo
mgmt:
  # network is not mandatory, if not specified it default to $name-network
  network: $name-network
  ipv4_range: <ipv4 mgmt range>
  ipv6_range: <ipv6 mgmt range>
topology:
  defaults:
    kind: srl
  kinds:
    srl:
      image: srlinux20.6.1-286
      type: ixr6
    alpine:
      image: henderiw/client-alpine:1.0.0
  
  nodes:
    node1:
    node2:
    client1:
      kind: alpine
    client2:
      kind: alpine
  
  links:
    - endpoints: [ "node1:e1-1", "node2:e1-1"]
    - endpoints: [ "node1:e1-2", "client1:eth1"]
    - endpoints: [ "node2:e1-2", "client2:eth1"]
```

The areas that we need to have a discussion on:

1. how to model endpoints/links to allow for data augmentation. For example, when we will do the config generation for interfaces, how do we add the interface-related information to the endpoints.
2. where to keep the protocol and other logical-inventory related data for config generation?

hellt:
This PR will be in Draft whilist it undergoes development. The changes will appear hear, I will try to keep commits per sub-feature to make the review easier

1. renamed flag to provide docker network name (from `--bridge <name>` to `--network <name>`) 8c48f50
2. moved flag variables that are defined in root.cmd from config.go b07a8d1
3. Added comment clarifying the role of `cLab.Init()` in b7d42d2
4. reworked `GetTopology` (removed shortname for files, renamed File struct to TopoFile) in 3ad1543
5. renamed cLab.Prefix to cLab.name (5b0b5e0)
6. renamed docker_info section to `mgmt` section. Also removed ipv4/6 gateway calculation, since it will be used by docker automatically as a GW in the management network (1fec139)
7. renamed `duts` config element to `topology` with a nested structure of 
    ```
    type Topology struct {
       	    Defaults NodeConfig
	    Kinds    map[string]NodeConfig
	    Nodes    map[string]NodeConfig
    }
    ```
8. renamed lab examples topologies in ac98ce9
9. renamed `dut` vars/strings to nodeCfg or node accordingly. ( d9e6404)
closes #97

### DIFF
--- a/clab/cert.go
+++ b/clab/cert.go
@@ -79,7 +79,7 @@ func (c *cLab) GenerateCert(ca string, caKey string, csrJSONTpl *template.Templa
 		Name:     node.ShortName,
 		LongName: node.LongName,
 		Fqdn:     node.Fqdn,
-		Prefix:   c.Conf.Prefix,
+		Prefix:   c.Conf.Name,
 	}
 	CreateDirectory(path.Join(c.Dir.LabCA, input.Name), 0755)
 	var err error
--- a/clab/cert.go
+++ b/clab/cert.go
@@ -79,7 +79,7 @@ func (c *cLab) GenerateCert(ca string, caKey string, csrJSONTpl *template.Templa
 		Name:     node.ShortName,
 		LongName: node.LongName,
 		Fqdn:     node.Fqdn,
-		Prefix:   c.Conf.Name,
+		Prefix:   c.Config.Name,
 	}
 	CreateDirectory(path.Join(c.Dir.LabCA, input.Name), 0755)
 	var err error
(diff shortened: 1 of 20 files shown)


karimra:
https://github.com/srl-wim/container-lab/blob/3d5ee2bd1144c60f0ed5632c160e884c066d6a6f/clab/config.go#L376
This message needs to be changed as well, to reflect the new topology format.
It's a good idea to get rid of the panic while we are here, make it return an error, that can be handled in clab.ParseToplogy then in whatever cmd that called it.

hellt:
@karimra addressed this in dff9317

PR #116, opened 2020-11-09
Fix generation of graph for topologies with nodes not assigned to a group

steiler:
e.g. the clos01 topology fails to graph, since the group attribute in the resulting dot file is "group=" with no value assigned.
This fixes the issue.
However we might have a discussion on the different group styles if they are required and useful or not.


clab/graph.go
```diff
@@ -33,18 +33,18 @@ func (c *cLab) GenerateGraph(topo string) error {
...
-			attr["fillcolor"] = "green"
-			attr["color"] = "green"
-			attr["fontcolor"] = "black"
+		if len(strings.TrimSpace(node.Group)) != 0 {
+			attr["group"] = node.Group
+			if strings.Contains(node.Group, "bb") {
+				attr["fillcolor"] = "blue"
+				attr["color"] = "blue"
+				attr["fontcolor"] = "white"
+			} else if strings.Contains(node.Kind, "srl") {
```
hellt, line 42: I wonder why we color SRL nodes only if the groups were defined?
Is it because that way the nodes are visible on the groups background?

PR #121, opened 2020-11-13
add graph generation using d3js

karimra:
This PR add the possibility to generate a graph using d3js force graph, https://github.com/d3/d3-force
It adds 2 flags to graph command, `--srv` and `--template` 

`--srv` enables this feature is present, it specifies an address a http server will listen to.
The server serves a single path `/`, where a d3js force graph is generated based on a default template.
`--template` can be used to override the default template

The html template is kind of bare bones now but it works...
The page shows the generated d3js force graph, each node can be dragged to a desired position.
The nodes have a default radius of 20px.
The graph section has an invisible grid (10px *10px) to help align the nodes
There is a button to export/download the graph as a PNG file.

In the future, we can include some extra fields in the topo file to customize the graph generation per kind/group/node
or add buttons to manipulate the different parameters of the graph.

image


hellt:
Knowing how many hours you spent in this rabbit hole its sad that I cant thumbs up this PR twice =)

steiler:
Is that a plain static graph that is being generate?
I have made very good experience with https://visjs.github.io/vis-network/examples/ .
We could add popups with additional Infos like IPs and so on...
Anyhow that also already looks great!

steiler:
Ok, I've had a look around and figured, that the feature set is quite the same I guess. Also the on click events are supported.
So we can maybe add popups on node and link hover / click which show additional infos like like the mentioned IPs etc.

karimra:
Yes that's the idea, but I didn't want to PR to get too big and hard to review.
Wanted to get this out for people to use and for someone with better mastery of html/css ( I have none :D ) to improve.

Management IP addresses and container state should be easy to add. 
But link IPs would be node specific and will require a gnmi client within containerlab.
it can even become a dashboard for all labs, http://address:port/${LAB_NAME}

karimra:
btw it's possible to use visjs as well, with the `--template flag`.
Currently, the `graphTopo` is passed to the golang template as a json object, 
it has most of the data available to containerlab, except IPs and state. 
I can add that in a future PR, I wanted to change something in clab api first

karimra:
To use info from docker, we need to explicitly tell graph cmd to create a docker client. 
Via a flag ideally.
Any suggestions for a flag name? @hellt @steiler

hellt:
@karimra any reason not to make docker client to init for graph? Perfomance hit is not something we should care here imo

karimra:
in case a user wants to generate a graph locally, without deploying any container

cmd/graph.go
```diff
@@ -1,34 +1,110 @@
...
+		if err != nil {
+			return err
 		}
+		return nil
 	},
 }
 
 func init() {
 	rootCmd.AddCommand(graphCmd)
+	graphCmd.Flags().StringVarP(&srv, "srv", "", "", "server address to view, customize and download your topology PNG")
```
hellt, line 108: maybe use a long name as `--server` with a shorthand `-s`?

Also I would have changed the description to `HTTP server address...and export your topology`

cmd/graph.go
```diff
@@ -1,34 +1,110 @@
...
+	graphCmd.Flags().StringVarP(&tmpl, "template", "", templatePath, "golang html template used to generate the graph")
```
hellt, line 109: add a shorthand `-t`?

I would have removed golang from description: "path to HTML template used for graph generation"

hellt:
> in case a user wants to generate a graph locally, without deploying any container

to me that is a less popular case that can be handled with a flag (not the other way around)

cmd/graph.go
karimra: `-t` is already taken by `--topo` in rootCmd.
why remove golang?, it's not any template

cmd/graph.go
karimra: will change

karimra:
> > in case a user wants to generate a graph locally, without deploying any container
> 
> to me that is a less popular case that can be handled with a flag (not the other way around)

Right now `graph` doesn't create a docker client. want to keep the same behavior.
anyways, question still stands, what would be a good flag name for not creating a docker client

cmd/graph.go
hellt: this additional info doesn't seem important to me to explain it in the flag descr

but it might be my personal feeling on what makes a good flag descr.
If you intend to keep it, maybe change Golang->Go as the former is discouraged to use

hellt:
> what would be a good flag name for not creating a docker client

I'd play with

* `--brief` - brief means there is not augmented information derived from docker 
* `--only-topology`- means that only the topology information will be used to draw a graph

cmd/graph.go
```diff
@@ -1,35 +1,110 @@
...
+		})
+		for _, l := range c.Links {
+			gtopo.Links = append(gtopo.Links, link{
+				Source:         l.A.Node.ShortName,
+				SourceEndpoint: l.A.EndpointName,
+				Target:         l.B.Node.ShortName,
+				TargetEndpoint: l.B.EndpointName,
+			})
+		}
+		b, err := json.Marshal(gtopo)
```
hellt, line 84: add a message what is this dump about for a debug reader to understand it better?

hellt: sorry, wrong line, it was about the debug statement a few lines below

cmd/graph.go
```diff
@@ -21,15 +49,63 @@ var graphCmd = &cobra.Command{
...
+			Data: template.JS(string(b)),
 		}
+		tmpl := template.Must(template.ParseFiles(tmpl))
+		http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+			tmpl.Execute(w, topoD)
+		})
 
-		if err := c.GenerateGraph(topo); err != nil {
-			log.Error(err)
+		log.Printf("Listening on %s...", srv)
```
hellt, line 98: should it be `log.Infof`?

cmd/graph.go
hellt: I think we also need to set a default path here in the flag that will point to `/etc/containerlab/templates/d3js/index.html`

hellt:
It would be nice to try and use the custom SVG instead of circles as done here http://bl.ocks.org/explunit/5988971

I failed to make it work without looking into d3 docs though

cmd/graph.go
karimra: The default value is in the const `templatePath`

karimra: updated golang to Go, and changed the const name to `defaultTemplatePath`

cmd/graph.go
karimra: done

cmd/graph.go
karimra: done

karimra:
> It would be nice to try and use the custom SVG instead of circles as done here http://bl.ocks.org/explunit/5988971
> 
> Another example - https://observablehq.com/@saradelessio/d3-scatterplot-with-imported-svg-symbols
> 
> I failed to make it work without looking into d3 docs though

Yes, there are lots of examples on bl.ocks.org 
I think there is a lot that can be done on the html/css/js side.
I'd rather have it done by adding other templates that can be shipped with containerlab or added to the docs website.
We can have different "tiers" of more or less advanced templates

All this remains a matter of personal taste.
I prefer that users try to write their own templates, add their own svgs/icons, etc...

templates/d3js/index.html
```diff
@@ -0,0 +1,313 @@
...
+        height: 50%;
+      }
+    </style>
+  </head>
+
+  <body>
+    <div id="content" style="margin:0 auto;">
+      <h1>ContainerLab Topology '{{ .Name }}'</h1>
+      <div>
+        <button id='saveButton'>Export topology '{{ .Name }}' as PNG</button>
```
hellt, line 27: can you add `class="button"` here to make this button inherit bootstrap styles?

karimra: done

karimra:
image

Issue #68, opened 2020-09-27
container attachement without subshelling


hellt:
In this PR weaveworks moved from shell based container attachement to Go/syscall based
this can be used for inspiration; for instance remove ethtool subshelling and/or creating veth pairs

https://github.com/weaveworks/weave/pull/2307

PR #123, opened 2020-11-13
removed subshelling from veth between containers

steiler:
Added MTU capability
Had to remove concurrency -> further investigation required

This is still work in progress, fixing #68


hellt:
@steiler when I was researching the topic, I found out that the CNI repo is the best one to draw inspiration from on how to create p2p links. https://github.com/containernetworking/plugins

They have a good explanation how dangerous is it to use netns switching with goroutines -  https://github.com/containernetworking/plugins/tree/master/pkg/ns

For our cause of p2p links creation we should browse this repo and see how they create veth pairs, instead of using vishavanda/netns package directly

karimra:
@steiler here is a blog post about locking goroutines to a thread and thus to a network namespace, 
https://www.weave.works/blog/linux-namespaces-golang-followup
this is probably there reason you had to  disable concurrency

steiler:
I've tested this now and it works fine for me.

This should implicitly also resolve #197 and lays foundation for #196 which is already available as a field in the internal struct LinkAttributes  which just needs to be filled. hence just the front-end needs implementation.

hellt:
@steiler it seems this PR doesn't address #197? In that issue the prime focus is on the single-interfaces linux containers (eth0) which we do not explicitly create
They are created by docker

steiler:
@hellt you're right. It wasn't clear to me that the default docker bridge was meant.
Any additional feedback on this pull request?

hellt:
I think I will get to it starting next year
I still want to see if reusing the koko code will make it cleaner

hellt:
Changes are made in #209

PR #124, opened 2020-11-13
allow user defined mounts for kinds and nodes

### LINKED ISSUE #88: bind mount directories should be possible to set per kind and node
To allow mounting files to the containerlab modes, the `mount` option have to be created for kinds and nodes.
The use case is to add custom binaries or configs for testing containers (like adding gobgp to a vanilla alpine)

```yaml
kind_defaults:
    my_tools:
      type: custom
      image: pklepikov/ubuntu-tools
      mounts:
       - $(pwd)/exabgp /home/admin/exabgp
       - $(pwd)/gobgp /home/admin/gobgp
```

hellt:
closes #88 

this PR allows users to define Bind Mounts for nodes or kinds. For example:

```yaml
topology:
  kinds:
    srl:
      type: ixr6
      image: srlinux
      license: license.key
      binds:
        - "/tmp/1test:/tmp/bla:ro"
```

or 
```yaml
topology:
  nodes:
    srl:
      kind: srl
      binds:
        - "/tmp/1test:/tmp/bla"
```

The format of bindmount string follows the format of docker/oci.

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -99,7 +99,6 @@ type Node struct {
 	Cmd       string
 	Env       []string
 	Mounts    map[string]volume
-	Volumes   map[string]struct{}
 	Binds     []string
 
 	TLSCert   string
@@ -342,14 +341,6 @@ func (c *cLab) NewNode(nodeName string, nodeCfg NodeConfig, idx int) error {
 		v.ReadOnly = true
 		log.Debug("Topology File: ", v.Source)
 		node.Mounts["topology"] = v
-
-		node.Volumes = make(map[string]struct{})
-		node.Volumes = map[string]struct{}{
-			node.Mounts["license"].Destination:  {},
-			node.Mounts["config"].Destination:   {},
-			node.Mounts["envConf"].Destination:  {},
-			node.Mounts["topology"].Destination: {},
-		}
 
 		bindLicense := node.Mounts["license"].Source + ":" + node.Mounts["license"].Destination + ":" + "ro"
 		bindConfig := node.Mounts["config"].Source + ":" + node.Mounts["config"].Destination + ":" + "rw"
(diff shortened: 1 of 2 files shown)


clab/config.go
```diff
@@ -183,6 +183,13 @@ func (c *cLab) kindInitialization(nodeCfg *NodeConfig) string {
 	return c.Config.Topology.Defaults.Kind
 }
 
+func (c *cLab) mountsInit(nodeCfg *NodeConfig) []string {
+	if len(nodeCfg.Mounts) != 0 {
+		return nodeCfg.Mounts
+	}
+	return c.Config.Topology.Kinds[nodeCfg.Kind].Mounts
```
karimra, line 190: This probably needs a check, if key `nodeCfg.Kind` exists in the map

clab/config.go
```diff
@@ -98,8 +99,7 @@ type Node struct {
 	User      string
 	Cmd       string
 	Env       []string
-	Mounts    map[string]volume
-	Volumes   map[string]struct{}
+	Mounts    []string // Bind mounts strings (src:dest:options)
```
karimra, line 102: btw why add an additional fields here, `Mounts` we can use the already existing `Binds` field. 
It's also called binds on the docker api.
Otherwise, `&container.HostConfig.Mounts` can be used with `MountPoint.Type = 	mount.TypeBind`
Haven't tried it, so not sure it works.

hellt: modified this in e4d5dd9
PTAL

clab/config.go
hellt: since Go will return an empty value for a type that is stored under a missing key in a map, that serves our purpose here

karimra: ok, it's not a pointer there

karimra:
lgtm

hellt:
merging
docs will be added in #119

PR #127, opened 2020-11-16
Per node cfg template

karimra:
this PR, allows to specify a per node config template using yaml filed `config` under defaults, kinds or the node itself


hellt:
lgtm

PR #128, opened 2020-11-16
rename --prefix to --name to match the new topo file format

karimra:
this PR renames `--prefix` to `--name` to match the new topo file format


hellt:
lgtm

PR #133, opened 2020-11-16
destroy cleanup

karimra:
This PR adds `--cleanup` flag to destroy command to delete the lab directory when destroying a lab

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -16,6 +16,8 @@ import (
 	"github.com/srl-wim/container-lab/clab"
 )
 
+var cleanup bool
+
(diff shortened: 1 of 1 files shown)


cmd/destroy.go
```diff
@@ -81,6 +89,7 @@ var destroyCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(destroyCmd)
+	destroyCmd.Flags().BoolVarP(&cleanup, "cleanup", "", false, "delete config artefacts")
```
hellt, line 92: we know what config artefacts are, likely users dont
maybe give a simplified description of `delete lab directory`?

PR #140, opened 2020-11-17
License is now stored in node.Labdir instead of clab.Dir

### LINKED ISSUE #131: license should not be kept in lab directory with license.key name
Currently, the clab doesn't allow to have different license files for different nodes/kinds, as it copies the license file to the lab directory - https://github.com/srl-wim/container-lab/blob/81e0f749835e25d6d1f0360b38f96a67b7f9b8f0/clab/file.go#L146

proposal is to copy the license file under the node directory explicitly, and mount it from there

hellt:
This allows for different nodes to have different license files

closes #131


karimra:
lgtm

PR #144, opened 2020-11-17
Generate topology file

karimra:
This PR adds `generate` command that is used to generate a topology file based on supplied flags,
The topology file is printed to stdout or saved to a custom location if specified with `--file` 

flags:
`--kind`: `srl` or `ceos` for the moment, specifies the kind of node to be deployed

`--network`: the docker network name to be used a management network

`--ipv4-subnet` : management network IPv4 subnet range

`--ipv6-subnet` : management network IPv6 subnet range

`--image` : container image name, can be prefixed with the node kind. <kind>:<image_name>

 if the kind value is missing from `--image` is is taken from `--kind` flag

`--license`: path to license file, can be prefix with the node kind. <kind>:/path/to/file

if the kind value is missing from `--license` is is taken from `--kind` flag

`--nodes` : comma separated nodes definitions in format <num_nodes>:<kind>:<type>, each defining a Clos network stage

if the kind value is missing from `--nodes` is is taken from `--kind` flag

for e.g: `--nodes 3:srl,2:srl,1:ceos` results in a 3 states fabric with 3 srls, 2 srls, then 1 ceos node

`--node-prefix` : prefix used in node names

`--group-prefix` : prefix used in group names

EDIT: added `--deploy` flag which will save the generated topology to `--file` value or `$name.yaml` if `--file` is not set.
it then calls deployCmd.RunE with `reconfigure=true` and `topo=file`

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -42,7 +42,7 @@ func init() {
 	configCmd.AddCommand(configApplyCmd)
 	//
 	configGenerateCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
-	generateCmd.Flags().BoolVarP(&workload, "workload", "", false, "generate workloads config")
+	configGenerateCmd.Flags().BoolVarP(&workload, "workload", "", false, "generate workloads config")
 	//
 	configApplyCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
 	configApplyCmd.Flags().BoolVarP(&workload, "workload", "", false, "generate workloads config")
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -19,7 +19,7 @@ var configCmd = &cobra.Command{
 }
 
 // generateCmd represents the generate command
-var generateCmd = &cobra.Command{
+var configGenerateCmd = &cobra.Command{
 	Use:   "generate",
 	Short: "generate configuration for the lab",
 	Run: func(cmd *cobra.Command, args []string) {
@@ -28,7 +28,7 @@ var generateCmd = &cobra.Command{
 }
 
 // applyCmd represents the apply command
-var applyCmd = &cobra.Command{
+var configApplyCmd = &cobra.Command{
 	Use:   "apply",
 	Short: "apply the generated configuration on the lab nodes",
 	Run: func(cmd *cobra.Command, args []string) {
@@ -38,12 +38,12 @@ var applyCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(configCmd)
-	configCmd.AddCommand(generateCmd)
-	configCmd.AddCommand(applyCmd)
+	configCmd.AddCommand(configGenerateCmd)
+	configCmd.AddCommand(configApplyCmd)
 	//
-	generateCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
+	configGenerateCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
 	generateCmd.Flags().BoolVarP(&workload, "workload", "", false, "generate workloads config")
 	//
-	applyCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
-	applyCmd.Flags().BoolVarP(&workload, "workload", "", false, "generate workloads config")
+	configApplyCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
+	configApplyCmd.Flags().BoolVarP(&workload, "workload", "", false, "generate workloads config")
 }
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -158,7 +158,7 @@ func generateTopologyConfig(name, network, ipv4range, ipv6range string, images m
 			config.Topology.Kinds[k] = knd
 			continue
 		}
-		config.Topology.Kinds[k] = clab.NodeConfig{Image: lic}
+		config.Topology.Kinds[k] = clab.NodeConfig{License: lic}
 	}
 	for i := 0; i < numStages-1; i++ {
 		interfaceOffset := uint(0)
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -71,10 +71,12 @@ var generateCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		fmt.Println("licenses:", licenses)
 		images, err := parseFlag(kind, image)
 		if err != nil {
 			return err
 		}
+		fmt.Println("images:", images)
 		nodeDefs, err := parseNodesFlag(kind, nodes...)
 		if err != nil {
 			return err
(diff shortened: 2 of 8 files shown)


karimra:
@hellt, good to move the user id check to rootCmd.PersistentPreRun instead of init() ?

cmd/generate_test.go
```diff
@@ -0,0 +1,189 @@
...
+		result, err := parseFlag(set.in.kind, set.in.lics)
+		if !cmp.Equal(result, set.out.out) {
+			t.Errorf("failed at '%s', expected %v, got %+v", name, set.out.out, result)
+		}
+		if err != set.out.err {
+			t.Errorf("failed at '%s', expected error %+v, got %+v", name, set.out.err, err)
+		}
+		t.Logf("test '%s' OK", name)
+	}
+}
```
hellt, line 175: maybe implement `t.Run(name, func(t *testing.T)` here to allow for named test runs?

cmd/config.go
```diff
@@ -38,12 +38,12 @@ var applyCmd = &cobra.Command{
...
 func init() {
 	rootCmd.AddCommand(configCmd)
-	configCmd.AddCommand(generateCmd)
-	configCmd.AddCommand(applyCmd)
+	configCmd.AddCommand(configGenerateCmd)
+	configCmd.AddCommand(configApplyCmd)
 	//
-	generateCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
+	configGenerateCmd.Flags().BoolVarP(&infra, "infra", "", false, "generate infra config")
 	generateCmd.Flags().BoolVarP(&workload, "workload", "", false, "generate workloads config")
```
hellt, line 45: this belongs to configGenerateCmd

cmd/generate.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// generateCmd represents the generate command
+var generateCmd = &cobra.Command{
+	Use:     "generate",
+	Aliases: []string{"gen"},
+	Short:   "generate a Clos topology file, based on provided flags",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if name == "" {
+			return errors.New("--name is mandatory")
```
hellt, line 71: what about an expanded error message like: "provide a lab name with `--name` flag"

cmd/generate.go
```diff
@@ -0,0 +1,294 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+var interfaceFormat = map[string]string{
+	"srl":    "e1-%d",
+	"ceos":   "eth%d",
+	"linux":  "eth$d",
+	"bridge": "veth%d",
+}
+var supportedKinds = []string{"srl", "ceos", "linux", "bridge", "sonic", "crpd"}
```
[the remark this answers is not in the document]
karimra: let's keep it here for now, we can reconsider when sonic and crpd are added to clab.
It would be an exported global var...

cmd/config.go
karimra: fixed, we can consider removing the config cmd as well

cmd/generate.go
```diff
@@ -0,0 +1,294 @@
...
+func generateTopologyConfig(name, network, ipv4range, ipv6range string, images map[string]string, licenses map[string]string, nodes ...nodesDef) ([]byte, error) {
+	numStages := len(nodes)
+	config := &clab.Config{
+		Name: name,
+		Topology: clab.Topology{
+			Kinds: make(map[string]clab.NodeConfig),
+			Nodes: make(map[string]clab.NodeConfig),
+		},
+	}
+	config.Mgmt.Network = network
```
hellt, line 145: do we want to render the `mgmt` container if the user did not pass the value?
What if we skip setting the default value for mgmtNetName at L123? That way during the deploy we will apply the necessary defaults

cmd/generate.go
```diff
@@ -0,0 +1,294 @@
...
+	"ceos":   "eth%d",
+	"linux":  "eth$d",
+	"bridge": "veth%d",
+}
+var supportedKinds = []string{"srl", "ceos", "linux", "bridge", "sonic", "crpd"}
+
+const (
+	defaultSRLType     = "ixr6"
+	defaultNodePrefix  = "node"
+	defaultGroupPrefix = "group"
```
hellt, line 43: since by default we build a CLOS topo, maybe we should use the group prefix value as `tier`?
That will make sense to a user, since CLOS topos break down by tiers, which will make a structure:

* tier1
* tier2
* etc

cmd/generate.go
karimra: the defaults of `--network` in `generate` and `deploy` are the same, so it will be the same result

cmd/generate.go
karimra: I have no opinion on the default value, it can be changed with `--group-prefix` anyways

cmd/generate.go
hellt: yes, but my point is that it is not really needed to set a network to a default name `clab` here, as this will be done in the backend

i.e. if we change the default name later in deploy, we would need to remember to change the default value of that flag here as well

cmd/generate.go
```diff
@@ -0,0 +1,294 @@
...
+	for k, img := range images {
+		config.Topology.Kinds[k] = clab.NodeConfig{Image: img}
+	}
+	for k, lic := range licenses {
+		if knd, ok := config.Topology.Kinds[k]; ok {
+			knd.License = lic
+			config.Topology.Kinds[k] = knd
+			continue
+		}
+		config.Topology.Kinds[k] = clab.NodeConfig{Image: lic}
```
hellt, line 161: clab.NodeConfig{Image: lic} -> clab.NodeConfig{License: lic}

karimra: fixed

cmd/generate.go
karimra: removed the default value

cmd/generate.go
```diff
@@ -0,0 +1,294 @@
...
+var license []string
+var nodePrefix string
+var groupPrefix string
+var file string
+var deploy bool
+
+type nodesDef struct {
+	numNodes uint
+	kind     string
+	typ      string
```
hellt, line 61: typo or you wanted the field to be `typ`?

cmd/generate.go
```diff
@@ -0,0 +1,294 @@
...
+	generateCmd.Flags().StringVarP(&kind, "kind", "", "srl", fmt.Sprintf("container kind, one of %v", supportedKinds))
+	generateCmd.Flags().StringSliceVarP(&nodes, "nodes", "", []string{}, "comma separated nodes definitions in format <num_nodes>:<kind>:<type>, each defining a Clos network stage")
+	generateCmd.Flags().StringSliceVarP(&license, "license", "", []string{}, "path to license file, can be prefix with the node kind. <kind>=/path/to/file")
+	generateCmd.Flags().StringVarP(&nodePrefix, "node-prefix", "", defaultNodePrefix, "prefix used in node names")
+	generateCmd.Flags().StringVarP(&groupPrefix, "group-prefix", "", defaultGroupPrefix, "prefix used in group names")
+	generateCmd.Flags().StringVarP(&file, "file", "", "", "file path to save generated topology")
+	generateCmd.Flags().BoolVarP(&deploy, "deploy", "", false, "deploy a fabric based on the generated topology file")
+}
+
+func generateTopologyConfig(name, network, ipv4range, ipv6range string, images map[string]string, licenses map[string]string, nodes ...nodesDef) ([]byte, error) {
```
hellt, line 136: might be opinionated:

I would have renamed the `nodes ...nodesDef` to `tiers ...tiersDef` to be inline with our logic of CLOS fabric to be composed out of tiers.

cmd/generate.go
karimra: type is reserved, can be confusing

cmd/generate.go
karimra: let's keep it like this, who knows, we might add other topologies in the future...

Issue #147, opened 2020-11-18
enhance inspect cmd to list all containerlab labs


hellt:
quite often its needed to be seen which labs are running on the system

`containerlab inspect -all` could be a good candidate that will list all labs/containers by using a common tag

PR #152, opened 2020-11-18
Inspect all

karimra:
#147

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -51,7 +51,11 @@ var inspectCmd = &cobra.Command{
 		}
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		labels = append(labels, "containerlab=lab-"+name)
+		if all {
+			labels = append(labels, "containerlab")
+		} else {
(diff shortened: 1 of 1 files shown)


cmd/inspect.go
```diff
@@ -77,21 +83,32 @@ func init() {
 
 	inspectCmd.Flags().BoolVarP(&details, "details", "", false, "print all details of lab containers")
 	inspectCmd.Flags().StringVarP(&format, "format", "f", "table", "output format. One of [table, json]")
+	inspectCmd.Flags().BoolVarP(&all, "all", "a", false, "show all containerlab labs in the host")
```
hellt, line 86: "show all running containerlab labs"?

"on the host" seems to be irrelevant, since we do not support multi host deployments. at the same time adding "running" makes sense to me, since that excludes assumption that this command will analyze labs which are not running, but defined in the files in the CWD

hellt: apart from that the rest lgtm and it works as expected

karimra: changed

hellt:
tested, lgtm

PR #153, opened 2020-11-18
List all containers

karimra:
This PR allows to list all containers from clab (including non-running ones).
This will help destroying lab with exited/failed containers


hellt:
lgtm

PR #173, opened 2020-11-21
added support for config generation for ceos

hellt:
TODO: add a minimal config that supports gnmi access

### DIFF
--- a/clab/file.go
+++ b/clab/file.go
@@ -132,7 +132,7 @@ func CreateDirectory(path string, perm os.FileMode) {
 	}
 }
 
-// CreateNodeDirStructure create the directory structure and files for the clab
+// CreateNodeDirStructure create the directory structure and files for the lab nodes
 func (c *cLab) CreateNodeDirStructure(node *Node) (err error) {
 	c.m.RLock()
 	defer c.m.RUnlock()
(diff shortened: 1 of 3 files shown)


clab/file.go
```diff
@@ -132,19 +132,20 @@ func CreateDirectory(path string, perm os.FileMode) {
...
 }
 
-// CreateNodeDirStructure create the directory structure and files for the clab
+// CreateNodeDirStructure create the directory structure and files for the lab nodes
 func (c *cLab) CreateNodeDirStructure(node *Node) (err error) {
 	c.m.RLock()
 	defer c.m.RUnlock()
+
+	// create node directory in the lab directory
+	CreateDirectory(node.LabDir, 0777)
```
karimra, line 141: we don't need to create directories for bridge, alpine and linux kinds, so maybe keep this under `case "srl"` and `case "ceos"` only ?

templates/arista/ceos.cfg.tpl
```diff
@@ -0,0 +1 @@
+hostname {{ .ShortName }}
```
karimra, line 1: this is yet to be updated right ?

clab/file.go
hellt: we will have more NOS kinds which will require some configs or files to be mounted.
Lets maybe make the creation conditional if kind is not `linux | bridge`

karimra: that works too

hellt: added this in c508c96

I want to create another PR for removing alpine kind, do you agree?

karimra: you already removed it from the dir structure func, no ?

PR #176, opened 2020-11-23
deploy max-workers

karimra:
This PR adds `--max-workers` to deploy command, 
defaults to "0" meaning no limit ( as many as the number of nodes or links )

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -113,7 +113,7 @@ var deployCmd = &cobra.Command{
 							log.Debugf("Worker %d terminating...", i)
 							return
 						}
-						log.Debugf("Worker %d received node: %+v", node)
+						log.Debugf("Worker %d received node: %+v", i, node)
 						if node.Kind == "bridge" {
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -89,60 +92,107 @@ var deployCmd = &cobra.Command{
 		if err != nil {
 			return fmt.Errorf("failed to parse certCsrTemplate: %v", err)
 		}
-		// create directory structure and container per node
+
+		nodesMaxWorkers := maxWorkers
+		linksMaxWorkers := maxWorkers
```
hellt, line 97: does it make sense to set nodes/links max workers to len(nodes/links) in case if max-workers > len(nodes/links)?

hellt:
I added a flag doc here https://github.com/srl-wim/container-lab/pull/181

cmd/deploy.go
karimra: changed

PR #183, opened 2020-11-29
Binds abspath

### LINKED ISSUE #177: implement Absolute path for local path in the binds structure
Currently, the string we accept in `binds` section does not get expanded if it contains bash variables.

To make it more useful, we should translate the `${PWD}` and/or relative paths such as `./dir`

hellt:
close #177 

first merge #182 as this branch is based on it

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/docker/docker/api/types"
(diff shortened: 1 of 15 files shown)


hellt:
I think I will move resolveBindPaths function out of the bindsInit function, to allow for testing of binds propagataion the same way thats done in #182

PR #200, opened 2020-12-18
added ability to specify static IP addresses on mgmt net

### LINKED ISSUE #192: implement static IP assignment for nodes
To make a lab management addresses predictable and definable via topology file the proposal is to add the node-specific fields:

* `ipv4`
* `ipv6`

When values are given for these fields, the docker daemon will assign the respective IP addresses to the nodes.

ref https://docs.docker.com/config/containers/container-networking/#ip-address-and-hostname

hellt:
closes #192


hellt:
Hi @karimra
do you see any issues with that implementation of static assignment of IP addresses?

karimra:
This can fail if a container without a static IP gets assigned an IP that is configured statically under another container.
A few option to go around this:
- All containers should have a static IP
- IPs should be automatically added under all containers if at least one container has a configured IP.
- Schedule containers with a static IP first

hellt:
Good suggestion.
I am inclined to put a note in the docs that is staticly assigned addresses
are needed, they should be configured for all containers to prevent address
collision issues

PR #202, opened 2020-12-18
added default type for srl based on first srlType slice entry

steiler:
This is the implementation of #199


hellt:
Hi @steiler 
didn't know you was working on this. 
I added #203 with tests, so I am closing this one as a dup

PR #203, opened 2020-12-18
refactored type init

### LINKED ISSUE #199: make srlinux default type to be ixr6

hellt:
close #199

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -226,11 +226,22 @@ func (c *cLab) groupInitialization(nodeCfg *NodeConfig, kind string) string {
 	return c.Config.Topology.Defaults.Group
 }
 
-func (c *cLab) typeInitialization(nodeCfg *NodeConfig, kind string) string {
-	if nodeCfg.Type != "" {
+// initialize SRL HW type
+func (c *cLab) typeInit(nodeCfg *NodeConfig, kind string) string {
+	switch {
+	case nodeCfg.Type != "":
 		return nodeCfg.Type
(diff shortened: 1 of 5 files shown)


clab/config.go
```diff
@@ -226,11 +226,22 @@ func (c *cLab) groupInitialization(nodeCfg *NodeConfig, kind string) string {
...
 		return nodeCfg.Type
+	case c.Config.Topology.Kinds[nodeCfg.Kind].Type != "":
+		return c.Config.Topology.Kinds[nodeCfg.Kind].Type
+	case c.Config.Topology.Defaults.Type != "":
+		return c.Config.Topology.Defaults.Type
+	}
+	// default type if not defined
+	switch nodeCfg.Kind {
+	case "srl":
+		return "ixr6"
```
steiler, line 242: This should maybe not be a hardcoded value but rather reference the srlTypes map.
My approach considered the first key in the map.

hellt: I dont think its a good idea. The map might be shuffled by someone without noticing that its values are used in something like default type

steiler: I just do not like arbitrary strings that much. I'd rather opt for an enum kind of type.
If someone messes with that string it breaks and if ixr6 is entirely removed from the list there is inconsistency as well.
But ok.

hellt: I made it a constant, so that this will be visible better 53c7b4c

hellt: btw, range over dict will yield inconsistent results, as map access is random

PR #217, opened 2021-01-07
added env vars configuration

### LINKED ISSUE #170: expose container Envs via nodes/kinds
Having fixed kinds can bring problems.
The env vars might change across releases, we must support ENV vars to be overridable.

hellt:
close #170

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -310,17 +310,10 @@ func (c *cLab) cmdInit(nodeCfg *NodeConfig, kind string) string {
 
 // initialize environment variables
 func (c *cLab) envInit(nodeCfg *NodeConfig, kind string) map[string]string {
-	switch {
-	case nodeCfg.Env != nil:
-		return nodeCfg.Env
-
-	case c.Config.Topology.Kinds[kind].Env != nil:
-		return c.Config.Topology.Kinds[kind].Env
-
-	case c.Config.Topology.Defaults.Env != nil:
-		return c.Config.Topology.Defaults.Env
-	}
-	return nil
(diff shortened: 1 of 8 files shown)


clab/config.go
```diff
@@ -307,6 +308,21 @@ func (c *cLab) cmdInit(nodeCfg *NodeConfig, kind string) string {
...
+
+	case c.Config.Topology.Kinds[kind].Env != nil:
+		return c.Config.Topology.Kinds[kind].Env
+
+	case c.Config.Topology.Defaults.Env != nil:
+		return c.Config.Topology.Defaults.Env
+	}
+	return nil
+}
+
```
karimra, line 325: as discussed, maybe allow merging of envs. node > kind > defaults

clab/config.go
```diff
@@ -569,3 +595,32 @@ func resolveBindPaths(binds []string) error {
...
+	m := make(map[string]string)
+	for k, v := range m1 {
+		m[k] = v
+	}
+
+	for k, v := range m2 {
+		m[k] = v
+	}
+	return m
+}
```
karimra, line 626: minor, but there is no need to create a new map, m2 can be merged into m1 and return m1.

hellt: yes, that was v1 of that function, but then I thought it would make sense to make this func a bit more generic for cases when modifying one of the maps is not desirable

karimra: ignore then

clab/config.go
hellt: added in cb223b2

PR #228, opened 2021-01-12
align srl config template with 20.6.3

hellt:
its not particularly great that we have some static template carried within the code for srlinux. When new features are added we must ensure that the CPM filters are added for those (if needed).
This was the case with OSPF now

It would be nice to think if there is a way to enable gnmi server, tls and lldp (these are the changes that containerlab adds on top of a default config) in addition to the config

/cc @karimra

### DIFF
--- a/templates/srl/srlconfig.tpl
+++ b/templates/srl/srlconfig.tpl
@@ -430,6 +430,44 @@
           },
           {
             "sequence-id": 290,
+            "description": "Accept incoming OSPF messages",
+            "action": {
+              "accept": {
+              }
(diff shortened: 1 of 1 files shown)


karimra:
The template is not static nor it is part of the code, nothing stops a user from replacing it with a another one they prefer.

hellt:
Yes, but that wasn't my point.

I wonder if it's possible to make Srl apply portions of config on top of a
default one.

The post deploy actions can be used for that, but I'd rather explore Srl
capabilities first

PR #229, opened 2021-01-13
added concurrent save and save for crpd + ceos

hellt:


hellt:
@karimra maybe you spot some wildly suboptimal steps in that approach?

PR #309, opened 2021-03-04
shell completions

### LINKED ISSUE #296: implement shell completions
for bash and zsh

steiler:
This is a very basic bash completion file. Which does not yet resolve any dynamic data.
There where discussions on keeping running topo files somewhere specific and so on, unfortunately I'm not on top of those discussions.
Hints on what / which parameters should be resolved dynamically are very welcome.

close #296

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -14,13 +14,13 @@ var completionCmd = &cobra.Command{
 
 Bash:
 
-  $ source <(yourprogram completion bash)
+  $ source <(containerlab completion bash)
 
   # To load completions for each session, execute once:
   # Linux:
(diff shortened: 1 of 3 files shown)


hellt:
Thanks for helping, @steiler 
Can you maybe expore the functionality that cobra offers - https://github.com/spf13/cobra/blob/master/shell_completions.md?
This promises to generate completions on-the-fly for many shells

steiler:
Alright, thanks for the hint.
Added the completion subcommand now.
However the dynamic file / folder and interface etc completion is to be introduced as well.

hellt:
that looks good, thanks @steiler 
I will later add a doc in this PR and afterwards merge it

hellt:
that looks good to me, merging

Issue #211, opened 2021-01-01
allow to create containers with the `host` networking 


hellt:
https://docs.docker.com/network/host/

I propose to expose this setting (string) on the node level only for now

```yaml
  nodes:
    telegraf:
      kind: linux
      network: host
```

the user-defined IP addresses and exposed ports will not have effect with host networking.

hellt:
added in #336

PR #336, opened 2021-03-17
allow for network mode host

steiler:
Referring to #211
Allow for the creation of containers in host mode.

### DIFF
--- a/clab/docker.go
+++ b/clab/docker.go
@@ -182,7 +182,7 @@ func (c *CLab) CreateContainer(ctx context.Context, node *Node) (err error) {
 	switch node.NetworkMode {
 	case "host":
 		containerHostConfig.NetworkMode = container.NetworkMode("host")
-	case "default":
+	default:
 		containerHostConfig.NetworkMode = container.NetworkMode(c.Config.Mgmt.Network)
 
 		containerNetworkingConfig.EndpointsConfig = map[string]*network.EndpointSettings{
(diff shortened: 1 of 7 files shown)


hellt:
I have refactored this piece a bit by making the `network-mode` field to be a selector of the network mode.

But a few things need to be added here

1. host-networking inherently prevents the node to have other links (such as veths), so we need to run a check of the topology like I do [here](https://github.com/srl-wim/container-lab/blob/0813cf8700284f414f2fd6aa2158e158aa8f396d/clab/config.go#L613) and fail early if the node which has `network-mode: host` also mentioned in the links section. I think one of the existing functions might already traverse the links section and there we can add this check as well
2. there is a notion of post-deploy checks - and there for linux node we need to skip the task of TX offload disablement as this is not permitted for host interfaces as it seems.

I used the following topo file:

```yaml
name: hostnet

topology:
  nodes:
    n1:
      kind: linux
      image: alpine:3
      network-mode: host
    n2:
      kind: linux
      image: alpine:3
```

and here is the log
image

clab/docker.go
```diff
@@ -177,9 +177,12 @@ func (c *CLab) CreateContainer(ctx context.Context, node *Node) (err error) {
...
 	}
 
-	containerNetworkingConfig := new(network.NetworkingConfig)
+	containerNetworkingConfig := &network.NetworkingConfig{}
 
-	if !node.NetworkModeHost {
+	switch node.NetworkMode {
+	case "host":
+		containerHostConfig.NetworkMode = container.NetworkMode("host")
+	case "default":
```
steiler, line 185: I'm confused here. Do you somewhere set NetworkMode to "default"?
Or was this meant to be  `default:` instead of `case "default:"` ?

hellt: sorry, it should read

```
	switch node.NetworkMode {
	case "host":
		containerHostConfig.NetworkMode = container.NetworkMode("host")
	default:
```

hellt:
thanks @steiler 
we are good to merge it now

PR #338, opened 2021-03-18
revised certificate handling

steiler:
- no new root cert/key pair is created if files exist
- node cert/key creation is skipped if cert/key pair exists (behaviour per node).


steiler:
Right now first it gets checked that there is a CA Cert present if not we create it, otherwise all good, anyways continue.
Then we check if node certs are already present (on a per node basis). If a nodes cert is not there we use the CA from step 1 to generate them. If it exists just use it.
I can imagine someone wants to only provide externally managed certs, so we would need to be able to disable the CA generation.
Whats your opinion here?

Other question:
If there is a CA cert present do we want to validate that the node certs are issued by the CA cert? Oder do we not care; there is a file available, go and use it.

hellt:
Thanks @steiler 
will check your PR tomorrow CET time

As to the user provided certificates I was thinking maybe to make use of node Labels for that. Something like that:

```yaml
nodes:
  node1:
    labels:
      node_cert_path: a/b/cert.pem
      node_priv_key_path: a/b/key
```

@karimra do you have a view on that?

karimra:
why is it a label, why not a field under the node ?

hellt:
I am struggling between bloating the node structure with many fields and making a right choice with this particular one

Maybe key/cert are generic enough to be promoted to all nodes structures as a field. whereas labels can be seen as a generic way to provide metadata to nodes

steiler:
I don't see why this needs to become a label or anything specifically defined, to be honest.
We are creating the folder structures anyhow and when creating certs they will be stored in that folder structure.
So whats the benefit of having the freedom to drop them $somewhere?

hellt:
valid claim
maybe a first start could be indeed with asking users to provide certs by the path in the lab directory

steiler:
That's exactly how it is implemented right now.
If certs exist use them, if not generate them. This is the case for CA as well as Node certs.

Haven't checked if the actual implementation does follow symlinks. But if it should be possible to store the certs somewhere else symlinking them in the scenario specific lab directory would be my advice.

hellt:
@steiler I have made a few changes in the last 3 commits, specifically I moved the logic of checking if rootCA is needed into the clab package, out from cmd package

hellt:
I think eventually we would need a flag in the topo file that will tell containerlab to generate node certificate/key for non-srl nodes.

The need is basically to create TLS files for other nodes which do not require them per se, but may use them for use cases which can leverage TLS certificates. I.e. we might want to have certs for ceos/vmx/xr nodes as we can import them via CLI/other_means and demonstrate/test how it works.

the question becomes to choose a right flag for a node config to signal that we want containerlab to generate certificates. I think this must be done outside of this PR

hellt:
another comment for history: it might still be desirable to make cert location configurable.
The reason is to make user-provided certs less challenging to configure. With the current approach if a user wants to make containerlab use custom certificates the user needs:

1. to know where containerlab will create lab directory with TLS files
2. create directories knowing all node names
3. copy files there

My biggest problem with step No2. If you change any node name you need to change the directory structure and that is irritating.
If we expose certificates location via topo file, the code will put those files over to the necessary directories on users' behalf.
Something that we can add later if there is a pressing need for that.

PR #352, opened 2021-03-26
Allow netconf save

kellerza:
Add a netconf save option for *vr-sros* containers

The netconf part works, but not 100% sure on the integration into clab yet, my local compiled containerlab executable cannot find any nodes any longer 😦 

```bash
$ sudo ../../containerlab/container-lab save -t ring1.yaml
INFO[0000] Getting topology information from ring1.yaml file...
Error: no containers found
```

Not 100% sure how this is related... at least it compiles, which is a step up from "hello world" 🚀


hellt:
Awesome, will have a look at it
Thank you for the contribution!

hellt:
Hey @kellerza 
I have added a few things to your PR:

1. used dns name instead of IP as clab creates /etc/hosts entries for each node
2. introduced a new map which will hold nodes credentials per node kind

kellerza:
@hellt this seems to be ok, initially tested against an existing deployment, but with the changes you mentioned was obviously breaking that. 

New deployments works well.

hellt:
moved the PR to #355 as some unexpected commits were added

kellerza:
Yeah, indeed, realised & fixed it. 

Typically suggestions on how to improve & telling people to go fix it goes a long way to build a community.

PR #360, opened 2021-04-02
Fixed etba drive to 4

burnyd:
Fixed the etba driver from env 1 to env 4.

### DIFF
--- a/clab/ceos.go
+++ b/clab/ceos.go
@@ -64,10 +64,10 @@ func initCeosNode(c *CLab, nodeCfg NodeConfig, node *Node, user string, envs map
 
 	// defined env vars for the ceos
 	kindEnv := map[string]string{
-		"CEOS":                                "4",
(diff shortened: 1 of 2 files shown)


clab/ceos.go
```diff
@@ -64,7 +64,7 @@ func initCeosNode(c *CLab, nodeCfg NodeConfig, node *Node, user string, envs map
 
 	// defined env vars for the ceos
 	kindEnv := map[string]string{
-		"CEOS":                                "1",
```
hellt, line 67: Hi @burnyd 
should you rather be changing `ETBA` env var here and not `CEOS`?

hellt: I mean why are you changing CEOS env var here in that map and not ETBA?

hellt: fixed it in 5e64f46

PR #363, opened 2021-04-05
Update ceos.cfg.tpl

dharmbhai:
cEOS-Lab requires Multimode Agent running for BGP instead of default RibD
Switching agents requires reboot
Hence requesting this be default as most of the work we do on cEOS-lab will be on EVPN/VxLAN
```
leaf1#conf
leaf1(config)#service routing protocols model ?
  multi-agent  Configure protocols model as multi-agent
  ribd         Configure protocols model as ribd (single agent)

leaf1(config)#service routing protocols model multi-agent
! Change will take effect only after switch reboot
leaf1(config)#
```

https://eos.arista.com/evpn-configuration-layer-2-evpn-design-with-type-2-routes/

### DIFF
--- a/templates/arista/ceos.cfg.tpl
+++ b/templates/arista/ceos.cfg.tpl
@@ -1,6 +1,8 @@
 hostname {{ .ShortName }}
 username admin privilege 15 secret admin
 !
+service routing protocols model multi-agent
+!
 interface Management0
 {{ if .MgmtIPv4Address }}ip address {{ .MgmtIPv4Address }}/{{.MgmtIPv4PrefixLength}}{{end}}
 {{ if .MgmtIPv6Address }}ipv6 address {{ .MgmtIPv6Address }}/{{.MgmtIPv6PrefixLength}}{{end}}


hellt:
thanks @dharmbhai 
let me ping some Arista engineers to validate the change request

@burnyd @noredistribution @ksator, guys, any objections to that change?

dharmbhai:
> thanks @dharmbhai
> let me ping some Arista engineers to validate the change request
> 
> @burnyd @noredistribution @ksator, guys, any objections to that change?

Hi @hellt 
I too work at Arista :-)
But there is no harm in reaching out to other more qualified than me :-)

burnyd:
This needs to be pushed within a startup-config that happens within the container mount.  Example..

nodes:
    spine1:
      kind: ceos
      image: coesimage:4.25.1F
      ports:
      - 4001:6030
      - 4444:443
      - 888:80 
      config: configs/spine1

# configs/spine1
! Startup-config last modified at  Thu Mar  4 20:57:52 2021 by root
! device: spine1 (cEOSLab, EOS-4.23.2F)
!
transceiver qsfp default-mode 4x10G
!
service routing protocols model multi-agent
!
hostname spine1
!

hellt:
I have a few "Go" received via different channels, so it looks like this change is quite _needed_. 
Thanks for your contribution @dharmbhai

ksator:
LGTM :-)

PR #367, opened 2021-04-06
Autocomplete Topo yaml files

kellerza:


hellt:
ah, so that is why when I hit `tab` after `deploy --topo` I didn't get a list of files?
where can I read about it more?

docs/cmd/completion.md
```diff
@@ -33,6 +33,10 @@ To load completions for each session, execute once:
 ```bash
 containerlab completion zsh > "${fpath[1]}/_containerlab"
 ```
+> Note: `$fpath[1]` in this command refers to the first path in `$fpath`. Ensure you use
```
hellt, line 36: can you change the blockquote style here to a theme supported info block that is enabled like this:

```
!!!info
    your text here
```

hellt (changes requested):

cmd/completion.go
```diff
@@ -40,19 +40,12 @@ fish:
...
-
-  # To load completions for every new session, run:
-  PS> containerlab completion powershell > containerlab.ps1
-  # and source this file from your PowerShell profile.
 `,
 	DisableFlagsInUseLine: true,
 	ValidArgs:             []string{"bash", "zsh", "fish"},
 	Args:                  cobra.ExactValidArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
+		cmd.Root().MarkFlagFilename("topo", "*.yaml", "*.yml")
```
hellt, line 48: I also think that it is best to add this line not in the completions file, but in the file where `topo` flag is defined

kellerza: ok, moved back to root.go

PR #368, opened 2021-04-06
Schema: Allow null if using defaults

kellerza:


hellt:
Hi @kellerza 
good catch! thank you!

hellt:
the clab files that errored before that change:

```yaml
name: test
topology:
  defaults:
    kind: srl
    image: srl:latest
  nodes:
    srl1:
    srl2:

```

Issue #317, opened 2021-03-05
allow CLI/SSH based configuration from a file referenced in topo


hellt:
Currently containerlab allows to mount full config files for some kinds (like srl).
But it is not yet possible to specify a file with CLI commands that would have been applied on the boot of the system.

A proposal is to allow referencing a file that contains CLI commands that should be executed on nodes SSH availability.

As an MVP the code should

1. distinct if the file mounted in `config` statement contains full config or CLI commands. The distinction can be made by the file extension. Maybe `*.cli` will be a reasonable  extension to notify containerlab that the file contains CLI commands, rather than full config.
2. once it is known that the config contains CLI commands the code should wait till SSH service is available on the default port
3. when ssh becomes available, the code should send the CLI commands over SSH channel
4. the username/password will be default ones as known to containerlab. If custom user/pwd needed, they should be handled via ENV vars (TBD)

/cc @jbemmel

hellt:
can be used - https://github.com/automatico/jato

hellt:
this can be closed as we how config engine in place
additional work will be done in different purpose-built PRs

PR #383, opened 2021-04-12
Config template support 🚀 

kellerza:
Configuration templates. 

- The configuration template system support roles, if no role specified, "role"="kind". Roles are used to find the template file. `<template>-<role>.tmpl`. A list of `<template>`s is specified with the `-l` parameter
- Template functions moved to external package, see functions [here](https://pkg.go.dev/github.com/kellerza/template#pkg-functions). All functions are chainable & where internal ones were duplicated it supports the original syntax and the chainable syntax
- This PR includes a basic template for vr-sros & SRL to configure Ports, Interfaces, IGP (ISIS) and SR on SROS
- The SSH deployer support SROS MD-CLI and SRL. Transaction support is automatic, except if the template starts with "show"
- The SSH deployer can be used for #317

## How does this work?

A clab.yml file with some additional labels. see [lab-examples/vr05](lab-examples/vr05)

```yaml
name: conf1
topology:
  defaults:
    kind: vr-sros
    image: registry.srlinux.dev/pub/vr-sros:21.2.R1
    license: /home/kellerza/containerlabs/license-sros21.txt
    labels:
      isis_iid: 0
  nodes:
    sr1:
      labels:
        systemip: 10.0.50.31/32
        sid_idx: 1
    sr2:
      labels:
        systemip: 10.0.50.32/32
        sid_idx: 2
    sr3:
      labels:
        systemip: 10.0.50.33/32
        sid_idx: 3
    sr4:
      labels:
        systemip: 10.0.50.34/32
        sid_idx: 4
  links:
    - endpoints: [sr1:eth1, sr2:eth2]
      labels:
        port: 1/1/c1/1, 1/1/c2/1
        ip: 1.1.1.2/30
        vlan: 99
    - endpoints: [sr2:eth1, sr3:eth2]
      labels:
        port: 1/1/c1/1, 1/1/c2/1
    - endpoints: [sr3:eth1, sr4:eth2]
      labels:
        port: 1/1/c1/1, 1/1/c2/1
    - endpoints: [sr4:eth1, sr1:eth2]
      labels:
        port: 1/1/c1/1, 1/1/c2/1
```

This lab can be deployed with the existing clab

```bash
./containerlab deploy -t conf1.clab.yml 
```

## Whats new?

The fun starts when all nodes are up and running :-)

```bash
./containerlab config -t conf1.clab.yml --template-path ./templates
```

Presto. An instant lab

SSH & run some show commands

```
A:admin@sr4# show router route-table

===============================================================================
Route Table (Router: Base)
===============================================================================
Dest Prefix[Flags]                            Type    Proto     Age        Pref
      Next Hop[Interface Name]                                    Metric
-------------------------------------------------------------------------------
1.1.1.0/30                                    Remote  ISIS      00h00m10s  18
       1.31.34.0                                                    20
1.31.34.0/31                                  Local   Local     00h00m20s  0
       to_sr1                                                       0
1.32.33.0/31                                  Remote  ISIS      00h00m09s  18
       1.33.34.0                                                    20
1.33.34.0/31                                  Local   Local     00h00m20s  0
       to_sr3                                                       0
10.0.50.31/32                                 Remote  ISIS      00h00m10s  18
       1.31.34.0                                                    10
10.0.50.32/32                                 Remote  ISIS      00h00m10s  18
       1.31.34.0                                                    20
10.0.50.33/32                                 Remote  ISIS      00h00m09s  18
       1.33.34.0                                                    10
10.0.50.34/32                                 Local   Local     00h00m20s  0
       system                                                       0
-------------------------------------------------------------------------------
No. of Routes: 8
Flags: n = Number of times nexthop is repeated
       B = BGP backup route available
       L = LFA nexthop available
       S = Sticky ECMP requested
===============================================================================
```

## How does this work?

The labels are variables that will be used in the configuration templates.. In the future this will be settings

Variables are prepared for the node. It will include an array for all links on the node. For each link, any string with a comma (e.g. a label of `port: 1/1/1, 1/1/2`) refers to the two endpoints of the link. In this case, the value for `port` will be `1/1/1` and `port_far` will be `1/1/2` for node A and swapped for node B.

Some fields have special generators, like link IPs. Specify only a single IP, the far end will be calculated. Specify nothing and an IP will be calculated from the systemip's

An example of prepared variables (you can also see this by adding `-dd` to the CLI

```
vars = {
      "isis_iid": "0",
      "links": [
            {
                  "ip": "1.50.51.1/31",
                  "ip_far": "1.50.51.0/31",
                  "isis_iid": "0",
                  "name": "to_srl",
                  "name_far": "to_sros",
                  "port": "1/1/c1/1",
                  "port_far": "ethernet-1/1",
                  "systemip": "10.0.50.51/32",
                  "systemip_far": "10.0.50.50/32",
                  "vlan": "10"
            }
      ],
      "role": "vr-sros",
      "sid_idx": "10",
      "systemip": "10.0.50.51/32"
  }
```

## Layering config templates
The current 'base' template might be a bit opinionated (ISIS, SR, RSVP-TE), but you can always use your own.

In general it is good to keep the config templates small, as they can be layered: `-l base,show-route-table` will configure the route table and then run the show commands

## Help

```bash
$ clab config --help
configure a lab based using templates and variables from the topology definition file
reference: https://containerlab.srlinux.dev/cmd/config/

Usage:
  containerlab config [flags]

Aliases:
  config, conf

Flags:
  -c, --check int               render dry-run & print n lines of config
  -h, --help                    help for config
  -p, --template-path string    directory with templates used to render config

Global Flags:
  -d, --debug count        enable debug mode
  -n, --name string        lab name
  -r, --runtime string     container runtime (default "docker")
      --timeout duration   timeout for docker requests, e.g: 30s, 1m, 2m30s (default 30s)
  -t, --topo string        path to the file with topology information
```


hellt:
Thanks @kellerza 
that is a nice step forward and a big chunk of work you did!

I think we won't be able to get away without making the architecture extensible. I wonder if you feel the same?

## Kind specific configs
First, I think we need to have a certain set of per-kind driver files which define the core configuration pieces of a kind. Maybe something similar to what scrapli does - https://github.com/scrapli/scrapli_community#adding-a-platform

With a per-kind config we can flexibly work with a session manager (i.e. ssh), without hardcoded `on-open`, `on-close`, `commit` commands. 

## SSH session config
I think with the above you will unlock the possibility to configure the session flexibly. Maybe you can look into https://github.com/networklore/netrasp to see how they handle session configuration and vendor specific configs

hellt:
Maybe @karimra can help us out here with architectural advise on how to pack nicely the interfaces to implement 
- pluggable transports
- pluggable drivers (aka kinds)

hellt:
@kellerza yes, I should have clarified that by pluggable I meant something that can read necessary inputs from a file (like different prompts, command sequences that are needed to be used on open and close of session)

I am all in to try and bring the sros/srl support for netrasp and reuse it's code, it is even more awesome. We/I can chat with Patrick (netrasp maintainer) about that.

In the meantime we can focus on templation architecture and discuss how to properly model configuration piece.
Maybe it is not a bad idea to explore a separate container under topo file to define configuration.

I'll hook up @hansthienpondt here as well as he might be interested in this

karimra:
For transport you could do something along the lines of:

```go
type Transport interface {
        // creates ssh connection and session
        // or creates a gNMI client and authenticates ( use capabilities to discover the kind ? )
        // or creates an http client for json-rpc
	Open(address string, opts ...opts) error
        // here the byte slice would be what is generated from the templates,
        // cli commands && write in case of ssh, 
        // json body in case of gNMI then sends a gNMI Set update to path `/`
        // json body or commands in case of json-rpc, then sends a set or cli RPC
	Write(data []byte) error
        // cleanup
	Close() error
}
```

Where opts can be used to pass params like auth method for ssh, dialOptions for gRPC, http headers for json-rpc.
And pass the kind to the transport.

an ssh implementation would look like:

```go
type sshTransport struct {
    // unexported config fields to customize the behavior; prompt, kind, etc
   c *ssh.Connection
   s *ssh.Session
}
```
gnmi implementation

```go
type gnmiTransport struct {
    // unexported fields to save the kind,...
    c       gnmi.GNMIClient
    cfn     context.CancelFunc
}
```

json-rpc implementation:

```go
type sshTransport struct {
   // unexported fields to save the kind,...
    c *http.Client
}
```

For Kinds, I had a discussion with @steiler last week about this, we reached this:

```go
type Kind interface{
    Initialize()
    Deploy()
    PostDeploy()
    Destroy()
}
```

Not sure how far this has gotten, and there was not much thought on the specific methods signatures.

steiler:
Haven't had the time up until now. It is quite some code that needs to be reorged. However I'm still on top of it.

kellerza:
> ```go
> type sshTransport struct {
>    // unexported fields to save the kind,...
>     c *http.Client
> }
> ```
> 
> ```go
> type Kind interface{
>     Initialize()
>     Deploy()
>     PostDeploy()
>     Destroy()
> }
> ```
> 

Thaks @karimra for an SSH session in this PR I have

```go
type Session struct {
	In      io.Reader
	Out     io.WriteCloser
	Session *ssh.Session
}
```

here we should probably allow an extendable kind using one of these SSH session, on an interface that adds the relevant `Connect`, `configure global`, `commit` actions 

```go
type KindSSH interface {
  func connect()
  func config()
  func write()
  func commit()
  func close()
}
```

karimra:
yes my answer was about to be your edit :)
your code can be reused as sshTransport for SRL.

BTW the Kind here is the Node kind (srl, ceos,...) not the transport kind.

hellt:
@kellerza what if the next step would be to make this common `Transport` interface and make your sshTransport to implement it?
What I would want to do next is to create `gnmic` transport and create a few templates that could be pushed over it to the nodes.

To me that would make an MVP for this one. It will deliver the extensibility of transports and will have two variants of templation ways (CLI and YAML)

hellt:
Another proposal from internal discussions

```yaml
config:
  vars:
    isis_id: 0
  transport: ssh
  templates:
    - bgp.yml
    - ifaces.yml
```

henderiw:
I believe we need something more flexible.

here is what I am using myself for now:

there is a dynamic case to build the infra -> it auto-assigns the AS, ISL subnets, Loopbacks from a pool, you can select single-stack, dual stack or v6-only; e/I/BGP, ISIS, OSPF

infrastructure:
  # dual-stack, ipv4-only, ipv6-only
  addressing_schema: "dual-stack"
  networks:
    loopback: {ipv4_cidr: 100.112.100.0/24, ipv6_cidr: 3100:100::/48}
    isl: {ipv4_cidr: 100.64.0.0/16, ipv6_cidr: 3100:64::/48, ipv4_itfce_prefix_length: 31, ipv6_itfce_prefix_length: 127}
  protocols:
    protocol: ebgp
    as_pool: [65000, 65100]
    overlay_as: 65002
    overlay_protocol: evpn

After you add labels to the links. Lag become interesting as you have to add some logical object to the link.

I did this so far.

links:
    # server connectivity
    - endpoints: ["leaf1:e1-1", "master0:eno5"]
      labels: {"kind": "access", "type": "esi1", "client-name": "bond0", "sriov": true, "ipvlan": true, "speed": "10G", "pxe": true}
    - endpoints: ["leaf2:e1-1", "master0:eno6"]
      labels: {"kind": "access", "type": "esi1", "client-name": "bond0", "sriov": true, "ipvlan": true, "speed": "10G"}
    - endpoints: ["leaf1:e1-2", "master0:ens2f0"]
      labels: {"kind": "access", "type": "esi2", "client-name": "bond1", "sriov": true, "ipvlan": true, "speed": "10G", "pxe": true}
    - endpoints: ["leaf2:e1-2", "master0:ens2f1"]
      labels: {"kind": "access", "type": "esi2", "client-name": "bond1", "sriov": true, "ipvlan": true, "speed": "10G"}

I also added a position the the node to know if this device is part of the infra or not.


topology:
  kinds:
    srl:
      type: ixrd2
      position: network
    vr-sros:
      type: sr-1s
      position: network
    linux:
      type: rhel8
      labels: {"need_update_dist": false, "install_rt_sched": false, "install_net_driver": true} 
      position: access
      storage:
        nfs_server: 100.112.1.201
        nfs_mount: /nfs
        csi: nfs-csi
  nodes:
    leaf1:
      kind: srl
      mgmt_ipv4: 172.20.20.3
      labels: {"target": "leaf-grp1"}      
    leaf2:
      kind: srl
      mgmt_ipv4: 172.20.20.4
      labels: {"target": "leaf-grp1"} 
    master0:
      kind: linux
      mgmt_ipv4: 100.112.3.11
      labels: {"target": "servers"}
    worker0:
      kind: linux
      mgmt_ipv4: 100.112.3.12
      labels: {"target": "servers"}
    worker1:
      kind: linux
      mgmt_ipv4: 100.112.3.13
      labels: {"target": "servers"}
    worker2:
      kind: linux
      mgmt_ipv4: 100.112.3.14
      labels: {"target": "servers"}
    worker3:
      kind: linux
      mgmt_ipv4: 100.112.3.15
      labels: {"target": "servers"}
    dcgw1:
      kind: sros
      mgmt_ipv4: 172.20.20.1
      labels: {"target": "dcgw-grp1"}
    dcgw2:
      kind: sros
      mgmt_ipv4: 172.20.20.2
      labels: {"target": "dcgw-grp1"}


After I have workloads which do the even/ipvn, etc.

workloads:
  provisioning:
    servers:
      vlans:
        itfce: {vlan_id: 0, kind: bridged}
  infrastructure:
    servers:
      vlans:
        itfce: {vlan_id: 40, kind: irb, ipv4_cidr: 100.112.3.11/24, ipv6_cidr: 2010:100:3::/64,}
    dcgw-grp1:
      vlans:
        itfce: {vlan_id: 45, kind: routed, ipv4_cidr: 10.100.40.0/24, ipv6_cidr: 2010:100:40::/48, ipv4_itfce_prefix_length: 31, ipv6_itfce_prefix_length: 127, addressing_schema: "dual-stack"}
  multus-mgmt:
    servers:
      vlans:
        ipvlan: {vlan_id: 101, kind: irb, ipv4_cidr: 10.1.11.0/24, ipv6_cidr: 2010:100:11::/64,}
        sriov1: {vlan_id: 102, kind: irb, ipv4_cidr: 10.1.12.0/24, ipv6_cidr: 2010:100:12::/64, target: leaf1}
        sriov2: {vlan_id: 103, kind: irb, ipv4_cidr: 10.1.13.0/24, ipv6_cidr: 2010:100:13::/64, target: leaf2}
      networks:
        loopback: {ipv4_cidr: 10.254.15.0/24}
    dcgw-grp1:
      vlans:
        itfce: {vlan_id: 105, kind: routed, ipv4_cidr: 10.100.15.0/24, ipv6_cidr: 2010:100:15::/48, ipv4_itfce_prefix_length: 31, ipv6_itfce_prefix_length: 127, addressing_schema: "dual-stack"}
  multus-internal:
    servers:
      vlans:
        ipvlan: {vlan_id: 201, kind: irb, ipv4_cidr: 10.1.21.0/24, ipv6_cidr: 2010:100:21::/64}
        sriov1: {vlan_id: 202, kind: irb, ipv4_cidr: 10.1.22.0/24, ipv6_cidr: 2010:100:22::/64, target: leaf1}
        sriov2: {vlan_id: 203, kind: irb, ipv4_cidr: 10.1.23.0/24, ipv6_cidr: 2010:100:23::/64, target: leaf2}
      networks:
        loopback: {ipv4_cidr: 10.254.25.0/24}
    dcgw-grp1:
      vlans:
        itfce: {vlan_id: 205, kind: routed, ipv4_cidr: 10.100.25.0/24, ipv6_cidr: 2010:100:25::/48, ipv4_itfce_prefix_length: 31, ipv6_itfce_prefix_length: 127, addressing_schema: "dual-stack"}
  multus-external:
    servers:
      vlans:
        ipvlan: {vlan_id: 301, kind: irb, ipv4_cidr: 10.1.31.0/24, ipv6_cidr: 2010:100:31::/64}
        sriov1: {vlan_id: 302, kind: irb, ipv4_cidr: 10.1.32.0/24, ipv6_cidr: 2010:100:32::/64, target: leaf1}
        sriov2: {vlan_id: 303, kind: irb, ipv4_cidr: 10.1.33.0/24, ipv6_cidr: 2010:100:33::/64, target: leaf2}
      networks:
       loopback: {ipv4_cidr: 10.254.35.0/24}
    dcgw-grp1:
      vlans:
        itfce: {vlan_id: 305, kind: routed, ipv4_cidr: 10.100.35.0/24, ipv6_cidr: 2010:100:35::/48, ipv4_itfce_prefix_length: 31, ipv6_itfce_prefix_length: 127, addressing_schema: "dual-stack"}
  multus-sba:
    servers:
      vlans:
        ipvlan: {vlan_id: 401, kind: irb, ipv4_cidr: 10.1.41.0/24, ipv6_cidr: 2010:100:41::/64}
        sriov1: {vlan_id: 402, kind: irb, ipv4_cidr: 10.1.42.0/24, ipv6_cidr: 2010:100:42::/64, target: leaf1}
        sriov2: {vlan_id: 403, kind: irb, ipv4_cidr: 10.1.43.0/24, ipv6_cidr: 2010:100:43::/64, target: leaf2}
      networks:
       loopback: {ipv4_cidr: 10.254.45.0/24}
    dcgw-grp1:
      vlans:
        itfce: {vlan_id: 405, kind: routed, ipv4_cidr: 10.100.45.0/24, ipv6_cidr: 2010:100:45::/48, ipv4_itfce_prefix_length: 31, ipv6_itfce_prefix_length: 127, addressing_schema: "dual-stack"}
  multus-internet:
    servers:
      vlans:
        ipvlan: {vlan_id: 501, kind: irb, ipv4_cidr: 10.1.51.0/24, ipv6_cidr: 2010:100:51::/64}
        sriov1: {vlan_id: 502, kind: irb, ipv4_cidr: 10.1.52.0/24, ipv6_cidr: 2010:100:52::/64, target: leaf1}
        sriov2: {vlan_id: 503, kind: irb, ipv4_cidr: 10.1.53.0/24, ipv6_cidr: 2010:100:53::/64, target: leaf2}
      networks:
        loopback: {ipv4_cidr: 10.254.55.0/24}
    dcgw-grp1:
      vlans:
        itfce: {vlan_id: 505, kind: routed, ipv4_cidr: 10.100.55.0/24, ipv6_cidr: 2010:100:55::/48, ipv4_itfce_prefix_length: 31, ipv6_itfce_prefix_length: 127, addressing_schema: "dual-stack"}


The result is a bunch of parameters you get that can feed into a template or any other system if we want to use this. it provides a lot of flexibility as I am using this for myself for a while and would benefit others.

kellerza:
@henderiw indeed having pools or lists for auto-assignment would be great and can look at adding some infrastructure and a template function to get some of this at some point

Do you keep lease state for assigned IPs? Its probably required if we want to keep this idempotent.. or will we be ok to have an IP change if there is a slight topology change for example?

kellerza:
Should be good to go. SROS MD-CLI and SRL works well

```bash
# View config
clab_ config --path $TEMPLATES0 --topo ring1.clab.yml --templates base -p 100
# Deploy config
clab_ config --path $TEMPLATES0 --topo ring1.clab.yml --templates base
# Show route tables
clab_ config --path $TEMPLATES0 --topo ring1.clab.yml --templates show-route-table
```

Converted the global `-d` to be a count, so you can have various level of debug to display sent/received from SSH etc
- `-d` normal
- `-dd` show sent/received and prompt parsing
- `-ddd` show internal channels debug
- `-dddd` add byte based printing to debug ssh&prompts

cmd/root.go
```diff
@@ -41,7 +43,7 @@ func Execute() {
 
 func init() {
 	rootCmd.SilenceUsage = true
-	rootCmd.PersistentFlags().BoolVarP(&debug, "debug", "d", false, "enable debug mode")
+	rootCmd.PersistentFlags().CountVarP(&debugCount, "debug", "d", "enable debug mode")
```
hellt, line 46: What do you say if we change CountVarP to an integer value which will correspond to debug level

Personally I am not a fan of `-xxxx` syntax, to me `-d 4` makes more sense and it is consistent with the long/short flags notation `--debug/-d`

hellt:
there seems to be an empty file `templates/vr-sros/show-route-table-link.tmpl`

cmd/root.go
kellerza: All this is provided by cobra/pFlags - [here](https://github.com/spf13/pflag/blob/master/count_test.go#L20)

supported today to give 2 as example: `-dd` / `-d -d` / `-d=2` / `--debug -d` / `--debug --debug` / `--debug=2`

Not working `-d 2` or `--debug 2` .. but believe this is a change that needs to happen to pFlag/Cobra upstream

The alternative is to make it an int, but that breaks the existing debug flag syntax

kellerza:
> there seems to be an empty file `templates/vr-sros/show-route-table-link.tmpl`

This is expected, since all templates needs an explicit -node and -link file. This show command is not applicable to links

The alternative is to silently ignore if not found/warn if debugging, but believe more explicit is better in this case (else bad filenames will hide issues).

cmd/root.go
hellt: ah, I didn't know that `-d=2` works. then its fine

cmd/config.go
```diff
@@ -0,0 +1,175 @@
...
+			return err
+		}
+
+		// config map per node. each node gets a couple of config snippets []string
+		allConfig := make(map[string][]config.ConfigSnippet)
+
+		renderErr := 0
+
+		for _, node := range c.Nodes {
+			kind := node.Labels["clab-node-kind"]
```
hellt, line 59: lets use `node.Kind` instead here?

cmd/config.go
```diff
@@ -0,0 +1,175 @@
...
+			clab.WithTopoFile(topo),
+			clab.WithEnvDockerClient(),
+		}
+		c := clab.NewContainerLab(opts...)
+
+		//ctx, cancel := context.WithCancel(context.Background())
+		//defer cancel()
+
+		setFlags(c.Config)
+		log.Debugf("lab Conf: %+v", c.Config)
```
hellt, line 47: while we are here, lets use

```
log.Debugf("Topology definition: %+v", c.Config)
```

just a housekeeping stuff

cmd/config.go
```diff
@@ -0,0 +1,175 @@
...
+			c.K = &config.SrlSshKind{}
+		}
+		return c, nil
+	}
+	return nil, fmt.Errorf("no tranport implemented for kind: %s", kind)
+}
+
+func init() {
+	rootCmd.AddCommand(configCmd)
+	configCmd.Flags().StringVarP(&templatePath, "path", "", "", "specify template path")
```
hellt, line 170: the `--path` flag name doesn't match in my head with the role of this flag

maybe use smth like `--templates-dir` with shorthand `-p` and a doc: `directory with templates used in configuration rendering`

cmd/config.go
```diff
@@ -0,0 +1,175 @@
...
+	configCmd.MarkFlagDirname("path")
+	configCmd.Flags().StringVarP(&config.TemplateOverride, "templates", "", "", "specify a list of template to apply")
```
hellt, line 172: what about adding shorthand `-n` and making the backing datastructure a slice of strings so multiple names can be provided?

The doc would be `comma separate list of template names to use`

cmd/config.go
```diff
@@ -0,0 +1,175 @@
...
+	configCmd.Flags().IntVarP(&printLines, "print-only", "p", 0, "print config, don't send it. Restricted to n lines")
```
hellt, line 173: maybe a more common `dry-run` name should be used?

cmd/config.go
```diff
@@ -0,0 +1,175 @@
...
+	configCmd.Flags().BoolVarP(&config.LoginMessages, "login-message", "", false, "show the SSH login message")
```
hellt, line 174: if this is only used as a debug thingy, maybe remove the flag and use debug level instead?

cmd/config.go
kellerza: used `--check`

clab/config/functions.go
```diff
@@ -0,0 +1,230 @@
...
+	switch val.(type) {
+	case string:
+		return "string"
+	case int, int16, int32:
+		return "int"
+	}
+	return ""
+}
+
+func hasInt(val interface{}) (int, bool) {
```
hellt, line 22: I wonder why the func is named `hasInt` and not `isInt`?

clab/config/functions.go
```diff
@@ -0,0 +1,230 @@
...
+	if i, err := strconv.Atoi(fmt.Sprintf("%v", val)); err == nil {
+		return i, true
+	}
+	return 0, false
+}
+
+func expectFunc(val interface{}, format string) (interface{}, error) {
```
hellt, line 29: can you please add a few doc lines for this func to explain what it does?

clab/config/functions.go
```diff
@@ -0,0 +1,230 @@
...
+	},
+	"default": func(in ...interface{}) (interface{}, error) {
+		if len(in) < 2 {
+			return nil, fmt.Errorf("default value expected")
+		}
+		if len(in) > 2 {
+			return nil, fmt.Errorf("too many arguments")
+		}
+
+		val := in[len(in)-1]
```
hellt, line 111: if `default` only takes in exactly two arguments, then it seems feasible to use `val := in[1]`?

clab/config/functions.go
```diff
@@ -0,0 +1,230 @@
...
+				start += len(v)
+			}
+			if end < 0 {
+				end += len(v)
+			}
+			return v[start:end], nil
+		}
+		return nil, fmt.Errorf("not an array")
+	},
+	"index": func(idx int, val interface{}) (interface{}, error) {
```
hellt, line 214: since `template` package has an `index` builtin function, maybe you need to rename this one to something else?
https://pkg.go.dev/text/template#hdr-Functions

clab/config/functions.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+			return strings.Join(res, sep), nil
+		case []string:
+			return strings.Join(v, sep), nil
+		case []int, []int16, []int32:
+			return strings.Trim(strings.ReplaceAll(fmt.Sprint(v), " ", sep), "[]"), nil
+		}
+		return nil, fmt.Errorf("expected array [], got %v", val)
+	},
+	"slice": func(start, end int, val interface{}) (interface{}, error) {
```
hellt, line 192: since template package has a `slice` builtin function, maybe you need to rename this one to something else?
https://pkg.go.dev/text/template#hdr-Functions

clab/config/transport.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+type Transport interface {
+	// Connect to the target host
+	Connect(host string) error
+	// Execute some config
+	Write(snip *ConfigSnippet) error
```
hellt, line 12: I think `Write()` shouldn't know about `ConfigSnippet`, it needs to write raw byte slice instead

clab/config/transport.go
```diff
@@ -0,0 +1,73 @@
+package config
```
hellt, line 1: wdyt @kellerza if we move the transport.go, ssh.go and sshkind.go to a `transport` package under the config dir? that should clean things up, but it will require some rework, since there will be some cycle dependencies at this point

cmd/config.go
```diff
@@ -0,0 +1,174 @@
...
+			return err
+		}
+
+		// config map per node. each node gets a couple of config snippets []string
+		allConfig := make(map[string][]config.ConfigSnippet)
+
+		renderErr := 0
+
+		for _, node := range c.Nodes {
+			err = config.LoadTemplate(node.Kind, templatePath)
```
hellt, line 60: shall we make the template load function to use node long name instead? using kind as a map key will limit us in having different templates for different nodes

clab/config/template.go
```diff
@@ -0,0 +1,203 @@
...
+
+func LoadTemplate(kind, templatePath string) error {
+	if templates == nil {
+		templates = make(map[string]*template.Template)
+	}
+	if _, ok := templates[kind]; ok {
+		return nil
+	}
+
+	tp := filepath.Join(templatePath, kind, "*.tmpl")
```
hellt, line 36: I think we shouldn't enforce dir structure by asking `kind` to be a subdirectory
IMHO it is better to just use `templatePath` and `*.tmpl`

kellerza: If the kind is not included you cannot build an interoperating lab. In the example I build connectivity & IGP between SR-OS and SRL.

Not having the kind in the path or filename means you'll need it in the template itself, pass a kind variable, but the template will become overly complex...

cmd/config.go
kellerza: But for the same kind of node you would want the same template? (un-rendered of course)

Maybe you can share an example of your intention?

clab/config/transport.go
kellerza: Let me have a look at this again, but believe from simple debugging etc the code currently uses much more than just the byte slice

clab/config/transport.go
kellerza: Can look into that, it should be fairly self-contained (except for ConfigSnippet, which might now have to become some sort of interface iso a struct)

clab/config/functions.go
kellerza: It can even be `parseInt` ;-)

clab/config/functions.go
kellerza: They are there, but their arguments all seem wrong, especially if you want to chain them:

> A pipeline may be "chained" by separating a sequence of commands with pipeline characters '|'. In a chained pipeline, the result of each command is passed as the last argument of the following command. The output of the final command in the pipeline is the value of the pipeline.

So my version of the functions all accept them as the LAST argument, the built-in all accept them as the FIRST argument

One of my test cases: `{{ "1.1.1.1/32" | split "/" | slice 1 2 | join "" }}` --> `32` [here](https://github.com/kellerza/containerlab/blob/config/clab/config/functions_test.go#L40)

Now try something similar in go template [here](https://play.golang.org/p/U54aSLXhjs8). There is a line you can uncomment to make it work...

clab/config/functions.go
kellerza: See the comment on slice. It seems as if they have their parameters does not work with pipelining... or I'm missing something?

clab/config/functions.go
hellt: how do you use this func in the templates? What inputs you feed to it?
 Just want to understand it better

kellerza: it receives any unknown type (`interface{}`) where we expect an `int`. So this could be a real `int` (mostly unlikely since our "labels" are strings) or it could be a `string` containing a number.

The function then returns this int and a bool to indicate if it was really an int

clab/config/functions.go
hellt: thanks, that is clear, and I don't object having custom funcs like that. What I am  proposing though is to avoid shadowing the embedded function names like `slice` and `index` and instead use some custom name

Users are in charge of writing the templates, and some of them will rely on `index` to behave the same way it behaves in stdlib, so overwriting this with custom functionality may be dangerous and frustrating.

What if we just give those custom func some distinctive names like `index-pipe`, `slice-pipe`?

clab/config/functions.go
hellt: gotcha, I left a comment there https://github.com/srl-labs/containerlab/pull/383#discussion_r634138965

clab/config/transport.go
hellt: yes, I see the reasoning behind propagating a richer struct as an input, but imho that makes writing transports harder as we drag the strcut ConfigSnippet to a transport which shouldn't really know anything about it

by making the transport work on primitive types like `[]byte` I believe we will have a cleaner code and distinctive packages separation.

cmd/config.go
hellt: no, I think templates are node-specific, not kind specific
i.e. in my lab with two sros nodes I might want R1 to have a certain template with, say, interfaces + bgp, whereas R2 will have just interfaces template

Do you see this differently?

kellerza: Guess I do see it a bit different. If you want BGP, you will have an additional label/var on that node that either indicates a BGP neighbor address/or the fact that it should peer toward the RR (the RR can be in your template or a global level var) etc. So the level of conditionality is built into the template, since you have to render all templates uniformly over the entire lab, else our command line is broken (we only allow 1 path and one set of template names). 

The example template has `sid_idx` which specifies Segment Routing, or if not there no segment routing if omitted. You can also specify `isis_iid` to change the IGP instance (and in a future commit skip ISIS). This is all part of the IGP part of the `base` template. The lab topology describe all the vars and the templates are responsible to change it into kind specific config. The command line to apply this across the entire lab (could be a mix of kinds) is simply: `clab config -t mylab.clab.yml -p /to/templates -n base`

clab/config/template.go
hellt: When I started to think about it more, that is the flow I got pictured in my head:

1. lets imagine the goal is the same -  to configure a basic ip connectivity between `srl1` and `sr1` nodes
2. In my view, the natural way of handling a task like that is to create a template (or a set of templates) and let the user to decide if the template(s) should be per-node, per-kind or global defaults.
3. by building on top of this idea, I thought that is how I would have done it:
    a user creates a template for srl named `~/srl.tmpl` and for sr `~/sr.tmpl`. At this moment the templates files belong to the same directory. 
2. then in the clab file the user can specify the paths to the templates which are used for a particular node, smth like this:
    ```yaml
    nodes:
      srl1:
        config:
          templates:
            - ~/srl.tmpl
    ```
     this definition will tell clab to use `~/srl.tmpl` for this node (and `sr.tmpl` for sros accordingly)
3. Note how here I pass the paths to the templates that should be used for this particular node. I could also create a `config` section on the kind level with the same path and that would have the effect of using the provided template for all SRL kinds.
3. This approach assumes that `--template-path` [flag](https://github.com/srl-labs/containerlab/pull/383/files/2a8df0115d1e284ea4889beec5f4f32fe9e8d8b3) is not really needed, as the paths to the templates are passed in the clab file
4. We can also add glob support in the provided paths, i.e. 
    ```yaml
    nodes:
      srl1:
        config:
          templates:
            - ~/srl*.tmpl
    ```
    and that will parse all the templates prefixed with `srl` string.


### Why?
Now to the why I think this is a more powerful approach.

1. we are not limiting users or dictate the rules as to how the structure of the template files should look like. They can keep the files anywhere they like and have the flexibility on selecting which templates will be in use
2. the clab file stores the paths to the templates and that makes configuration even easier. You don't need to have provide the template path in the cli, as this information is stored in the clab file itself and it is clear which template will be used for which nodes
3. by scoping the templates selection all the way to a per-node basis we have the best flexibility in templates application. Some nodes (even of the same kind) can use different templates if a user wants that

kellerza: But templates is not a fire-and-forget thing. It is incremental. Incremental enables us (slow) humans to test templates, capture output, etc. Templates in my vision was also more than config, it's a whole way of automating workflow across an entire lab (and this go implementation is lightning fast, almost like typing CLI!)

As an example I ran this lab last week:
```
# 1. Clear all config
# 2. Apply base - router/port/ISIS
# 3. apply seamless for ABRs, export system address etc
# 4.apply VPRN and set VPRN loopback address
clab_ config -t rrfwd2as.clab.yml -p $TEMPLATES1 -l delete,base,seamless,vprn

# Fix the ASBR
clab_ config -t rrfwd2as.clab.yml -p . -l rrfwd2as

# Show route table
clab_ config -t rrfwd2as.clab.yml -p $TEMPLATES1 -l show-route-table

# Hunt around
clab_ config -t rrfwd2as.clab.yml -p . -l show 

clab_ config -t rrfwd2as.clab.yml -p . -l set-rr-fwd

clab_ config -t rrfwd2as.clab.yml -p . -l show
```

As an example `delete` is a nice utility to make sure you captured all the changes in the layers of templates. Very powerful to restart this iterative process while building and testing my configs: `clab_ config -t rrfwd2as.clab.yml -p $TEMPLATES1 -l delete`

The `vr-sros/delete-node.tmpl` template contains:
```
delete /configure router
delete /configure policy-options
delete /configure service
```

I'll mail you the result as powerpoint for reference

hellt: fair point, it makes sense.
what then about this functionality -> currently you assume that the templates are aggregated after kinds. So that template files are kept under the `kind/template.tmpl` path.

I think in the future what would make sense is to use labels to tell the templating feature which base paths to use when configuring the nodes. Consider an example where you have a leaf-spine pop and you want to create templates for leafs and spines. Both of these network functions are of SRL kind, and that will make it impossible to create different templates for those nodes.
If, on the other hand, we say that templating engine can look for a base path that equals to some label, then it will make possible to flexibly create a folder structure as user sees fit.

In my example above, the file structure can be as follows:

```
.
├── config
│   ├── leaf
│   │   └── base.tmpl
│   └── spine
│       └── base.tmpl
└── mylab.clab.yml

3 directories, 3 files
```

and my clab file will have a label for each node
```yaml
leaf:
  kind: srl
  labels:
    role: leaf
spine:
  kind: srl
  labels:
    role: spine
```

and then `config` command will have a `--label role` flag to tell templating feature where to find the template files

wdyt?

clab/config/functions.go
hellt: got it. then to me `toInt` makes a nice name for that func

clab/config/functions.go
hellt: or maybe less verbose `_index` and `_slice` that will keep the original meaning but will not clash with the built-in functions

clab/config/template.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Strip blank lines
+	res := strings.Trim(buf.String(), "\n")
+	res = strings.ReplaceAll(res, "\n\n\n", "\n\n")
+	c.Data = []byte(res)
+
+	return nil
+}
+
+func RenderNode(node *types.Node) ([]ConfigSnippet, error) {
```
hellt, line 73: I wonder can we hugely improve the way we do rendering by actually merging the variable/labels of a node and the links that has a leg in this node

What i find quite complicated about the current approach is that we demand to have `*-node.tmpl` and `*-link.tmpl` and render two different templates per node.
If we merge the labels of a node and all its links, we will be able to just render the node and has a single template `$name.tmpl` to render.

I think this greatly improves the simplicity. What is your opinion on that?

clab/config/template.go
kellerza: you can easily do that today:

```
{{ if eq .role "leaf" }}
do leafy stuff
{{ end }}

{{ if eq .role "spine" }}
do spiny stuff
{{ end }}
```

What you are suggesting here and in the next comment is a contradiction ;-)

- here: Add another role based file resolution indirection 
- next: Remove a file resolution based indirection in favour of a var (we'll end up with `links := []map`)

btw, I like your suggestion below, since I end up doing many more empty -link templates than not.

clab/config/template.go
kellerza: I like this idea, it makes life simpler. (less indirection and rules to remember)

The only drawback I see in this approach is that you cannot incrementally apply `-node` and then individual `-links` (today I'll have 5 transactions if I have 1 node with 4 links)

but in the overall scheme of things (and for simplicity sake, its a good compromise).

clab/config/template.go
kellerza: If you really want the role base, I would suggest `/config/<kind>/base-<role>.tmpl`, but would prefer not to do that initially.

So this is a generalization of today's -node & -link (which I am going to remove now...)

kellerza: After some more thought, we could combine these ideas:

<pre>/templatePath/<b>{name}</b>-<b>{role or kind}</b>.tmpl</pre>

So this gives you:
(1) a flat template folder
(2) keeps the *kind* based config as it is today
(3) allow you to overwrite the default *kind* by specifying a *role*

The shows & debug is quite explicit in what it uses, so should be easy to follow.

The only trick now is if you have different devices AND want to use roles, you need to add the kind/some id. So a SROS `role: leaf` and a SRL `role: srleaf`, but that seem pretty acceptable.

So you can also see this as the default role being the kind.

hellt: > /templatePath/{name}-{role or kind}.tmpl

I like that. Maybe let's take this as an agreed file-search logic and let it out in the wild to see how this will be perceived and what are the limitations of it

clab/config/template.go
hellt: > The only drawback I see in this approach is that you cannot incrementally apply `-node` and then individual `-links`

I guess if your workflow needs separation between node and links portions, you would still be able to split the templates in two files and just ignore the nodes variable for link template (and vice versa)

kellerza:
Ready for review. Top comment updated.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
+repos:
```
hellt, line 1: Hi @kellerza 
what is this yaml for?

kellerza: Hi Roman, it is a framework to install git pre-commit hooks (https://pre-commit.com/) that performs a bunch of simple checks before accepting a commit.

Might duplicate some of the other CI checks/tasks you have on github, but codespell really helps for people (myself included) who are too dependent on just typing and Office correcting

hellt:
thanks @kellerza 
we will continue refining the ideas of config provisioning here https://github.com/srl-labs/containerlab/discussions/487

Issue #387, opened 2021-04-16
check that across bridge interfaces there are no duplicates


hellt:
the following links definition will fail at runtime, since two bridges reference the same interface `eth1`.

```yaml
# br1 and br2 are bridges
links:
- endpoints: ["br1:eth1", "spine1:e1-1"]
- endpoints: ["br2:eth1", "spine2:e1-1"]
```

A check is needed to get all interfaces of all bridges and check that none of them repeat

PR #400, opened 2021-04-22
added checks for duplicate interface names across bridges and per con…

steiler:
…tainer nodes

implements checks as per #387.
Further checks interfaces per node are unique.


hellt:
thanks @steiler 
I will have a look

PS> discard Deepsource failures

PR #401, opened 2021-04-22
Unique rootns ifaces

### LINKED ISSUE #387: check that across bridge interfaces there are no duplicates

hellt:
ref #400 

close #387 

Hi @steiler 
I had a look at #400 and I propose we change it as per this PR.

There are two small issues that I noticed in your PR

1. we need to account for all kinds which interfaces appear in root netns: `bridge`, `ovs-bridge` and `host` kinds.
2. we don't need to check for uniqueness of container interfaces, as it has already been checked [here](https://github.com/srl-labs/containerlab/blob/cb400ef37f867e13d0132e5cd83041f147455ef1/clab/config.go#L690)

So I cut out the pieces that I don't think are needed here.
Can you check this PR and see if I missed something imoportant?

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -672,6 +672,9 @@ func (c *CLab) CheckTopologyDefinition(ctx context.Context) error {
 	if err := c.VerifyImages(ctx); err != nil {
 		return err
 	}
+	if err := c.VerifyInterfaceUniqueness(); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


steiler:
Looks good to me.
I have just had a check for host-mode in it at the beginning but that is not adding any interfaces.
But ovs-bridge and host kinds seem to me to be right to be checked here.

PR #410, opened 2021-04-26
added $nodeDir as a variable parameter when binding files or dirs via *.clab.yml

### LINKED ISSUE #398: introduce path variable `$nodeDir` that will expand to abs path of the node directory
To support the use case expressed in #394 a path variable such as `$nodeDir` can be implemented to allow the lookup of the files located at node's directory.

This will make possible the following binds configuration:

```yaml
name: test
topology:
  nodes:
    leaf1:
      kind: linux
      image: cx:latest
      binds:
      - $nodeDir/daemons:/etc/frr/daemons # expands to ./clab-test-leaf1/daemons
      - $nodeDir/frr.conf:/etc/frr/frr.conf
    leaf2:
      kind: linux
      image: cx:latest
      binds:
      - $nodeDir/daemons:/etc/frr/daemons # expands to ./clab-test-leaf2/daemons
      - $nodeDir/frr.conf:/etc/frr/frr.conf
```

steiler:
Proposed Implementation for #398.

Are we file with using `node.LabDir` or should this be a different directory, maybe even a sub-directory, anything?

close #398

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -480,7 +480,7 @@ func (c *CLab) NewNode(nodeName string, nodeCfg NodeConfig, idx int) error {
 
 	// initialize bind mounts
 	binds := c.bindsInit(&nodeCfg)
-	err := resolveBindPaths(binds)
+	err := resolveBindPaths(binds, node.LabDir)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


clab/test_data/topo8-bind-nodedir.yml
```diff
@@ -0,0 +1,11 @@
+name: topo8
```
hellt, line 1: this file seems not being used in any way?

hellt:
> Are we file with using `node.LabDir` or should this be a different directory, maybe even a sub-directory, anything?

the idea was specifically to use node directory under lab directory, so that users can put files on a per-node basis in the directories that clab would have created automatically on the first run

clab/test_data/topo8-bind-nodedir.yml
steiler: You're right, my first intention was to test this also via the TestBindsInit() method.
Changed my mind along the path. Will remove it!

hellt:
I was thinking that it would be nice to add a customization of the node dir path. 
By default it uses the nodedir under the clab dir, but I can imagine that it would be nice to add the other locations as well

For example I often keep node specific files under some file dir on the same level as the clab file. So maybe we can add an env var NODE_DIR that would overwrite the default location of the nodedir if set

steiler:
So how are we going about with this pull request now?
What you are describing is another enhancement @hellt, right?

hellt:
yes, we can tackle this in another PR later, if there will be interest in this
I still need to test this one. Plan to do this today

hellt:
thanks @steiler  for the initial work on that one
I added test cleanup and docs in the latest two commits, and will merge it as it is

Issue #412, opened 2021-04-29
clab graph shoudn't be run without --topo flag


blinklet:
The "graph" command does not display anything.

To recreate, run a lab example on your local machine. I am running Containerlab on a laptop running Ubuntu 20.04 desktop. I used _lab-examples/srl02_. After the lab is running, run the `sudo clab graph` command. Then, open a browser window at the URL `http://localhost:50080/`. I see the display shown in the attached image. It just shows the page title with no lab information.

![graph-problem](https://user-images.githubusercontent.com/11431390/116573663-016a9b00-a8db-11eb-9a46-ee75feee1cfc.png)

However, if I run the graph command on the topology file, it displays the lab information. The command ` sudo clab graph --offline --topo srl02.clab.yml` results in the display below:
![graph-prob-2](https://user-images.githubusercontent.com/11431390/116576161-4ee80780-a8dd-11eb-98e5-db22fa76c913.png)

This is a small problem but I though you'd like to know.

hellt:
thanks,
it seems reasonable to not allow `clab graph` to be run without provided topo file

PR #413, opened 2021-04-30
check topology file is provided on graph command

steiler:
Implementing #412

### DIFF
--- a/cmd/graph.go
+++ b/cmd/graph.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"html/template"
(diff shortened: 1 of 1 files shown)


cmd/graph.go
```diff
@@ -47,6 +48,12 @@ var graphCmd = &cobra.Command{
 	Long:  "generate topology graph based on the topology definition file and running containers\nreference: https://containerlab.srlinux.dev/cmd/graph/",
 
 	RunE: func(cmd *cobra.Command, args []string) error {
+
```
hellt, line 51: I wonder why by adding `WithTopoFile` clab option we haven't stoped the graphing if the file is not present?
This option should have checked if the file is present

PR #414, opened 2021-05-02
import ceos environment variables into container for run time application

sulrich:
## overview

this PR builds the node.Cmd string to import the merged environment variables from the into the container execution environment.  this allows the correct application and import of the topology/node-specific overrides.  notably, this allows for setting the `INTFTYPE` to provide alignment with the interfaces that ceos uses internally for various features.  

this also updates `docs/manual/kinds/ceos.md` to include text outlining the potential need for some users to adopt the `et` naming and provides an example for application in the topology file.

### DIFF
--- a/docs/manual/kinds/ceos.md
+++ b/docs/manual/kinds/ceos.md
@@ -39,13 +39,24 @@ Arista cEOS node launched with containerlab can be managed via the following int
 !!!info
     Default user credentials: `admin:admin`
 
+## Interface naming and link definition
+
+When defining links in the topology, ceos container interfaces should be named `et` as opposed to `eth`.  This provides consistency with the underlying interface mappings and the interactions with some features internally to ceos.  This naming requirement does not apply to the `eth0` interface created by containerlab only to links that are used for interconnection with other elements in a topology.
+
+example:
+```yml
+  links:
+    - endpoints: ["ceos_rtr1:et1", "ceos_rtr2:et1"]
+    - endpoints: ["ceos_rtr1:et2", "ceos_rtr3:et1"]
+```
+
 ## Interfaces mapping
 ceos container uses the following mapping for its linux interfaces:
 
 * `eth0` - management interface connected to the containerlab management network
-* `eth1` - first data interface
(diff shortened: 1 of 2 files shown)


docs/manual/kinds/ceos.md
```diff
@@ -83,6 +83,26 @@ When containerlab launches ceos node, it will set IPv4/6 addresses as assigned b
     ```
     As you see, the management interface `Ma0` inherits the IP address that docker assigned to ceos container management interface.
 
+
+## Additional Interface naming considerations
+
+While many users will be fine with the default ceos naming of `eth`, some ceos users may find that they need to name their interfaces `et`. Interfaces named `et` provide consistency with the underlying interface mappings within ceos. This enables the correct operation of commands/features which depend on `et` format interface naming.
+
+In order to align interfaces in this manner, the `INTFTYPE` environment variable must be set to `et` in the topology definition file and the links which are defined must be named `et`, as opposed to `eth`. This naming requirement does not apply to the `eth0 interface automatically created by containerlab. This is only required for links that are used for interconnection with other elements in a topology.
```
hellt, line 91: on this line the `eth0` word is not wrapped with the closing backtick, which caused formatting issues

docs/manual/kinds/ceos.md
```diff
@@ -83,6 +83,26 @@ When containerlab launches ceos node, it will set IPv4/6 addresses as assigned b
     ```
     As you see, the management interface `Ma0` inherits the IP address that docker assigned to ceos container management interface.
 
+
+## Additional Interface naming considerations
```
hellt, line 87: do you mind to lowercase the Interface word?

hellt (changes requested):

docs/manual/kinds/ceos.md
sulrich: ack, fixed.

docs/manual/kinds/ceos.md
sulrich: sure thing.  done.

Issue #415, opened 2021-05-04
Code re-organisation


networkop:
Hi 👋 , I'm thinking about extending clab beyond docker to other container runtimes, .e.g. containerd. Reading through the code I've come across a few issues that would make this difficult, so I was going to run a few suggestions by yall first , to see if you're interested in this. The main culprit is the `/clab` directory that tightly integrates container runtime with the rest of the code from this directory so I thought about a few steps how it can be disentangled:

1. Take all of general utility functions, mostly found in `file.go`, `ethtool.go`, maybe `cert.go` and move them into a separate package, e.g. `utils/`
2. Create a new `runtime` package with interface type `ContainerRuntime` that would have a set of common functions , e.g. `CreateContainer`, `ListContainers` and move all Docker API functions behind this interface.

wdyt?

hellt:
Hi @networkop 
That is a good initiative, and something I be been asked of a few times. I can see how both podman and containerd can make a perfect fit for this. 

That is, though, quite a chunk of work, mostly ensuring that the existing functionality works after the split. But I'm willing to dive in with you in this endeavor. 

There's also another piece of decoupling that we were postponing for quite a while, which is the interface based kinds. But this can be done before this I hope

@karimra

networkop:
yeah, i was gonna suggest interfaces for kinds, but then i thought that'd be too much for a single issue. 
anyhow, I think these pieces of work can all be done independently.

hellt:
thanks @networkop for this ginormous contribution!
delivered in #416

PR #416, opened 2021-05-05
Moved docker behind a new runtime interface

networkop:
This is some quick prototyping I've done to kick-start the discussion raised in #415. Most of the code logic is still the same, the only difference is all docker api interactions are hidden behind a generic container runtime interface. 
There are 3 new folders:
1. `./runtime` - contains interface definition and docker runtime
2. `./types` - generic types that are required by both clab and runtimes (moved some types from `config.go` to avoid circular imports)
3. `./utils` - again some generic util functions 

Have a look, let me know what you think 😄


hellt:
thanks @networkop
basic integration tests for ceos and srl have passed as well - https://gitlab.com/rdodin/containerlab/-/pipelines/298197975

I wonder how you imagine the runtime selection needs to be handled?

networkop:
@hellt there's a new CLI argument `--runtime` that defaults to docker. check out `./cmd/root.go`. that's how pretty much most of the tools I'd seen are working. auto-sensing is too laborious to implement and maintain.

networkop:
latest commit was tested on docker-ce 19.03.15.

cmd/deploy.go
```diff
@@ -260,11 +255,13 @@ func enrichNodes(containers []types.Container, nodes map[string]*clab.Node, mgmt
...
-			node.MgmtNet = mgmtNet
-			node.MgmtIPv4Address = c.NetworkSettings.Networks[mgmtNet].IPAddress
-			node.MgmtIPv4PrefixLength = c.NetworkSettings.Networks[mgmtNet].IPPrefixLen
-			node.MgmtIPv6Address = c.NetworkSettings.Networks[mgmtNet].GlobalIPv6Address
-			node.MgmtIPv6PrefixLength = c.NetworkSettings.Networks[mgmtNet].GlobalIPv6PrefixLen
+
+			if c.NetworkSettings.Set {
+				node.MgmtIPv4Address = c.NetworkSettings.IPv4addr
+				node.MgmtIPv4PrefixLength = c.NetworkSettings.IPv4pLen
+				node.MgmtIPv6Address = c.NetworkSettings.IPv4addr
```
hellt, line 262: should be IPv6 address here I suppose

clab/clab.go
```diff
@@ -10,19 +10,20 @@ import (
...
-	m            *sync.RWMutex
-	Nodes        map[string]*Node
-	Links        map[int]*Link
-	DockerClient *docker.Client
-	Dir          *Directory
+	Config   *Config
+	TopoFile *TopoFile
+	m        *sync.RWMutex
+	Nodes    map[string]*types.Node
+	Links    map[int]*Link
```
hellt, line 24: if we move types to their own package, should the `Link` type be moved as well?

clab/clab.go
```diff
@@ -83,16 +81,48 @@ func NewContainerLab(opts ...ClabOption) *CLab {
...
+		if c.Config.Mgmt.IPv4Subnet == "" {
+			c.Config.Mgmt.IPv4Subnet = dockerNetIPv4Addr
+		}
+		if c.Config.Mgmt.IPv6Subnet == "" {
+			c.Config.Mgmt.IPv6Subnet = dockerNetIPv6Addr
+		}
+	}
+	// init docker network mtu
+	if c.Config.Mgmt.MTU == "" {
+		m, err := getDefaultDockerMTU()
```
hellt, line 115: likely this func needs to be runtime-scoped? i.e. with containerd/podman the logic should be runtime-specific as `docker0` interface won't be there

clab/clab.go
networkop: yeah, perhaps. I was trying to minimize the number of changes. if you think it makes sense as a part of this PR, i'll move it.

cmd/deploy.go
networkop: hm.. yeah..

clab/clab.go
networkop: so should we move this whole function under the runtime interface? or do you want to just skip MTU here and set it in the runtime?

networkop: tbh, I don't fully understand the need to initMgmtNetwork. why can't we just defer to the runtime to set it all up (after all it's managing it's own IPAM). maybe you can elaborate @hellt ?

hellt: I think we can start with a small change of adding a new runtime-specific function named `DefaultMgmtNetMTU` which will return the default management network MTU using the runtime-specific methods.

cmd/deploy.go
```diff
@@ -225,33 +226,27 @@ func createHostsFile(containers []types.Container, bridgeName string) error {
...
-			}
+		if cont.NetworkSettings.Set {
+			buff.WriteString(cont.NetworkSettings.IPv4addr)
+			buff.WriteString("\t")
+			buff.WriteString(strings.TrimLeft(cont.Names[0], "/"))
+			buff.WriteString("\n")
+			buff.WriteString(cont.NetworkSettings.IPv6addr)
+			buff.WriteString("\t")
+			buff.WriteString(strings.TrimLeft(cont.Names[0], "/"))
+			buff.WriteString("\n")
```
hellt, line 243: there might be a situation where IPv6 addresses are not set while IPv4 addresses are (and vice versa)

for example, if you use the following mgmt config:

```yaml
name: ipv4only
mgmt:
  ipv4_subnet: 172.30.30.0/24
```

then your nodes will be addressed with IPv4 addresses only from the specified range. At the same time, the current function will add hosts entires for empty ipv6 addresses, it seems

clab/clab.go
networkop: fixed in f000c0f

clab/clab.go
hellt: @networkop the `initMgmtNetwork` was created to set the default configuration values for mgmtNetName (clab) and default subnets for ipv4/6

It might be that this can be removed/changed with the default values to be passed over to the respective runtime func that creates the network

hellt:
@networkop can we address this deprecation issue as well? https://github.com/srl-labs/containerlab/pull/416/checks?check_run_id=2545109898#step:4:26

clab/clab.go
networkop: right, this makes sense. but this `getDefaultDockerMTU()` creates circular dependency between the runtime and the init function: you're building values to configure runtime while at the same time relying on runtime to provide some of these values.
In particular, let's assume we move this behind a runtime interface, we would then rely on runtime being set up before we can call initMgmt function, which looks like a scope creep -- we'd need to modify all the config parsing tests to init a runtime in order to set the MTU value.

runtime/runtime.go
```diff
@@ -0,0 +1,58 @@
...
+	// Pull container image if not present
+	PullImageIfRequired(context.Context, string) error
+	// Create container
+	CreateContainer(context.Context, *types.Node) error
+	// Start pre-created container by its name
+	StartContainer(context.Context, string) error
+	// Stop running container by its name
+	StopContainer(context.Context, string, *time.Duration) error
+	// Create a creates a symlink for containers network namespace so that it can be managed by iproute2 utility
+	LinkContainerNS(string, string) error
```
hellt, line 33: wdyt about removing this from runtime signature? this looks to me more like an utils function that just acts on a container name, but the runtime details are not in the scope of this function

cmd/deploy.go
networkop: I don't fully understand this. Is this determined inside the`initMgmtNetwork` ?

```
	if c.Config.Mgmt.IPv4Subnet == "" && c.Config.Mgmt.IPv6Subnet == "" {
		if c.Config.Mgmt.IPv4Subnet == "" {
			c.Config.Mgmt.IPv4Subnet = dockerNetIPv4Addr
		}
		if c.Config.Mgmt.IPv6Subnet == "" {
			c.Config.Mgmt.IPv6Subnet = dockerNetIPv6Addr
		}
	}
```

runtime/runtime.go
networkop: yep, makes sense

clab/clab.go
hellt: then maybe has an util function that takes the runtime name and returns the default MTU for a given runtime value?
i.e. for docker it will use the existing func that gets the mtu of a `docker0` iface

networkop:
> @networkop can we address this deprecation issue as well? https://github.com/srl-labs/containerlab/pull/416/checks?check_run_id=2545109898#step:4:26

This is a transitive dependency, I'll just add this to go.mod, i think that should fix it?
```
replace golang.org/x/crypto/ssh/terminal => golang.org/x/term v0.0.0-20210503060354-a79de5458b56
```

clab/clab.go
networkop: yeah, that might work. let me see

networkop: actually, it will have the same problem, i.e. you need to know the runtime name when you populate your default values.

networkop: at this stage, i think the only right solution (I think) is to leave MTU unchanged and do a check (if empty, then lookup) inside the `CreateNet` function, which is the only place where MTU is used.

hellt: but the runtime name is always provided via `rt` value, and has a `docker` default, no?

cmd/deploy.go
networkop: but that's exactly how this function works now. TBH, the extra if can be dropped as it's duplicate:
```
	if c.Config.Mgmt.IPv4Subnet == "" && c.Config.Mgmt.IPv6Subnet == "" {
		c.Config.Mgmt.IPv4Subnet = dockerNetIPv4Addr
		c.Config.Mgmt.IPv6Subnet = dockerNetIPv6Addr
	} 
```
That would set the default values only when both v4 and v6 subnets are missing. In all other cases, they will be populated with user-provided values.

Is this a purely incorrect reporting bug?

clab/clab.go
networkop: yes, there's always a runtime set. The problem is that runtime is currently optional and most of the `config_test.go` functions are set up just `WithTopoFile`. It's not a problem adding a `WithRuntime` to them as well, it'll work, but it is an indication of leaking dependencies -- your config test code should not really rely on a runtime being set.
But if you're ok with this, I'm happy to add runtime to every test in `config_test.go`, so they'll all look like 👍 
```diff
	for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
			opts := []ClabOption{
				WithTopoFile(tc.got),
+                              WithRuntime(...),
			}
```

cmd/deploy.go
hellt: @networkop the culprit was in moving `initMgmtNetwork` from `ParseTopology` to `NewContainerlab` here - https://github.com/srl-labs/containerlab/pull/416/files#diff-831ecb8143ceda28e72434a64e3b7f88d15f9492071603608fd0180ae3105e67R88-R92

it led to reading the `mgmt` config element before it was parsed from the clab file


Let's move it back to `ParseTopology` as it [was](https://github.com/srl-labs/containerlab/blob/169c8bdb33d78a6bb5006d568dbc396561af8eb9/clab/config.go#L229) before?

clab/clab.go
hellt: maybe we should defer this till we sort the other things

cmd/deploy.go
networkop: interesting. the reason why I moved it out was what we're discussing above - interaction with runtime. Let's assume I put it back into ParseTopology, I would then have to call `runtime.SetMgmtNet` with `c.Config.Mgmt`.  I can't call it inside `WithRuntime` because topology hasn't been parsed yet, so the only place to do that is inside each of the `./cmd/<action>.go` files, similar to this:

```
if err = c.ParseTopology(); err != nil {
	return err
}
c.Runtime.SetMgmtNet(c.Config.Mgmt)
```

This didn't look right to me at the time, as I was trying to minimise the impact of the change. Does this look ok to you?

clab/clab.go
networkop: another issue with getting default MTU is determining which bridge to use. Some runtimes may defer to CNI plugins to setup the networking, which means we'd have to parse all CNI configuration files in order to extract the name of the bridge, and it may not even exist in case of a P2P CNI plugin. So maybe we should just have the magic `getDefaultMTU` function that looks picks the lowest MTU off several well-known bridge names, and, if none found, falls back to some safe default, e.g. 1400. wdyt @hellt

cmd/deploy.go
hellt: alternatively we might put the logic that sets the management network defaults inside the `WithTopoFile` func, it reads the clab yaml file and then sets the necessary defaults

```diff
func WithTopoFile(file string) ClabOption {
	return func(c *CLab) {
		if file == "" {
			return
		}
		if err := c.GetTopology(file); err != nil {
			log.Fatalf("failed to read topology file: %v", err)
		}

+		err := c.initMgmtNetwork()
+		if err != nil {
+			log.Fatalf("initMgmtNetwork: %s", err)
+		}
	}
}
```

It is not "clean" logic-wise, but seems to do the job

networkop: cool. check the latest PR

go.mod
```diff
@@ -26,3 +26,5 @@ require (
 	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
 	gopkg.in/yaml.v2 v2.4.0
 )
+
+replace golang.org/x/crypto/ssh/terminal => golang.org/x/term v0.0.0-20210503060354-a79de5458b56
```
hellt, line 30: this hasn't fixed this it seems https://github.com/srl-labs/containerlab/pull/416/checks?check_run_id=2546702135#step:4:26

hellt: why not swapping the pkg import to x/term though?

cmd/deploy.go
hellt: coming back to the original issue using the latest commit:

the current code for writing host entries doesn't account for the fact that ipv6/4 address might be empty, this leads to the following lines created when only ipv4_subnet is specified:

```
172.20.30.2	clab-srl01-srl
	clab-srl01-srl
```

the original code in maste checks first if the ipv4/6 address is not `""` and then writes the host entry, this I believe should be used here as well to omit the entries which do not have ipv4/6 address

go.mod
networkop: it's an external dependency, from mysocket-io

hellt: It's not. I don't use mysocket API, rather shelling out and using it's api
Here is the term thing I use to capture the password from the terminal 

https://github.com/srl-labs/containerlab/blob/169c8bdb33d78a6bb5006d568dbc396561af8eb9/cmd/mysocketio.go#L14

cmd/inspect.go
```diff
@@ -194,38 +194,18 @@ func printContainerInspect(c *clab.CLab, containers []types.Container, bridgeNam
...
-		}
-	}
-	for _, br := range container.NetworkSettings.Networks {
-		if br.GlobalIPv6Address == "" {
-			return "NA"
-		}
-		return fmt.Sprintf("%s/%d", br.GlobalIPv6Address, br.GlobalIPv6PrefixLen)
-	}
-	return ""
+	return fmt.Sprintf("%s/%d", ctr.NetworkSettings.IPv6addr, ctr.NetworkSettings.IPv6pLen)
```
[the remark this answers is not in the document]
networkop: so the latest commit does return NA for when v6 address is missing. but it appears that v4 address is assigned always, so it gets reported even when there's only `ipv6_subnet` specified.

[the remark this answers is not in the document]
hellt: @networkop, yes, this is fine. We can look later on if we need to have ipv6 only networks.
Your last commit brings the parity with `v0.13.0`, and that was what I wanted to achieve in the first place

In my opinion, the only thing that currently left is https://github.com/srl-labs/containerlab/pull/416#discussion_r629439864
once this is resolved, I will do some tests which are not in the CI and if all goes well, we will have a green light

cmd/deploy.go
networkop: oh yeah, must have missed that.

PR #417, opened 2021-05-05
FFRouting lab 01

blinklet:
This is a simple OSPF lab example using FRRouting containers in a ring topology. I created it to demonstrate how to create a lab of "linux"-type nodes that start with a working configuration prepared by the lab developer. 

Note that I am using the frrouting:7.5.1 tag for the routers, instead of the latest tag. There is currently a major bug in frrouting:latest that prevents the systems from booting. 

For your consideration.


hellt:
Hi @blinklet 
I have added a doc site entry for you contribution and made a few slight changes. Let me know if this looks good to you.

PS> you can run a web server with local changes with a single command:

```
docker run --rm -d  -p 8000:8000 -v $(pwd):/docs squidfunk/mkdocs-material:7.1.0
```

blinklet:
Hi @hellt 

Yes, this looks good.

Issue #420, opened 2021-05-12
Add per-node link count


networkop:
In certain situations, it may be desirable to delay container process initialisation until all links are connected by containerlab. For example, this can be done by passing some env var or a configuration file that the containerised process will be able to understand.
 
However, right now the only meaningful arg in `CreateContainer` is `types.Node` which has no awareness of the number of links it's expected to have. Can we introduce an extra field in the `Node` struct to keep count of the total number of links?
This could be done by incrementing this counter in [here](https://github.com/srl-labs/containerlab/blob/14a2cd0e040941debb597b57efff8d5490e3dc1a/clab/config.go#L536), e.g.

```diff
		for name, n := range c.Nodes {
			if name == split[0] {
				endpoint.Node = n
+                              n.LinkCount++
				break
			}
```

hellt:
yes, that makes sense.
we can also leverage the link count in `vr-xxx` nodes where the desired number of network interfaces can open paths for some improvements.

Do you want to contribute that @networkop ?

networkop:
yep, will add this soon

hellt:
added in #422

PR #422, opened 2021-05-15
Pseudo-random MACs & Node.Endpoints

networkop:
This is a PR to address #420 . It's slightly broader in scope and adds the following:

1. Each endpoint of the link now gets a random MAC with `clabOUI` prefix. This is needed to have a pre-determined MAC addresses at the point when container is created. In case of vr-xxx and other VM-based images, this information can be used to build a udev rules file to change the interface naming inside the VM to match the one in the container.
2. `types.Node` now stores all of its local Endpoints (pointers) in a slice. This way its easy to see not only how many extra links Node has, but what are their names and MACs.
3. Delete action now accepts `max-workers` argument, which acts the same way as for the `deploy` action.

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -33,6 +33,8 @@ const (
 	hostNSPath = "__host"
 	// veth link mtu. jacked up to 65k to allow jumbo testing of various sizes
 	defaultVethLinkMTU = 65000
+	// containerlab's reserved OUI
+	clabOUI = "aa:c1:ab"
 )
 
 // supported kinds
@@ -368,6 +370,7 @@ func (c *CLab) NewNode(nodeName string, nodeCfg NodeConfig, idx int) error {
 	node.Fqdn = nodeName + "." + c.Config.Name + ".io"
 	node.LabDir = c.Dir.Lab + "/" + nodeName
 	node.Index = idx
+	node.Endpoints = []*types.Endpoint{}
 
 	node.NetworkMode = strings.ToLower(nodeCfg.NetworkMode)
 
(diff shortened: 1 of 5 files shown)


cmd/destroy.go
```diff
@@ -181,22 +182,44 @@ func destroyLab(ctx context.Context, c *clab.CLab) (err error) {
 		labDir = filepath.Dir(containers[0].Labels["clab-node-lab-dir"])
 	}
 
+	if maxWorkers == 0 {
+		maxWorkers = uint(len(containers))
```
hellt, line 186: what is the idea behind casting to `uint` the length of a slice?

clab/config.go
```diff
@@ -33,6 +33,8 @@ const (
 	hostNSPath = "__host"
 	// veth link mtu. jacked up to 65k to allow jumbo testing of various sizes
 	defaultVethLinkMTU = 65000
+	// containerlab's reserved OUI
+	clabOUI = "aa:c1:ab"
```
hellt, line 37: love how `c1:ab` == `clab` in leetspeak. solid!

cmd/destroy.go
networkop: `maxWorkers` was of `uint` type, I simply re-used the same var. TBH, I don't know why it has that type originally, maybe you would know?

clab/config.go
networkop: yeah, and it's so easy to spot the interfaces managed by clab!

cmd/destroy.go
hellt: ah, yes, it is because cobra flag is of uint type - https://github.com/srl-labs/containerlab/blob/70351f84b7d26cd6e1d3ed5bc89e2db3e46edae0/cmd/deploy.go#L186

Thats ok

hellt:
thanks, LGTM

PR #425, opened 2021-05-20
fixed data race at nodes deletion

hellt:


hellt:
@networkop there was a data race when you delete the lab with >2 nodes. In the meantime I will change the chan to use copies of the GenericContainer structs as a quick fix, but it would be nice to get to the root of it and understand the cause of the data race

networkop:
@hellt do you have steps to reproduce?

hellt:
@networkop , I got help from @karimra to have a fix that keeps the pointer channel
The culprit was with a pointer to the loop variable that got sent into the channel. As I understood Karims explanations, that variable is kept the same throughout the loop cycle, so we had the same pointer passed all over to the workers

https://stackoverflow.com/questions/49123133/sending-pointers-over-a-channel

in   bf13970 I applied the suggested quick fix

networkop:
so i guess this looks like a long-term fix? or did you still want me to find an alternative @hellt ?

hellt:
that is a long term one, yes
no actions are needed

PR #427, opened 2021-05-25
containerd

steiler:
start implementing containerd runtime for containerlab.
Work in progress.


steiler:
Status:
- Containers 
  - ContainerCreate
    - containerd Namespaces
      containerd uses ["namespaces"](https://github.com/containerd/containerd/blob/master/docs/namespaces.md) read "views", to allow for multi tenancy, containerlab utilizes the "clab" namespace.
    - Feature:
      - [x] sysctl
      - [x] labels
      - [x] container user 
      - [x] cmd
      - [x] Environment variables
      - [x] hostname
      - [x] elevated priviledges  
      - [ ] mounts
      - [ ] networking
        - [x] host mode
		- [ ] specify MAC-address for mgmt
        - [ ] port bindings
  - [x] ContainerStart
  - [x] ContainerDelete
  - [x] ContainerStop
  - [x] ContainerList
  - [ ] ContainerInspect
    - Not required right now. So far returned GenericContainer with just PID populated. Refactored also docker runtime such that PID is populated in container list on container create.
  - [ ] Exec
  - [ ] ExecNotWait
  - [ ] PullImageIfRequired
- Network
  - [ ] DeleteNet
  - [ ] CreateNet
  - [ ] SetMgmtNet
  

# Quickstart
The cli tool for containerd is `ctl`. To get started you need to import the SRLinux image into the _clab_ namespace into containerd.
## Import SRLinux image
```
# create namespace
#   might be omitted if you've started containerlab already
ctr namespace create clab
# decompress the SRLinux image
xzcat 21.3.2-129.tar.xz > 21.3.2-129.tar
# import it into containerd clab namespace
ctr -n clab images import /tmp/21.3.2-129.tar
# retag the imported image, since it is imported as "docker.io/library/srlinux:21.3.2-129"
ctr -n clab images tag docker.io/library/srlinux:21.3.2-129 srlinux:21.3.2-129 srlinux:latest
# remove the "docker.io/library"-based tag
ctr -n clab images rm docker.io/library/srlinux:21.3.2-129
```

Now you're good to use clab with the `-r containerd` knob.

# Usefull commands
## List Containers
```
ctr -n clab containers ls
```
## show OCI spec of container
```
ctr -n clab containers info <CONTAINERNAME>
```
## show tasks (running container init processes)
```
ctr -n clab tasks ls
```

steiler:
@hellt 
We should check for entrypoint and cmd in the multitool image.

If I'm using the nginx image all works fine:
docker.io/library/nginx:latest

If using the ghcr.io/hellt/network-multitool:latest image, the oci.WithImageConfig(img),  yields `/docker-entrypoint bash` as the cmd to execute in the container .... 
adjusting the cmd in the topology file works fine. 
```
    linux:
      image: ghcr.io/hellt/network-multitool:latest
      cmd: /docker-entrypoint.sh nginx -g 'daemon off;'`
```
Not sure yet if this is about the image or what it is, since the official nginx image works fine.

clab/srl.go
```diff
@@ -107,6 +107,7 @@ func initSRLNode(c *CLab, nodeCfg NodeConfig, node *types.Node, user string, env
 	node.Sysctls["net.ipv6.conf.default.accept_dad"] = "0"
 	node.Sysctls["net.ipv6.conf.all.autoconf"] = "0"
 	node.Sysctls["net.ipv6.conf.default.autoconf"] = "0"
+	node.Sysctls["net.ipv4.ip_unprivileged_port_start"] = "0"
```
hellt, line 110: where is this new sysctl coming from?

cmd/inspect.go
```diff
@@ -128,9 +133,8 @@ func printContainerInspect(c *clab.CLab, containers []types.GenericContainer, br
 			IPv4Address: getContainerIPv4(cont, bridgeName),
 			IPv6Address: getContainerIPv6(cont, bridgeName),
 		}
-		if len(cont.ID) > 11 {
```
hellt, line 131: this we need to keep to trim container id to 12 chars

steiler: as discussed offline, we need to take care of the runtime ... containerd doesn't use UID kind of identifiers but the container names...

cmd/destroy.go
```diff
@@ -205,13 +207,12 @@ func destroyLab(ctx context.Context, c *clab.CLab) (err error) {
 						log.Debugf("Worker %d terminating...", i)
 						return
 					}
-					name := cont.ID
-					if len(cont.Names) > 0 {
-						name = strings.TrimLeft(cont.Names[0], "/")
-					}
-					err := c.Runtime.DeleteContainer(ctx, name)
+					//if len(cont.Names) > 0 {
```
hellt, line 210: shall we bring this back? to display which container we might filed to delete?

clab/srl.go
steiler: again me bruteforcing the permission issue ... seen that somewhere ... we can probably delete it.

cmd/destroy.go
steiler: yes we can do that. I'm not sure anymore why it was deleted

runtime/docker/docker.go
```diff
@@ -380,6 +356,19 @@ func (c *DockerRuntime) ListContainers(ctx context.Context, labels []string) ([]
 	return c.produceGenericContainerList(ctrs, nr)
 }
 
+func (c *DockerRuntime) containerFilterBuilder(gfilters []*types.GenericFilter) filters.Args {
```
hellt, line 359: let use the same name for these functions across docker.go and containerd.go?

like `buildFilterString`?

steiler: have adjusted that

clab/srl.go
steiler: have removed that sysctl

cmd/destroy.go
hellt: brought back in 0bde6cf

cmd/inspect.go
hellt: added in bee3dab

hellt:
@steiler let's add one check for CNI binaries to be present at `/opt/cni/bin`
after that it is ok to let @karimra maybe have a look at it =)

steiler:
@karimra happy for you to take over and review now.

runtime/containerd/containerd.go
```diff
@@ -37,6 +39,8 @@ const (
 	defaultTimeout    = 30 * time.Second
 )
 
+var cniPath string
```
hellt, line 42: let's not make this var global, it is not needed it seems

put it under `Init()` function where it is needed to check the cni bins

I'd do like this:

```go
cniBinPath := cniBin // init the path to be the default one
if cniBinEnvPath, ok = os.LookupEnv("CNI_BIN"); ok { // change the condition to be ok
		cniBinPath = cniBinEnvPath
	}
```

hellt: and you don't need to declare `ok`, before that, just use `:=` in the `if` statement

steiler: The path is used on cni execution as well... hence I've pulled that into init to not duplicate code that could then diverse ... 
then using the `:=` operator did not work ...
```
func cniInit(cId, ifName string, mgmtNet *types.MgmtNet) (*libcni.CNIConfig, *libcni.NetworkConfigList, *libcni.RuntimeConf, error) {
	// allow overwriting cni plugin binary path via ENV var

	cnic := libcni.NewCNIConfigWithCacheDir([]string{cniPath}, cniCache, nil)

	cniConfig := fmt.Sprintf(`
	{
		"cniVersion": "0.4.0",
		"name": "clabmgmt",
		"plugins": [
```

runtime/containerd/containerd.go
```diff
@@ -0,0 +1,756 @@
...
+
+func (c *ContainerdRuntime) StartContainer(ctx context.Context, containername string) error {
+	container, err := c.client.LoadContainer(ctx, containername)
+	if err != nil {
+		return err
+	}
+	task, err := container.NewTask(ctx, cio.LogFile("/tmp/clab/"+containername+".log"))
+	if err != nil {
+		log.Fatal(err)
+		log.Fatalf("Failed to start container %s", containername)
```
karimra, line 405: what is the intent here? terminate all ? 2x `log.Fatal` doesn't make sense

runtime/containerd/containerd.go
```diff
@@ -0,0 +1,756 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+func (c *ContainerdRuntime) StopContainer(ctx context.Context, containername string, dur *time.Duration) error {
+	ctask, err := c.getContainerTask(ctx, containername)
+	if err != nil {
+		log.Debugf("no task found for container %s", containername)
+		return nil
```
karimra, line 419: return err ? and add it to the debug log statement ?

runtime/containerd/containerd.go
```diff
@@ -0,0 +1,756 @@
...
+}
+
+func waitContainerStop(ctx context.Context, exitCh <-chan containerd.ExitStatus, id string) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case status := <-exitCh:
+		return status.Error()
+	}
+}
```
karimra, line 479: is `id string` needed ?

runtime/containerd/containerd.go
```diff
@@ -0,0 +1,756 @@
...
+	ctx = namespaces.WithNamespace(ctx, containerdNamespace)
+	cont, err := c.client.LoadContainer(ctx, containername)
+	if err != nil {
+		return nil, err
+	}
+	task, err := cont.Task(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+	return task, nil
```
karimra, line 491: I think just `return cont.Task()`

steiler:
pulled `GetCNIBinaryPath()` into utils package, such that it can be used by other runtimes as well.

runtime/containerd/containerd.go
```diff
@@ -0,0 +1,756 @@
...
+func (c *ContainerdRuntime) CreateNet(ctx context.Context) error {
+	log.Debug("CreateNet() - Not needed with containerd")
+	return nil
+}
+func (c *ContainerdRuntime) DeleteNet(context.Context) error {
+	log.Debug("DeleteNet() - Not yet required with containerd")
+	return nil
+}
+func (c *ContainerdRuntime) ContainerInspect(context.Context, string) (*types.GenericContainer, error) {
+	log.Fatalf("ContainerInspect() - Not implemented yet")
```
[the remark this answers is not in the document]
steiler: We do not need ContainerInspect anyways ...
The clab inspect Command utilizes ListContainers().  In Docker runtime it was just used to retrieve the containers pid.
My proposal is to get rid of the function.

runtime/containerd/containerd.go
steiler: well if the container is to be deleted but the task already died ... we log it and exit without error ... because the container is already stopped ...
So to me this looks fine.

clab/clab.go
```diff
@@ -341,7 +344,7 @@ func (c *CLab) DeleteNodes(ctx context.Context, workers uint, containers []types
 					if len(cont.Names) > 0 {
 						name = strings.TrimLeft(cont.Names[0], "/")
 					}
-					err := c.Runtime.DeleteContainer(ctx, name)
+					err := c.Runtime.DeleteContainer(ctx, cont)
 					if err != nil {
 						log.Errorf("could not remove container '%s': %v", name, err)
 					}
```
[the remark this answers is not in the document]
steiler: dropped it ... and used cont.ID instead under the `log.Errorf`

clab/config.go
```diff
@@ -661,7 +661,7 @@ func (c *CLab) VerifyContainersUniqueness(ctx context.Context) error {
 	nctx, cancel := context.WithTimeout(ctx, c.timeout)
 	defer cancel()
 
-	var labels []string
+	labels := []*types.GenericFilter{}
```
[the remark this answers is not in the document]
steiler: I don't see that. labels is iterated over, so it must at least be a slice / array ... although having a length of 0.

runtime/containerd/containerd.go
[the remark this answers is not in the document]
karimra: If it's used why get rid of it? 
But the point is the log.Fatal, that causes the process to terminate, which is not the intended behavior

cmd/exec.go
```diff
@@ -44,7 +45,7 @@ var execCmd = &cobra.Command{
 		}
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		labels = append(labels, "containerlab="+name)
+		labels := []*types.GenericFilter{{FilterType: "label", Match: name, Field: "containerlab", Operator: "="}}
```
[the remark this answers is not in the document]
steiler: https://github.com/srl-labs/containerlab/commit/9158e9eb70af982f5bace87c21c0d1785a4aaf15#diff-ce92eefe0908ac93cdc54bd33b7d01c347c159ae09a9324bd0479acdbc38c442R48-R50

introduced function to create genericfilter from commandline switch.

runtime/containerd/containerd.go
karimra: is the only possible error coming from the task dying ? or task not found?
the error itself it not logged

clab/config.go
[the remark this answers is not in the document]
karimra: you can still iterate over a slice with value `nil`, the type being comes from the function signature not from the variable you pass to it.

runtime/containerd/containerd.go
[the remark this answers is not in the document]
steiler: Used the IDE to search for references ... it is not used, not even as a "private" method ... 
but for now I'll just change to Debugf ...

clab/config.go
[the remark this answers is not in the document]
steiler: Thanks, my bad. Tried the following, which failed:
```
func main() {
	x = nil
	for y,_ := range x {
		fmt.Print(y)
	}
}
```
doing it properly, declaring x as array of some type it works fine... 
That was me being stuck in python I guess.
```
var x []interface{}
x = nil
```

runtime/containerd/containerd.go
steiler: using the following now:
```
	ctask, err := c.getContainerTask(ctx, containername)
	if err != nil {
		log.Debugf("container %s: %v", containername, err)
		return nil
	}
```

runtime/containerd/containerd.go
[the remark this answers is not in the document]
karimra: seems it's not used in your branch anymore.
let's keep it for now, and remove in another PR if it's not needed anymore

clab-clos01/leaf1/config/devices/slot_nums.json
```diff
@@ -0,0 +1,17 @@
+{
```
hellt, line 1: where is this file coming from?

clab/clab.go
```diff
@@ -58,15 +59,24 @@ func WithTimeout(dur time.Duration) ClabOption {
 
 func WithRuntime(name string, d bool, dur time.Duration, gracefulShutdown bool) ClabOption {
 	return func(c *CLab) {
+		if name == "" {
```
hellt, line 62: lets make a `switch` here

first order is the cli flag
then env var
default is docker runtime

clab-clos01/leaf1/config/devices/slot_nums.json
hellt: removed it

runtime/containerd/containerd.go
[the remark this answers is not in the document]
hellt: made it to return an error 8261c10c033a8c133181640e01393fb52bc3765b

clab/clab.go
hellt: added in 91bcd27

hellt:
thanks @steiler for the initial work and @karimra for a review!

let's merge it and test the regular docker runtime from the master build to check if those touches to a non containerd code didn't break in some edge cases

Issue #429, opened 2021-05-27
mgmt-net regression in 0.14


hellt:
when deploying a topology like that

```yaml
name: mgmt
topology:
  nodes:
    n1:
      kind: srl
      image: srl/auto-config:latest
      license: license.key
  links:
    - endpoints:
        - "n1:e1-1"
        - "mgmt-net:n1-e1-1"
```

the deployment fails with 
```
[root@kvm61 clab]# clab deploy --topo mgmt.yml
INFO[0000] Parsing & checking topology file: mgmt.yml  
INFO[0000] Creating lab directory: /root/clab/clab-mgmt
INFO[0000] Creating root CA                            
INFO[0000] Creating container: n1                      
INFO[0001] Creating virtual wire: n1:e1-1 <--> mgmt-net:n1-e1-1
ERRO[0001] failed to Statfs "": no such file or directory
INFO[0001] Writing /etc/hosts file  
```

potentially a problem with MAC addresses introduced in #422

PR #431, opened 2021-05-27
fix mgmt-net feature regression

karimra:
#429 
Use a pointer to `types.MgmtNet` so that the bridge name is set on the same MgmtNet struct as the one used to create the links.
This caused the `mgmt-net` feature to fail due to an empty bridge name set on the virtual link.


clab/clab.go
```diff
@@ -84,7 +84,9 @@ func WithGracefulShutdown(gracefulShutdown bool) ClabOption {
 // NewContainerLab function defines a new container lab
 func NewContainerLab(opts ...ClabOption) *CLab {
 	c := &CLab{
-		Config:   new(Config),
+		Config: &Config{
+			Mgmt: new(types.MgmtNet),
```
hellt, line 88: I wonder if we need to initialize the `Mgmt` with `new()`? Since now `Config.Mgmt` is of `*types.MgmtNet`, then won't `nil` be a default value if we do

```go
c := &CLab{
		Config: new(Config),
```

karimra: The goal from new() is not to have it be `nil`, so that even if it's not initialized later on, its values can be accessed without panicking

hellt: true, I missed the fact that  `new(types.MgmtNet)` is not `new(*types.MgmtNet)`

hellt:
thanks!

Issue #432, opened 2021-05-28
inspect command doesn't include ipv4/ipv6 IP Addressing


thinghacker:
With containerlab version 0.14.2 when I deploy a lab I can see the IPv4 and IPv6 address information but it is no longer shown when I use the inspect command.

Example is below.

```
adam@m4600-cl:~/simpledc$ sudo containerlab deploy --topo simpledc.yml
INFO[0000] Parsing & checking topology file: simpledc.yml
INFO[0000] Creating lab directory: /home/adam/simpledc/clab-simpledc
INFO[0000] Creating docker network: Name='clab', IPv4Subnet='172.20.20.0/24', IPv6Subnet='2001:172:20:20::/64', MTU='1500'
INFO[0000] Creating container: gateway
INFO[0000] Creating container: leaf1
INFO[0000] Creating container: leaf2
INFO[0000] Creating container: spine1
INFO[0000] Creating container: spine2
INFO[0001] Creating virtual wire: spine1:e1-3 <--> gateway:e1-1
INFO[0001] Creating virtual wire: leaf2:e1-2 <--> spine2:e1-2
INFO[0001] Creating virtual wire: leaf1:e1-1 <--> spine1:e1-1
INFO[0001] Creating virtual wire: spine2:e1-3 <--> gateway:e1-2
INFO[0001] Creating virtual wire: leaf2:e1-1 <--> spine1:e1-2
INFO[0001] Creating virtual wire: leaf1:e1-2 <--> spine2:e1-1
INFO[0002] Writing /etc/hosts file
+---+-----------------------+--------------+--------------------------+---------+-------+---------+----------------+----------------------+
| # |         Name          | Container ID |          Image           |  Kind   | Group |  State  |  IPv4 Address  |     IPv6 Address     |
+---+-----------------------+--------------+--------------------------+---------+-------+---------+----------------+----------------------+
| 1 | clab-simpledc-gateway | f180cd08991d | vrnetlab/vr-sros:21.2.R2 | vr-sros |       | running | 172.20.20.6/24 | 2001:172:20:20::6/64 |
| 2 | clab-simpledc-leaf1   | 7f8008a8c6a3 | srlinux:21.3.1-410       | srl     |       | running | 172.20.20.2/24 | 2001:172:20:20::2/64 |
[...]

PR #433, opened 2021-05-28
populate bridgeName if inspect cmd is run with --all or --name

karimra:
#432

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -28,6 +28,7 @@ jobs:
           filters: |
             code:
               - 'clab/**'
+              - 'runtime/**'
               - 'cmd/**'
               - 'tests/**'
               - '.github/workflows/cicd.yml'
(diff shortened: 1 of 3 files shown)


steiler:
Karim did you consider the network mode "host"?
Quickly scanning the code I unfortunately can't tell.
https://github.com/srl-labs/containerlab/blob/62e901c512cc4f8bd54d509065c3fb7ef2b1e8dd/runtime/docker/docker.go#L222-L236

karimra:
what would be printed in case of host mode ?

hellt:
I think for now it is ok if it prints nothing (output from the tip of this branch)

image

the potential improvements are to print `host` maybe?

karimra:
yes, there is basically nothing to print, network name is `host` and was not created by clab --> skip

Instead of writing `host` under an IP address column, maybe add a column `Network Mode` or `Bridge Name`?

steiler:
Printing host sounds good to me.
Noting is also fine for now, just thought that this case might cause errors, not sure if this would be covered by a test so far.

steiler:
@karimra to basically normalize it ... thats also a great idea!

hellt:
I think a column named `Network` with the potential values makes sense, wdyt?

- docker network name if it is a bridged network
- `host-mode` string if it is a host mode networ

PR #434, opened 2021-05-28
fixed ip addr retrieval for inspect --all case

hellt:
Hi @karimra 
I wonder if you consider this to be a less intrusive approach as opposed to #433 ?
The assumption here is that clab containers can have only one network they are connected to

### DIFF
--- a/runtime/docker/docker.go
+++ b/runtime/docker/docker.go
@@ -379,11 +379,23 @@ func (c *DockerRuntime) produceGenericContainerList(input []dockerTypes.Containe
 				Set: false,
 			},
 		}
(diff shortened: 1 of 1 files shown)


karimra:
That was the way it was done before the runtime introduction.

I did it that way first then decided to query the bridges that were created by clab, users can always do a `docker network connect someDockerNetwork clab-nodeX` making the output random.

But under the assumption that only one docker network is connected, both ways work. The extra step is a docker network listing.

hellt:
agree, that makes sense to let ppl to mess with container networking afterwards

PR #436, opened 2021-05-31
remove runtime flag shorthand

karimra:
`-r` is already used for flag`--remote` in cmd `vxlan create`


hellt:
I have a feeling that runtime shorthand has a potential to be used more frequently than `vxlan remote` when we add containerd runtime

Since these flags do not share the scope of the same command/level, maybe we can keep both?

karimra:
It's not possible to keep both, `runtime` is a global flag, so it applies to all sub commands, including `vxlan create`
I removed the last one added, I will switch

hellt:
thank you!
I somehow thought cobra does checks of flag names and shorthands, but apparently it only happens during runtime of the conflicting command

PR #444, opened 2021-06-12
Kind interface

hellt:
The goal of this PR is to make the `kind` interface that all other specific kinds should implement

I thinks I am hitting my Go skills ceiling, as I couldn't get this to work, hitting endless import dependencies.

Maybe @karimra or @kellerza or anybody can unblock this work and/or take this from scratch


karimra:
I don't see the `Node` interface definition under `containerlab/kind`, not defined yet ?

hellt:
Maybe that is a terminology issue? What should Node interface be?
I may thought that kind interface will be holding all the fields of Node
struct

karimra:
I saw this:
```golang
type CLab struct {
	Config   *Config
	TopoFile *TopoFile
	m        *sync.RWMutex
	Nodes    map[string]kind.Node // <-- this
      ///
}
```
So I assume you want to call the interface `Node` ?
But yes, the name doesn't matter, I also don't see a `Kind` interface.
Just trying to see how far this PR went.

hellt:
@karimra yes, it seems I named the common interface that represents a lab node - a `Node` interface https://github.com/srl-labs/containerlab/pull/444/files#diff-f8d2bd35e1a8c9b86c8c4448dde00bc5480652f85ed0630efc012c3906d3a14fR128

and then different kinds implement this interface that can be passed over to the runtime for creation

karimra:
The folders structure need to be changed to avoid the import cycle.
The interface definition will also need to be updated to expose the node attributes.
I can take it if no one else will

hellt:
If you can take this, that would be greate and very helpful as the toll
keeps growing with all other prs building on top of the existing non
modular architecture

hellt:
in favor of #449

Issue #445, opened 2021-06-12
Consider adding Makefile


networkop:
Would be nice to have a makefile with at least a minimal set of targets, e.g. build, clean, test
WDYT?

hellt:
I don't mind, till it has a simple logic under targets

PR #446, opened 2021-06-13
Added basic makefile

networkop:
Related issue #445


hellt:
thanks @networkop 
I've added one more target for container-based linting step

If you're ok with how it looks at the end, I am good to merge this

networkop:
yep. lgtm

PR #448, opened 2021-06-15
rework runtime interface

karimra:
This PR decouples runtime import from its implementations.
This should help avoiding import cycles when implementing node `kind` interface.


hellt:
thanks @karimra 
as far as I understand, the important piece is that here we do not anymore import the docker package?
https://github.com/srl-labs/containerlab/pull/448/files#diff-bb8ad382f59edfae9297e446895a161036f39115189e67da9d9f377a684a8f82L12

karimra:
yes, that's the idea.

PR #449, opened 2021-06-15
adding Node interface{}

karimra:
Attempt at #444 , still a WIP

Edit: Tempted to keep it at this line add count :D
image

PR #455, opened 2021-06-17
removed InspectContainer(...) 

steiler:
remove InspectContainer(...)  from runtime interface since it is not used


hellt:
maybe let's park this, till @karimra's work on #449 and related PRs take shape?
Just to make sure we won't need anything apart from PID

hellt:
done, merged it into #469 as this fits there nicely

Issue #456, opened 2021-06-17
`inspect --all` doesn't show IP information for nodes launched in default docker bridge


hellt:
consider a simple lab with one node connected to the default docker `bridge` network

```yaml
name: issue
mgmt:
  network: bridge

topology:
  nodes:
    client1:
      kind: linux
      image: ghcr.io/hellt/network-multitool
```

when this lab is deployed, it has IP information reported:

```
❯ clab dep -t ins-all-bridge.clab.yml 
INFO[0000] Parsing & checking topology file: ins-all-bridge.clab.yml 
INFO[0000] Creating lab directory: /root/testclab/clab-issue 
INFO[0000] Creating container: client1                  
INFO[0000] Writing /etc/hosts file                      
+---+--------------------+--------------+---------------------------------+-------+-------+---------+---------------+--------------+
| # |        Name        | Container ID |              Image              | Kind  | Group |  State  | IPv4 Address  | IPv6 Address |
+---+--------------------+--------------+---------------------------------+-------+-------+---------+---------------+--------------+
| 1 | clab-issue-client1 | e1aadcbff1af | ghcr.io/hellt/network-multitool | linux |       | running | 172.17.0.2/16 | NA           |
+---+--------------------+--------------+---------------------------------+-------+-------+---------+---------------+--------------+

```

But a subsequnet `ins -a` doesn't get it

```
❯ clab ins -a
+---+-------------------------+----------+--------------------+--------------+---------------------------------+-------+-------+---------+--------------+--------------+
| # |        Topo Path        | Lab Name |        Name        | Container ID |              Image              | Kind  | Group |  State  | IPv4 Address | IPv6 Address |
+---+-------------------------+----------+--------------------+--------------+---------------------------------+-------+-------+---------+--------------+--------------+
| 1 | ins-all-bridge.clab.yml | issue    | clab-issue-client1 | e1aadcbff1af | ghcr.io/hellt/network-multitool | linux |       | running |              |              |
+---+-------------------------+----------+--------------------+--------------+---------------------------------+-------+-------+---------+--------------+--------------+
```

PR #458, opened 2021-06-17
fixing #456

steiler:
The filter caused the default bridge to be excluded from the network listing. 
Hence the container IP wasn't found.
I'm not sure why the filter was introduced in #433 for issue #432. Topology mentioned in #456 works fine. #433 has no topo attached to test.

### DIFF
--- a/runtime/docker/docker.go
+++ b/runtime/docker/docker.go
@@ -52,7 +52,7 @@ type DockerRuntime struct {
 
 func (c *DockerRuntime) Init(opts ...runtime.RuntimeOption) error {
 	var err error
-	log.Info("Runtime: Docker")
+	log.Debug("Runtime: Docker")
 	c.Client, err = dockerC.NewClientWithOpts(dockerC.FromEnv, dockerC.WithAPIVersionNegotiation())
 	if err != nil {
 		return err
--- a/runtime/docker/docker.go
+++ b/runtime/docker/docker.go
@@ -360,14 +360,10 @@ func (c *DockerRuntime) ListContainers(ctx context.Context, gfilters []*types.Ge
 	}
 	var nr []dockerTypes.NetworkResource
 	if c.Mgmt.Network == "" {
-		netFilter := filters.NewArgs()
-		netFilter.Add("label", "containerlab")
 		nctx, cancel := context.WithTimeout(ctx, c.timeout)
 		defer cancel()
 
-		nr, err = c.Client.NetworkList(nctx, dockerTypes.NetworkListOptions{
-			Filters: netFilter,
-		})
+		nr, err = c.Client.NetworkList(nctx, dockerTypes.NetworkListOptions{})
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


hellt:
I believe the idea was to get the list of networks that containerlab created. hence the filter that has `containerlab`
but since `bridge` network is not the one that is created by containerlab, it was missing

runtime/docker/docker.go
```diff
@@ -360,14 +360,10 @@ func (c *DockerRuntime) ListContainers(ctx context.Context, gfilters []*types.Ge
 	}
 	var nr []dockerTypes.NetworkResource
 	if c.Mgmt.Network == "" {
-		netFilter := filters.NewArgs()
-		netFilter.Add("label", "containerlab")
```
hellt, line 364: I wonder if we can create a filter that will capture both networks that have a `containerlab` label OR network which have `bridge` name

That way we will avoid getting the networks that could be created by users/other_systems and which are not needed for our purposes.

It is potentially an optimization, which is not necessary, but seems like nice to have.

karimra: It will be an AND operation if you add `name=bridge`
To do an OR, 2 queries need to be performed.

hellt: yes, here is what I did

```go
	if c.Mgmt.Network == "" {
		defer cancel()
		// fetch containerlab created networks
		f := filters.NewArgs()
		f.Add("label", "containerlab")
		nr, err = c.Client.NetworkList(nctx, dockerTypes.NetworkListOptions{
			Filters: f,
		})

		if err != nil {
			return nil, err
		}

		// fetch default bridge network
		f = filters.NewArgs()
		f.Add("name", "bridge")
		bridgenet, err := c.Client.NetworkList(nctx, dockerTypes.NetworkListOptions{
			Filters: f,
		})

		if err != nil {
			return nil, err
		}

		nr = append(nr, bridgenet...)
```

not sure which is better though =)

karimra:
This wouldn't work if no mgmt network was specified in the topo file, i.e use network `clab`

I'd say either do 2 queries for the networks (label=containerlab and name=bridge),

or make the var `bridgeName` default to `bridge` if it's still empty after going through the networks list, here: https://github.com/srl-labs/containerlab/blob/master/runtime/docker/docker.go#L421

hellt:
> This wouldn't work if no mgmt network was specified in the topo file, i.e use network `clab`

what wouldn't work? current version of this branch c8142b6 where filters are empty seems to work fine with the default network (i.e. omitted `mgmt` container in clab file)

karimra:
it works because the containers are connected to a single bridge, so it defaults to that one.

The idea in #433 was to make sure the IPs printed come from a bridge that was created by clab, which turns to be too restrictive in case `--all` is used.
The condition should be, a bridge created by clab or a bridge called `bridge`

hellt:
in 38b11a3 the two-filters approach has been added

PR #460, opened 2021-06-17
Deduplicating destroy -t vs destroy --all

steiler:
The destroy action did carry a lot of duplicated code for the --all and the -t option.
This pull request is to streamline that process.

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -35,69 +35,62 @@ var destroyCmd = &cobra.Command{
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
+		opts := []clab.ClabOption{
+			clab.WithDebug(debug),
+			clab.WithTimeout(timeout),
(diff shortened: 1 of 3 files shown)


cmd/destroy.go
```diff
@@ -35,69 +35,62 @@ var destroyCmd = &cobra.Command{
...
 		switch {
 		case !all:
-			// stop if not topo file provided and not all labs are requested
-			// to be deleted
-			if err = topoSet(); err != nil {
-				return err
+			{
+				// stop if not topo file provided and not all labs are requested
+				// to be deleted
+				if err = topoSet(); err != nil {
```
hellt, line 51: I think that is not needed anymore, as when you do `c := clab.NewContainerLab(opts...)` later, the NetContainerlLab will return an error since the topo file is not found

steiler: removed the func now along with the remaining reference from cmd/deploy.go

PR #461, opened 2021-06-17
move topology struct and config items to types package

karimra:
This PR moves the topology struct to types package and adds some  helper functions to make the nodes initialization more uniform.
Still a WIP.


types/topology.go
```diff
@@ -0,0 +1,302 @@
...
+		// kind level ports
+		if kdef, ok := t.Kinds[ndef.Kind]; ok && kdef != nil {
+			if len(kdef.Ports) > 0 {
+				ps, pb, err := nat.ParsePortSpecs(kdef.Ports)
+				if err != nil {
+					return nil, nil, err
+				}
+				return ps, pb, nil
+			}
+		}
```
hellt, line 103: can we use kind-level ports? since a container port is bound to a specific host port, it seems it won't be possible to define a kind-scoped port forwarding as this would fail to bind the same host port >1 times

clab/ceos.go
```diff
@@ -52,18 +52,16 @@ func ceosPostDeploy(ctx context.Context, c *CLab, node *types.Node, lworkers uin
 	return err
 }
 
-func initCeosNode(c *CLab, nodeCfg NodeConfig, node *types.Node, user string, envs map[string]string) error {
+func (c *CLab) initCeosNode(nodeCfg *types.NodeConfig) error {
 	var err error
 
-	// initialize the global parameters with defaults, can be overwritten later
-	node.Config, err = c.configInit(&nodeCfg, node.Kind)
+	c.Config.Topology.GetNodeConfig(nodeCfg.ShortName)
```
hellt, line 58: looks like assignments are missing here

karimra: yes, will fix

hellt:
the rest LGTM,
thank you for being able to get this properly structured 🙏

Issue #271, opened 2021-02-16
add a --keep-mgmt-net flag to not attempt to delete a mgmt network 


hellt:
In certain situations it is not needed to delete the management network, even if it has no endpoints attached.
Proposed solution is to add a flag that will prevent mgmt net deletion

hellt:
done in #462

PR #462, opened 2021-06-18
Implement --keep-mgmt-net for destroy subcmd

steiler:
Implementing #271.

This is based upon changes introduced in #460.

At the same time it adds an implementation of the DeleteNet() for containerd.
Here we are listing all the Host interfaces to see if the bridge is still the master to any existing device, if not, and not --keep-mgmt-net is set, the deletion of the cni based mgmt bridge is executed.


cmd/destroy.go
```diff
@@ -209,7 +210,7 @@ func destroyLab(ctx context.Context, c *clab.CLab) (err error) {
 	}
 
 	// delete lab management network
-	log.Infof("Deleting network '%s'...", c.Config.Mgmt.Network)
```
hellt, line 212: why do you remove the explicit message with a network name?

runtime/docker/docker.go
```diff
@@ -183,28 +189,29 @@ func (c *DockerRuntime) CreateNet(ctx context.Context) (err error) {
 
 // DeleteNet deletes a docker bridge
 func (c *DockerRuntime) DeleteNet(ctx context.Context) (err error) {
-	if c.Mgmt.Network == "bridge" {
-		log.Debug("Skipping potential deletion of docker default bridge 'bridge'.")
+	network := c.Mgmt.Network
+	if network == "bridge" || c.keepMgmtNet {
+		log.Infof("Skipping potential deletion of docker default bridge '%s'", network)
```
hellt, line 194: lets make it like that maybe:

```go
network := c.Mgmt.Network
	if network == "bridge" || c.keepMgmtNet {
		log.Debugf("Skipping deletion of '%s' network", network)
```

clab/clab.go
```diff
@@ -88,6 +88,12 @@ func WithRuntime(name string, d bool, dur time.Duration, gracefulShutdown bool)
 	}
 }
 
+func WithKeepMgmtNet() ClabOption {
+	return func(c *CLab) {
+		c.Runtime.WithKeepMgmtNet()
```
hellt, line 93: do we need it to be a function? maybe just set a field on RuntimeConfig struct to `true`?

runtime/containerd/containerd.go
```diff
@@ -39,14 +39,14 @@ const (
...
-			Mgmt: new(types.MgmtNet),
-		}
+			Mgmt: new(types.MgmtNet)}
 	})
 }
 
 func (c *ContainerdRuntime) Init(opts ...runtime.RuntimeOption) error {
 	var err error
 	log.Debug("Runtime: containerd")
+	c.keepMgmtNet = false
```
hellt, line 49: no need to set it, this is the default value

cmd/destroy.go
steiler: because that message is  runtime dependent ... in containerd the bridge name is stored in c.Config.Mgmt.Bridge...
hence this message was replaced with the next line `log.Infof("Cleanup networking")` and the final message what that cleanup means comes from the runtime.

runtime/docker/docker.go
steiler: Changing the message makes absolute sense, but this debug level should be the same as the above.
Otherwiese that would be cleaning up networking, but that would be strange if it would then still exist ... 
so lets either do debug or info for both...

clab/clab.go
steiler: @karimra this seemed to me to be the most in line approach... what do you think?

runtime/containerd/containerd.go
steiler: ok removed it

cmd/destroy.go
hellt: the implementation is runtime specific, but as far as clab is concerned, there is always a management network with a certain name.

So what I'd propose is the following

1. do the check if the deletion of the network is needed (not bridge and not keepMgmtNet) here, and do the log.Info if the deletion is needed
2. Output the original `log.Infof("Deleting network '%s'...", c.Config.Mgmt.Network)` then, since now it will mean that the we will make an attempt to delete the network
3. the DeleteNet runtime function will not emit any Info messages

cmd/destroy.go
```diff
@@ -19,8 +19,11 @@ import (
 	"github.com/srl-labs/containerlab/types"
 )
 
-var cleanup bool
-var graceful bool
+var (
+	cleanup     bool
+	graceful    bool
+	keepmgmtnet bool
```
hellt, line 25: lets make var name`keepMgmtNet`

cmd/destroy.go
```diff
@@ -102,6 +109,7 @@ func init() {
 	destroyCmd.Flags().BoolVarP(&graceful, "graceful", "", false, "attempt to stop containers before removing")
 	destroyCmd.Flags().BoolVarP(&all, "all", "a", false, "destroy all containerlab labs")
 	destroyCmd.Flags().UintVarP(&maxWorkers, "max-workers", "", 0, "limit the maximum number of workers deleteing nodes")
+	destroyCmd.Flags().BoolVarP(&keepmgmtnet, "keep-mgmt-net", "", false, "do not remove the mgmt network")
```
hellt, line 112: lets use `do not remove the management network`

clab/clab.go
karimra: Both approaches are fine, I added the `WithConfig` methods to avoid having a method per option.
At the end, it's about whether you want to allow changing this value after it's set or not.
At this moment, when clab is used only by the cli, both approaches are the same.

hellt: To clarify, I am talking about the extra runtime interface method called `c.Runtime.WithKeepMgmtNet()`
to me it looks like doing this provides less boilerplate:

```go
func WithKeepMgmtNet() ClabOption {
	return func(c *CLab) {
             c.Runtime.keepMgmtNet = true
        }
}
```
and we keep interface smaller at the same time


Another approach is to use the existing `WithConfig` interface method and set the runtimeconfig's `keepMgmtNet` field there

karimra: You can't do it that way, `Runtime` is an interface, it doesn't have fields, only methods.

What I'm talking about is adding  ` KeepMgmtNet bool` to:
```golang
type RuntimeConfig struct {
	Timeout          time.Duration
	GracefulShutdown bool
	Debug            bool
}
```

steiler: @karimra which would result in adding that parameter to withConfig, which would require to specify this parameter even in deploy or inspect scenarios something that I found disturbing ...

cmd/destroy.go
steiler: Take a look at the actual approach ...
The issue again is, that we need to delegate the decision to the runtime if something can be deleted or not.
Names are for Docker defined in the c.Config.Mgmt.Network and for containerd in c.Config.Mgmt.Bridge because we do not need and additional Network abstraction here ... 
Also take a look at the comments on the struct
https://github.com/srl-labs/containerlab/blob/eb969dba156c9f7063cfe9f7fa797104ec7dc110/types/types.go#L38-L44

clab/clab.go
karimra: No need to specify it, the default value is enough in those cases.
anyways, this is just a matter of style/preference.

steiler:
Alright then, I'd consider this done now.

steiler:
Resolved merge conflicts. 
We're ready to merge here, aren't we?

utils/netlink.go
```diff
@@ -50,6 +50,34 @@ func DefaultNetMTU() (string, error) {
...
+	for _, link := range l {
+		if link.Attrs().MasterIndex == mgmtbridx {
+			InUse = true
+			break
+		}
+	}
+	return InUse, nil
+}
+
+func DeleteNetworkInterface(name string) error {
```
hellt, line 73: maybe this should be better called `DeleteLinkByName`?

hellt:
> We're ready to merge here, aren't we?

I think it is time to add a test for this case

utils/netlink.go
steiler: done

steiler:
@hellt are you fine with the tests

hellt:
@steiler can you please also add a doc flag to `destroy` command?

hellt:
@steiler also I think you wanted to add the schema entry for bridge under `mgmt`?

steiler:
@hellt That is not related to this flag, so it should go into a new PR.

steiler:
Since we've gone downa different path now, with karims work on refactoring the runtime interface, should the withKeepMgmtNet() func now be replaced by jsut the field in the withConfig piece?
https://github.com/srl-labs/containerlab/pull/462/files#diff-bb8ad382f59edfae9297e446895a161036f39115189e67da9d9f377a684a8f82R28

runtime/ignite/iginite.go
```diff
@@ -119,6 +119,9 @@ func (c *IgniteRuntime) WithConfig(cfg *runtime.RuntimeConfig) {
 	}
 
 }
+func (c *IgniteRuntime) WithKeepMgmtNet() {
+	c.ctrRuntime.WithKeepMgmtNet()
```
steiler, line 123: Can someone check if this makes sense? 
Do we need to do something else here?

hellt: ignite runtime uses the docker runtime for management network, so that seems ok
@networkop is that right?

hellt:
@

> Since we've gone downa different path now, with karims work on refactoring the runtime interface, should the withKeepMgmtNet() func now be replaced by jsut the field in the withConfig piece?

can you create a commit with that approach? Looks more clean to me

runtime/ignite/iginite.go
networkop: yes, it makes sense

steiler:
As discussed, we go with this approach for now and tackle a potential change in a different PR.

PR #469, opened 2021-06-22
implement Node as a Go interface{}

karimra:
continuing #449 here


cmd/tools_cert.go
```diff
@@ -167,11 +168,11 @@ func signCert(cmd *cobra.Command, args []string) error {
...
-	opts := []clab.ClabOption{
-		clab.WithDebug(debug),
-		clab.WithTimeout(timeout),
-	}
-	c := clab.NewContainerLab(opts...)
+	// opts := []clab.ClabOption{
+	// 	clab.WithDebug(debug),
+	// 	clab.WithTimeout(timeout),
+	// }
+	// c := clab.NewContainerLab(opts...)
```
hellt, line 175: safe to remove it seems

hellt:
@karimra I have tested a few non-node related deployments as well as added a few integration tests to cover those

- port forwarding
- host mode networking
- additional mgmt interfaces
- bind mounts
- manual mgmt ip assignments

I haven't tested nodes per se apart from srl, ceos, sros.

Since the tests pass, I propose we merge this, unless you have something you wanted to add

karimra:
Go ahead and merge it, if I missed something it can be added later.

Issue #467, opened 2021-06-21
cEOS and cgroups v2


nleiva:
This is just FYI in case someone else runs into this. 

cEOS images wouldn't run on a Fedora34 system. This is what the logs would show, after running the topology with Containerlab:

```bash
⇨  docker logs clab-mylab-router2
Failed to mount cgroup at /sys/fs/cgroup/systemd: Operation not permitted
systemd 219 running in system mode. (+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 -SECCOMP +BLKID +ELFUTILS +KMOD +IDN)
Detected virtualization docker.
Detected architecture x86-64.

Welcome to CentOS Linux 7 (AltArch)!

Set hostname to <router2>.
Cannot determine cgroup we are running in: No such file or directory
Failed to allocate manager object: No such file or directory
[!!!!!!] Failed to allocate manager object, freezing.
```

In a nutshell, Fedora31+ comes with cgroups v2 and reverting Systemd configuration to use cgroups v1 fixes the issue:

```bash
sudo grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=0"
reboot
```

Other systems like [RHEL 8 ship with cgroups v1](https://www.redhat.com/en/blog/world-domination-cgroups-rhel-8-welcome-cgroups-v2) as the default, so it's a non-issue there. 

I'm running Docker 20.10.7, which [supports cgroups v2 (both rootful and rootless)](https://github.com/docker/docker.github.io/pull/11042). So I'm not sure yet what it's causing this, but I thought I might share to save time and a headache to others.

hellt:
Hi @nleiva 
I confirm that, spun up fedora 33 and I see the same behaviour with `ceos:4.25.0F`
Since srlinux container doesn't experience any issues on the same fedora33, I take it as an issue with a ceos-Lab container.

I guess you have used the latest ceos container available? Because mine 4.25.0F is not of the latest version.

I propose we add a doc note on the ceos kind mentioning this limitation/known-issue, I will take it as an action point


I wonder if any aristians know anything about `cgroup v2` limitation, pinging @burnyd @sulrich @dharmbhai @noredistribution

sulrich:
huh. this whole cgroups v2 thing is news to me.  i'll do a bit of digging.  thanks for the heads up.

[...]

PR #470, opened 2021-06-23
update kind/ceos.md doc re: cgroup v1 requirement

sulrich:
reference issue: #467,  ceos-lab only supports cgroup v1 operation. this documentation update advises users that they may need to revert their runtime environment to accommodate this requirement.


hellt:
thanks @sulrich 
I moved this note to its own subchapter and did a few styling edits. But the rest is kept from your original commit
thanks for your contribution!

PR #481, opened 2021-06-28
fixing containerd filters

steiler:
adding additional escaping and fixing the filter string building


steiler:
Swapped in your cut orgy for a perl regex. ;-)
Now interface index is properly matched and can be found on the host.

hellt:
not ever in my wildest nightmares I had a line of perl code in the projects I take part in 🤣 

but thanks a lot, lets do this, maybe I will swap it later for regex-flavored grep

hellt:
I guess you also need to close that issue in containerd repo?

PR #482, opened 2021-06-29
Adding support for Cumulus with Ignite

networkop:
I think the title pretty much sums it up but since the PR is quite big, I'll try to break it down into multiple bullet points to better explain everything that has been affected:

* The main addition is the new runtime and node kind -- these are completely new files and don't change any of the existing interfaces or types
* Ability to specify runtime per node -- all nodes structs now have a private `runtime` field. This field is set during `node.Init` and defaults to the global runtime (as configured by the `--runtime` flag). However for `cvx` it defaults to a pre-defined `ignite` runtime with the ability to fall back to any other supported runtime via the `NodeConfig.Runtime`.
* As a result of the above, all of the calls to runtime (e.g. ListContainers) are now executed via a proxy functions defined for `CLab`, e.g. it is now `c.ListContainers(ctx, labels)` instead of `c.Runtime(ctx, labels)`. `CLab` always knows about all of the configured runtimes (since it knows about all of the nodes).
* Topology parsing is moved to `CLab.Init` function -- this way it's safer to reason about CLab since we know that it'll always have Nodes and Runtimes parsed and initialized.

I think overall the last 3 points help enforce the separation between different packages and prevent direct calls between, e.g. `cmd` and `runtime`.  It's much easier to read and troubleshoot code that relies on p2p interaction instead, i.e. `cmd` <-> `clab` <-> `node` <-> `runtime`.

AFAICT, none of the existing behavior has been affected, I've tried to make sure of that but happy to fix if there anything is off.


clab/config.go
```diff
@@ -36,7 +36,7 @@ const (
 	// veth link mtu. jacked up to 65k to allow jumbo testing of various sizes
 	defaultVethLinkMTU = 65000
 	// containerlab's reserved OUI
-	clabOUI = "aa:c1:ab"
+	clabOUI = "00:1c:73"
```
hellt, line 39: Hi @networkop 
why the change here?

nodes/vr_vmx/vr-vmx.go
```diff
@@ -59,17 +60,33 @@ func (s *vrVMX) PreDeploy(configName, labCADir, labCARoot string) error {
...
 }
 
-func (s *vrVMX) Deploy(ctx context.Context, r runtime.ContainerRuntime) error {
-	return r.CreateContainer(ctx, s.cfg)
+func (s *vrVMX) Deploy(ctx context.Context) error {
+	_, err := s.runtime.CreateContainer(ctx, s.cfg)
+	if err != nil {
+		return err
+	}
+	return nil
```
karimra, line 68: `return err` is enough here

nodes/vr_xrv/vr-xrv.go
```diff
@@ -58,17 +59,33 @@ func (s *vrXRV) PreDeploy(configName, labCADir, labCARoot string) error {
...
 }
 
-func (s *vrXRV) Deploy(ctx context.Context, r runtime.ContainerRuntime) error {
-	return r.CreateContainer(ctx, s.cfg)
+func (s *vrXRV) Deploy(ctx context.Context) error {
+	_, err := s.runtime.CreateContainer(ctx, s.cfg)
+	if err != nil {
+		return err
+	}
+	return nil
```
karimra, line 67: `return err` is enough here

nodes/vr_xrv9k/vr-xrv9k.go
```diff
@@ -61,17 +62,33 @@ func (s *vrXRV9K) PreDeploy(configName, labCADir, labCARoot string) error {
...
 }
 
-func (s *vrXRV9K) Deploy(ctx context.Context, r runtime.ContainerRuntime) error {
-	return r.CreateContainer(ctx, s.cfg)
+func (s *vrXRV9K) Deploy(ctx context.Context) error {
+	_, err := s.runtime.CreateContainer(ctx, s.cfg)
+	if err != nil {
+		return err
+	}
+	return nil
```
karimra, line 70: `return err` is enough here

runtime/ignite/iginite.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/weaveworks/ignite/pkg/operations"
+	"github.com/weaveworks/ignite/pkg/providers"
+	igniteDocker "github.com/weaveworks/ignite/pkg/providers/docker"
+	"github.com/weaveworks/ignite/pkg/providers/ignite"
+	"github.com/weaveworks/ignite/pkg/util"
+)
+
+const (
+	RuntimeName      = "ignite"
+	dockerRuntime    = "docker"
```
karimra, line 30: can probably use the docker runtime name from `nodes` rather than redeclaring the const

runtime/ignite/iginite.go
```diff
@@ -0,0 +1,404 @@
...
+	defaultTimeout   = 30 * time.Second
+	kvmPath          = "/dev/kvm"
+	dockerBridge     = "docker-bridge"
```
karimra, line 33: minor issue, but probably worth renaming `defaultNetworkPluginName`. I thought it was an actual bridge name

clab/clab.go
```diff
@@ -119,13 +125,18 @@ func NewContainerLab(opts ...ClabOption) *CLab {
...
 	for _, o := range opts {
 		o(c)
 	}
 
-	return c
+	if err := c.parseTopology(); err != nil {
+		return nil, err
+	}
+
+	return c, nil
```
karimra, line 139: `return c, err` should be enough

clab/config.go
```diff
@@ -161,6 +172,9 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 		Publish:         c.Config.Topology.GetNodePublish(nodeName),
 		Sysctls:         make(map[string]string),
 		Endpoints:       make([]*types.Endpoint, 0),
+		Sandbox:         nodeDef.Sandbox,
+		Kernel:          nodeDef.Kernel,
+		Runtime:         nodeDef.Runtime,
```
karimra, line 177: I think these 3 items need functions like `c.Config.Topology.GetNodeSandbox()`, `c.Config.Topology.GetNodeKernel()`, and `c.Config.Topology.GetNodeRuntime()`.

To be able to have a node inherit the values from the `kinds` and `defaults` sections.

clab/config.go
```diff
@@ -321,24 +335,24 @@ func (c *CLab) verifyLinks() error {
...
+			images[imageName] = node.GetRuntime().GetName()
 		}
 
-		images[node.Config().Image] = struct{}{}
 	}
 
-	for image := range images {
-		err := c.Runtime.PullImageIfRequired(ctx, image)
+	for image, runtimeName := range images {
+		err := c.Runtimes[runtimeName].PullImageIfRequired(ctx, image)
```
karimra, line 355: are we always sure that `c.Runtimes[runtimeName]` always exist and is not `nil` ?

karimra:
@networkop , did a first review, let me know if anything doesn't make sense to you.

If I got it right, cumulus uses ignite as a default runtime, but it supports docker ?
This could be confusing for users who will have to set the runtime explicitly to `docker` just for cvx nodes.

networkop:
thanks @karima, will get the changes in shortly.

regarding the default behavior, with `ignite` you get 100% of cumulus features, which is why I made it a default behaviour. Ultimately, from user's pov it shouldn't be noticeable what runtime is used under the hood, they would still be able to interact with it the same exact way.

clab/clab.go
```diff
@@ -247,37 +262,58 @@ func (c *CLab) CreateLinks(ctx context.Context, workers uint, postdeploy bool) {
...
+}
+
+func (c *CLab) GetNodeRuntime(query string) (runtime.ContainerRuntime, error) {
+	for _, node := range c.Nodes {
+		if node.GetName() == query {
+			return node.GetRuntime(), nil
+		}
+	}
+	return nil, fmt.Errorf("could not find a container matching name %q", query)
+}
```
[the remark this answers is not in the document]
networkop: The problem here is that query is a `NodeConfig.LongName` but map's key is a `NodeConfig.ShortName`

cmd/deploy.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/srl-labs/containerlab/cert"
 	"github.com/srl-labs/containerlab/clab"
 	"github.com/srl-labs/containerlab/nodes"
+	"github.com/srl-labs/containerlab/runtime/ignite"
```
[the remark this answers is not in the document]
networkop: this makes sense. will update this everywhere

nodes/node.go
```diff
@@ -41,10 +42,15 @@ type Node interface {
...
-	Deploy(context.Context, runtime.ContainerRuntime) error
-	PostDeploy(context.Context, runtime.ContainerRuntime, map[string]Node) error
+	Deploy(context.Context) error
+	PostDeploy(context.Context, map[string]Node) error
 	WithMgmtNet(*types.MgmtNet)
-	SaveConfig(context.Context, runtime.ContainerRuntime) error
+	WithRuntime(runtime.ContainerRuntime)
+	SaveConfig(context.Context) error
+	Delete(context.Context) error
+	GetImages() []string
```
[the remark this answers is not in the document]
networkop: why do you think the order is important? We still build the `map[string]string` inside the `VerifyImages` before we pull them, so that should take care of any duplicates.

nodes/node.go
```diff
@@ -41,10 +42,15 @@ type Node interface {
...
+	GetName() string
+	GetRuntime() runtime.ContainerRuntime
```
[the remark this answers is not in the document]
networkop: this is one of the things that allows nodes to override runtime. Nodes have runtime as a private field and `GetRuntime` is a getter function.

runtime/runtime.go
```diff
@@ -44,7 +44,12 @@ type ContainerRuntime interface {
 	// ExecNotWait executes cmd on container identified with id but doesn't wait for output nor attaches stodout/err
 	ExecNotWait(context.Context, string, []string) error
 	// Delete container by its name
-	DeleteContainer(context.Context, *types.GenericContainer) error
+	DeleteContainer(context.Context, string) error
+	// Getters for config options
+	GetDebug() bool
+	GetTimeout() time.Duration
+	GetGracefulShutdown() bool
+	GetName() string
```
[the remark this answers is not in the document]
networkop: yeah, maybe. is that gonna be safe though? this opens up to external code modifying these values.

clab/config.go
```diff
@@ -103,6 +104,15 @@ func (c *CLab) ParseTopology() error {
 		// i represents the endpoint integer and l provide the link struct
 		c.Links[i] = c.NewLink(l)
 	}
+
+	// update runtimes in case overridden by nodes
+	for _, node := range c.Nodes {
+		if node.GetRuntime() == nil {
+			continue
+		}
+		c.Runtimes[node.GetRuntime().GetName()] = node.GetRuntime()
```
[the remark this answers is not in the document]
networkop: `nil` runtime check is for `config_test.go` tests, some of them set up `CLab` without  a runtime. 
I thought this was a quick solution but I can also refactor tests to do `WithRuntime`  instead. wdyt?

I'll add the if exists check 👍

runtime/ignite/iginite.go
networkop: this is a default container runtime for ignite (to run its sandbox container), so it won't come from `nodes`. Unless I misunderstood you.

runtime/ignite/iginite.go
networkop: yep, makes sense

clab/config.go
networkop: Yes, in this case `runtimeName` comes from `node.GetRuntime().GetName()`, so this means runtime is already there (since it responds to `GetName`)
And `nodes` all have a runtime thanks to this 
```go
err = n.Init(nodeCfg, nodes.WithRuntime(c.GlobalRuntime()), nodes.WithMgmtNet(c.Config.Mgmt))
```

clab/config.go
networkop: ok, yep, makes sense

types/types.go
```diff
@@ -53,8 +53,8 @@ type MgmtNet struct {
 
 // NodeConfig is a struct that contains the information of a container element
 type NodeConfig struct {
-	ShortName        string
-	LongName         string
+	ShortName        string // node of the Node inside topology YAML
```
hellt, line 56: **name** of the node

networkop: haha, yep. btw, while we're at it, how about we rename `ShortName` to something like `TopoName` and `LondName` to `ContainerName`? or something similar to give a bit more context to these variables?

nodes/node.go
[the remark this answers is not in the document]
karimra: I mean order is not important, but seeing a `[]string` implies that the order is important.
The idea data struct here is a set not a list, since that doesn't exist in Go, the next best thing is a map.

nodes/node.go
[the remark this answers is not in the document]
karimra: You mean a node creates the runtime if overwritten and that's what gets added to the `c.Runtimes` map ?
This means potential multiple inits of runtimes

Wouldn't it be better to create all required runtimes before init of nodes, the set of configured runtimes can be retrieved from the `Topology` struct.
Then change `nodes.WithRuntime(runtime.ContainerRuntime)` to become `nodes.WithRuntimes(map[string]runtime.ContainerRuntime)`.
the node would then store whichever runtime it needs.

This would also allow `cvx` to use `docker` runtime if `ignite` is not installed, with a warning ? But not sure if we want that.

runtime/runtime.go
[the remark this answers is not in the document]
karimra: true, then let's expose the value not the ref, 
`Config() runtime.RuntimeConfig` instead of 
`Config() *runtime.RuntimeConfig`

runtime/ignite/iginite.go
karimra: yes, I meant that the `string` value `nodes.DockerRuntimeName` can be used.

networkop: I've reworked it slightly to use constants from ignite instead. 
```go
defaultContainerRuntime       = igniteRuntimes.RuntimeDocker
defaultContainerNetworkPlugin = igniteNetwork.PluginDockerBridge
```
This way it's consistent for both runtime and network plugin and also indicates where this dependency is coming from. Hope that's ok?

nodes/node.go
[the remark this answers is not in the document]
networkop: The issue is that we'd need to let nodes `Init` in order to know if they decide to overwrite the runtime. I'll try to move the `rInit()` out of the node's `Init` function.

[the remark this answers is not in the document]
networkop: @karimra I've changed this to `WithRuntime(globalRuntime string, allRuntimes map[string]runtime.ContainerRuntime)` and updated the PR. wdyt?

clab/clab.go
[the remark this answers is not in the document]
hellt: another option is to split the `LongName` by `labName` substring, this will produce a slice with two elements, where `[1]` is the element with the `ShortName`

clab/config.go
```diff
@@ -321,24 +335,24 @@ func (c *CLab) verifyLinks() error {
...
-		// skip image verification for bridge kinds
-		if node.Config().Kind == "bridge" || node.Config().Kind == "ovs-bridge" {
-			continue
-		}
-		if node.Config().Image == "" {
-			return fmt.Errorf("missing required image for node %s", node.Config().ShortName)
+
+		for _, imageName := range node.GetImages() {
+			if imageName == "" {
+				log.Errorf("missing required image for node %q", node.Config().ShortName)
```
hellt, line 347: shouldn't we return the error here instead of logging it?
if one of the required images is missing, the node has no chances to start, so we should stop here?

cmd/deploy.go
```diff
@@ -99,45 +99,37 @@ var deployCmd = &cobra.Command{
...
-
-		// generate graph of the lab topology
-		if graph {
-			if err = c.GenerateGraph(topo); err != nil {
-				log.Error(err)
+		// Serializing ignite workers due to busy device error
+		for _, node := range c.Nodes {
+			if node.GetRuntime().GetName() == runtime.IgniteRuntime {
+				nodeWorkers = 1
+				break
```
hellt, line 121: do we need a loop here? can we maybe have a look at `c.Runtimes` map and see if `runtime.IgniteRuntime` key is there?

cmd/deploy.go
```diff
@@ -152,16 +144,30 @@ var deployCmd = &cobra.Command{
...
+			return err
+		}
+
+		// generate graph of the lab topology
+		if graph {
+			if err = c.GenerateGraph(topo); err != nil {
+				log.Error(err)
+			}
+		}
+		log.Debug("containers created, retrieving state and IP addresses...")
```
hellt, line 167: this log message seems to have been moved not on purpose

cmd/deploy.go
```diff
@@ -99,45 +99,37 @@ var deployCmd = &cobra.Command{
...
-		c.CreateNodes(ctx, nodesMaxWorkers)
-		c.CreateLinks(ctx, linksMaxWorkers, false)
-
-		// generate graph of the lab topology
-		if graph {
-			if err = c.GenerateGraph(topo); err != nil {
-				log.Error(err)
+		// Serializing ignite workers due to busy device error
+		for _, node := range c.Nodes {
+			if node.GetRuntime().GetName() == runtime.IgniteRuntime {
```
hellt, line 119: I wonder, we import the whole runtime package because of a single constant. Just wondering, to make a cleaner interface separation, wouldn't it be better to have the number of workers to come from some runtime method?

clab/config.go
networkop: For some reason, I thought that there could be nodes with no images. Like bridge?

cmd/deploy.go
networkop: yep, you're right. will fix that

cmd/deploy.go
networkop: yep, will fix

nodes/cvx/cvx.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (c *cvx) WithMgmtNet(*types.MgmtNet) {}
+func (c *cvx) WithRuntime(globalRuntime string, allRuntimes map[string]runtime.ContainerRuntime) {
+
+	// fallback to the global runtime if overridden
+	if c.Config().Runtime != "" {
+		c.runtime = allRuntimes[globalRuntime]
+		return
+	}
```
karimra, line 94: why not `c.runtime = allRuntimes[c.Config().Runtime]` ?

nodes/cvx/cvx.go
```diff
@@ -0,0 +1,127 @@
...
+		defaultConfig := allRuntimes[globalRuntime].Config()
+
+		err := c.runtime.Init(
+			runtime.WithConfig(&defaultConfig),
+		)
+		if err != nil {
+			log.Fatalf("failed to init the container runtime: %s", err)
+		}
+
+	}
```
karimra, line 115: I think the idea was not to init the runtime per node, the current code will result in a runtime init for each cvx node.

nodes/node.go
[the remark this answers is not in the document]
karimra: this works as well, making it generic for the other nodes kind can be added in another PR I think.

Remains the issue of initializing the runtimes before the nodes, current cvx code creates an ignite client for each node.

By going through the topology struct, it's possible to get a `map[string]struct{}` holding the specified runtimes names under nodes/kinds/default as keys, adding the global one, initializing them as well as the global one and passing them as a map to the node init.

clab/config.go
karimra: I saw you added `nodeDef.GetSandbox()`,... but not the get methods from `Topology`

cmd/destroy.go
```diff
@@ -156,26 +161,37 @@ func deleteEntriesFromHostsFile(containers []types.GenericContainer, bridgeName
...
 
-	// get lab directory used by this lab to remove it later if cleanup is used
 	var labDir string
+	for _, ctr := range containers {
+		if n, ok := c.Nodes[ctr.ShortID]; ok {
+			c.Nodes[ctr.ShortID] = n
+			labDir = n.Config().LabDir
+		}
+	}
+
```
hellt, line 176: What was the idea here to swap the `containers[0].Labels["clab-node-lab-dir"]` as a way to derive the lab directory?

cmd/disableTxOffload.go
```diff
@@ -27,17 +27,25 @@ var disableTxOffloadCmd = &cobra.Command{
...
+		}
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
 		log.Infof("getting container '%s' information", cntName)
 
-		NSPath, err := c.Runtime.GetNSPath(ctx, cntName)
+		nodeRuntime, err := c.GetNodeRuntime(cntName)
+		if err != nil {
+			log.Fatal(err)
```
hellt, line 41: should we return the error, rather than `log.Fatal`?

cmd/deploy.go
karimra: For this, I thought that in a future PR we can create two groups of nodes, the ones using `ignite` and the others.
Create the nodes with `ignite` runtime sequentially then the others with the configured workers numbers.

cmd/exec.go
```diff
@@ -81,7 +86,13 @@ var execCmd = &cobra.Command{
 			if cont.State != "running" {
 				continue
 			}
-			stdout, stderr, err := c.Runtime.Exec(ctx, cont.ID, cmds)
+
+			nodeRuntime, err := c.GetNodeRuntime(cntName)
+			if err != nil {
+				log.Fatal(err)
```
hellt, line 92: should we return the error instead of log.Fatal?

docs/manual/kinds/cvx.md
```diff
@@ -0,0 +1,72 @@
...
+
+When running inside the `ignite` runtime, all mount binds work one way -- from host OS to the cvx node, but not the other way around. Currently, it's up to a user to manually update individual files if configuration updates need to be persisted.
+This will be addressed in the future releases.
+
+
+## Lab examples
+The following labs feature CVX node:
+
+- [Cumulus Linux Test Drive](https://github.com/srl-labs/containerlab/blob/master/lab-examples/lab-start.yml)
+- [Cumulus and FRR](https://github.com/srl-labs/containerlab/blob/master/lab-examples/topo.yml)
```
hellt, line 67: looks like these example URLs are missing the directory they are in

nodes/bridge/bridge.go
```diff
@@ -29,13 +30,32 @@ func (s *bridge) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
...
+
+func (s *bridge) SaveConfig(ctx context.Context) error {
+	return nil
+}
+
+func (s *bridge) GetImages() map[string]string {
+	images := make(map[string]string)
+	images[nodes.ImageKey] = s.cfg.Image
+	return images
+}
```
hellt, line 57: bridge kind doesn't have an associated image

nodes/crpd/crpd.go
```diff
@@ -82,10 +84,24 @@ func (s *crpd) PostDeploy(ctx context.Context, r runtime.ContainerRuntime, ns ma
 	return err
 }
 
+func (s *crpd) GetImages() map[string]string {
+	images := make(map[string]string)
+	images[nodes.ImageKey] = s.cfg.Image
+	return images
+}
```
hellt, line 91: wdyt about rewriting this as

```go
return map[string]string{
  nodes.ImageKey: s.cfg.Image,
}
```

nodes/host/host.go
```diff
@@ -31,12 +34,31 @@ func (s *host) Config() *types.NodeConfig { return s.cfg }
...
+func (s *host) PostDeploy(ctx context.Context, ns map[string]nodes.Node) error {
 	return nil
 }
 
+func (s *host) GetImages() map[string]string {
+	images := make(map[string]string)
+	images[nodes.ImageKey] = s.cfg.Image
+	return images
+}
+
```
hellt, line 47: host node doesn't have an associated image

nodes/mysocketio/mysocketio.go
```diff
@@ -33,26 +36,41 @@ func (s *mySocketIO) PreDeploy(configName, labCADir, labCARoot string) error {
 	// utils.CreateDirectory(s.cfg.LabDir, 0777)
 	return nil
 }
+func (s *mySocketIO) Deploy(ctx context.Context) error { return nil }
```
hellt, line 39: why the creation of container was removed here?

nodes/ovs/ovs.go
```diff
@@ -34,14 +35,31 @@ func (l *ovs) Config() *types.NodeConfig { return l.cfg }
...
-func (s *ovs) SaveConfig(ctx context.Context, r runtime.ContainerRuntime) error {
+func (s *ovs) Delete(ctx context.Context) error {
+	return nil
+}
+
+func (s *ovs) GetImages() map[string]string {
+	images := make(map[string]string)
+	images[nodes.ImageKey] = s.cfg.Image
+	return images
+}
```
hellt, line 62: ovs bridge doesn't have an image associated with it

runtime/docker/docker.go
```diff
@@ -229,15 +230,15 @@ func (c *DockerRuntime) DeleteNet(ctx context.Context) (err error) {
 }
 
 // CreateContainer creates a docker container
-func (c *DockerRuntime) CreateContainer(ctx context.Context, node *types.NodeConfig) (err error) {
+func (c *DockerRuntime) CreateContainer(ctx context.Context, node *types.NodeConfig) (interface{}, error) {
```
hellt, line 233: @networkop 
can you please add a docstring here exaplaining why `interface{}` got added to the signature?

runtime/ignite/iginite.go
```diff
@@ -0,0 +1,402 @@
...
+	genericCtrs, err := c.produceGenericContainerList([]*api.VM{vm})
+	if err != nil {
+		return result, err
+	}
+	if len(genericCtrs) != 1 {
+		return result, fmt.Errorf("unexpected number of matched containers %d", len(genericCtrs))
+	}
+
+	return &genericCtrs[0], nil
+}
```
hellt, line 316: the GetContainer doesn't seem to be part of the Runtime interface
Why was it added?

nodes/node.go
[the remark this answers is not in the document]
networkop: I don't think runtimes are initialized multiple times now. Due to this 
```
	// By default, running in ignite runtime
	if igniteRuntime, ok := allRuntimes[runtime.IgniteRuntime]; ok {
		c.runtime = igniteRuntime
		return
	}
```

I get your point that doing this in a single place is easier to understand and I'd like to do this too but the problem I have is that runtime override is a node-specific option, i.e. some nodes may decide to override it and the logic is inside that node's package. By simply reading the topology file, you won't know what runtimes nodes would select, unless it's explicitly specificed in `nodeCfg.Runtime`.

clab/config.go
networkop: oh, sry, I misunderstood. do you mean something like this?
```
func (t *Topology) GetNodeSandbox(name string) string {
	if ndef, ok := t.Nodes[name]; ok {
		if ndef.GetNodeSandbox() != "" {
			return ndef.GetNodeSandbox()
		}
		if t.GetKind(t.GetNodeKind(name)).GetNodeSandbox() != "" {
			return t.GetKind(t.GetNodeKind(name)).GetNodeSandbox()
		}
		return t.GetDefaults().GetNodeSandbox()
	}
	return ""
}
```

types/types.go
hellt: m, TopoName to me feels more like the name of the topology
If I were changing these two fields, I'd go maybe with `Name` for a short name and `FullName` or `FQN` for a fully qualified name

@karimra how do you feel about Short/Long name pair and the possible alternatives? For us I guess the context is clear, thus it doesn't mix too much which is which

cmd/deploy.go
hellt: good idea, let me capture this in the following issue https://github.com/srl-labs/containerlab/issues/493

clab/config.go
hellt: in that case the nil will be the vaule of that slice, and the range won't engage?

cmd/deploy.go
networkop: Does this matter that much? Even when you use `wg` containers still get created sequentially by `runc` (afair). 
There's not much of a difference in terms of time, maybe seconds.

nodes/cvx/cvx.go
networkop: good point. fixed that.

nodes/cvx/cvx.go
networkop: there's a section above

```
	if igniteRuntime, ok := allRuntimes[runtime.IgniteRuntime]; ok {
		c.runtime = igniteRuntime
		return
	}
```

That checks if this runtime already exists

cmd/destroy.go
networkop: hm... i don't remember doing this. I'll revert back to how it is now in main.

cmd/disableTxOffload.go
networkop: yep, fixing that.

cmd/exec.go
networkop: yep

docs/manual/kinds/cvx.md
networkop: Yes because the PR is not merged yet.

runtime/ignite/iginite.go
networkop: I added it initially as a way to ListContainers and probably forgot to remove it.

cmd/deploy.go
hellt: the time will grow with the number of nodes you have in the topo
for a basic 8 node lab the sequential mode takes 14sec, whereas the concurrent takes 7sec

if we can make things faster without investing much in it, I don't see why we shouldn't

docs/manual/kinds/cvx.md
hellt: How can we merge this PR if it currently has a wrong links here?
I meant the links should have `cvx03` after the `lab-examples` path elem

networkop: oh yeah, I missed that. I'll get it fixed

nodes/crpd/crpd.go
networkop: yep, no probs

nodes/host/host.go
networkop: fixed

nodes/mysocketio/mysocketio.go
networkop: oops. fixed

nodes/ovs/ovs.go
networkop: fixed.

clab/clab.go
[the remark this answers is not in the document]
networkop: I've done that with a helper funciton.

clab/config.go
networkop: I've re-worked this to return error instead. Inside `GetImages` I make sure I always return empty map for nodes that have no images.

cmd/deploy.go
networkop: yeah, fair enough.

runtime/docker/docker.go
networkop: done.

networkop:
@karimra @hellt all fixes have been pushed.

clab/config.go
karimra: yes, exactly

types/types.go
karimra: maybe just `Name` and `ContainerName` ?

nodes/cvx/cvx.go
karimra: But how would the ignite runtime end up in the `allRuntimes` map?

in `c.parseTopology`, `c.NewNode` gets called before the runtimes overwritten by the nodes are added to `c.Runtimes`

My suggestion is to create the required runtimes before initializing the nodes.

runtime/docker/docker.go
networkop: I've added it to the runtime interface instead

nodes/cvx/cvx.go
karimra: I see it now

nodes/node.go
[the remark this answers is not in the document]
karimra: I see, I think we can add that in a future PR, 
getting a default runtime for a node kind without initializing the node.

nodes/cvx/cvx.go
networkop: Let me try this now

karimra:
@networkop thanks for addressing all my comments, lgtm.

networkop:
thanks @karimra . Can you check the latest commit? I think it would make sure runtimes are initialized only once.

karimra:
Yes, even better, a single runtime passed to the node Init. Thanks

clab/config.go
```diff
@@ -557,3 +608,11 @@ func sysMemory(v string) uint64 {
...
 	return m
 }
+
+func parseLongName(input string) (string, string, string, error) {
+	result := strings.Split(input, "-")
+	if len(result) != 3 {
+		return "", "", "", fmt.Errorf("failed to parse container name %q", input)
+	}
+	return result[0], result[1], result[2], nil
+}
```
hellt, line 618: I'm afraid this won't work
a lab/node name can have dashes, so you can have `clab-my-lab-my-node` full name
so the way to parse the long name you need to take in the lab name, and split on that

nodes/cvx/cvx.go
```diff
@@ -0,0 +1,99 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/srl-labs/containerlab/nodes"
+	"github.com/srl-labs/containerlab/runtime"
+	"github.com/srl-labs/containerlab/types"
+	"github.com/weaveworks/ignite/pkg/operations"
+)
+
+var (
+	defaultCvxKernelImageRef  = "docker.io/networkop/kernel:4.19"
+	defaultIgniteSandboxImage = "networkop/ignite:dev"
```
hellt, line 15: in short, what is specific about this sandbox?

networkop: you mean why I host it in my repo? The way ignite works is they only build containers once they've shipped a release, which they haven't done yet, so I'm using a temp image hosted in my account. Once they make the next release, I'll update this to point to their image:tag.

hellt: aha, clear. that is what I though, but wanted to double check in case anyone ask

hellt:
@networkop In 362dd1b I removed a few leftovers of log.Fatal

networkop:
thanks @hellt . I also pushed a fix the shortname parsing logic.

types/types.go
hellt: that sounds good @karimra 
we can do this in a separate PR

hellt (approved):
Solid work @networkop 
I hope you had fun time ;) that is one of the biggest PRs in this repo

networkop:
awesome! thanks for reviewing it @karimra @hellt

PR #489, opened 2021-07-01
Update Base SROS template

kellerza:
This seems to have gotten lost during the file renames


hellt:
yes, I removed it originally, because I wanted to create somewhat generic template per function (interfaces, ospf, bgp), but then we first need to tackle more serious stuff that we go through in #487 
so let's have it now as it is, and then we will see how this evolves

kellerza:
ok, thanks. Almost thought I lost this when I rebased that branch 😅

PR #491, opened 2021-07-05
config: No text/template wrapper

kellerza:
Nr 1 from #487 

Implementation similar to the existing logic from https://github.com/kellerza/template/blob/main/template.go#L59

@karimra mentioned there is a built in method to search the paths, but did not find it.

### DIFF
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -84,7 +84,7 @@ func RenderAll(nodes map[string]nodes.Node, links map[int]*types.Link) (map[stri
 			}
 
 			var buf strings.Builder
-			err := tmpl.Lookup(tmplN).Execute(&buf, vars)
+			err := tmpl.ExecutTemplate(&buf, tmplN, vars)
 			if err != nil {
 				return nil, err
 			}
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -3,7 +3,6 @@ package config
 import (
 	"encoding/json"
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 
@@ -31,20 +30,16 @@ type NodeConfig struct {
 	Info []string
 }
 
-func LoadTemplate(tmpl *template.Template, name string) error {
-	for i := len(TemplatePaths) - 1; i >= 0; i-- {
-		fn := filepath.Join(TemplatePaths[i], name)
-		_, err := tmpl.ParseFiles(fn)
-		if os.IsNotExist(err) { // try in next path
-			continue
-		}
+// Load templates from all paths for the specific role/kind
+func LoadTemplates(tmpl *template.Template, role string) error {
+	for i := range TemplatePaths {
+		fn := filepath.Join(TemplatePaths[i], fmt.Sprintf("*__%s.tmpl", role))
+		_, err := tmpl.ParseGlob(fn)
 		if err != nil {
-			return fmt.Errorf("could not load template %s: %s", fn, err)
+			return fmt.Errorf("could not load templates from %s: %s", fn, err)
 		}
-		log.Debugf("template loaded %d. %s %s\n", i, name, fn)
-		return nil
 	}
-	return fmt.Errorf("could not find template %s in search path", name)
+	return nil
 }
 
 func RenderAll(nodes map[string]nodes.Node, links map[int]*types.Link) (map[string]*NodeConfig, error) {
@@ -77,9 +72,12 @@ func RenderAll(nodes map[string]nodes.Node, links map[int]*types.Link) (map[stri
 			tmplN := fmt.Sprintf("%s__%s.tmpl", baseN, vars["role"])
 
 			if tmpl.Lookup(tmplN) == nil {
-				err := LoadTemplate(tmpl, tmplN)
+				err := LoadTemplates(tmpl, tmplN)
 				if err != nil {
 					return nil, err
+				}
+				if tmpl.Lookup(tmplN) == nil {
+					return nil, fmt.Errorf("template not found %s", tmplN)
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


karimra:
you can do something like this:

```go
func (c *CLab) parseTemplates(pathToTemplates string) (*template.Template, error) {
    t := template.New(c.Config.Name)
    err := filepath.Walk(pathToTemplates, func(path string, info os.FileInfo, err error) error {
        if strings.HasSuffix(path, ".gotmpl") { // or any other condition identifying the template files
            _, err = t.ParseFiles(path)
           return err
    })

    return t, err
}
```

The individual templates can be called with their name, which defaults to the file base name.

```go
t.ExecuteTemplate(os.Stdout, name, vars)
```

you can get the list of templates associated with `t`:
```go
t.Templates()
```

you can lookup a specific template associated with `t`:
```go
t.Lookup(name)
```

clab/config/template.go
```diff
@@ -58,11 +75,21 @@ func RenderAll(nodes map[string]nodes.Node, links map[int]*types.Link) (map[stri
...
+
+			if tmpl.Lookup(tmplN) == nil {
+				err := LoadTemplate(tmpl, tmplN)
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			var buf strings.Builder
+			err := tmpl.Lookup(tmplN).Execute(&buf, vars)
```
kellerza, line 87:
```suggestion
			err := tmpl.ExecutTemplate(&buf, tmplN, vars)
```

kellerza:
Thanks @karimra. Changed to `ExecuteTemplate`

Not sure walking the root path is applicable. The idea here is to look for a specific file if you have a list of possible paths.

clab/config/template.go
```diff
@@ -58,11 +75,21 @@ func RenderAll(nodes map[string]nodes.Node, links map[int]*types.Link) (map[stri
...
+
+			if tmpl.Lookup(tmplN) == nil {
+				err := LoadTemplate(tmpl, tmplN)
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			var buf strings.Builder
+			err := tmpl.ExecutTemplate(&buf, tmplN, vars)
```
kellerza, line 87:
```suggestion
			err := tmpl.ExecuteTemplate(&buf, tmplN, vars)
```

karimra:
It's still possible to parse the list of directories, and call the desired template by name.
I'm not sure about the logic you want to implement.

Parsing the whole directory, helps combining templates from different files as well.

kellerza:
If we need templates from multiple files that is a requirement I did not have in mind. Do you see this as a requirement?

It seems quite ineffective, because of the possible amount of templates in a folder. You will end up loading all templates, for all kinds, over all folders specified.

If that is indeed the case, we should just run `ParseGlob` over all folders in order. From the docs:

> When parsing multiple files with the same name in different directories, the last one mentioned will be the one that results.  

Maybe with globs we could keep some of the existing logic, but only load all templates for the active role/kind

karimra:
Templates can be combined by importing each other. In that case you would need to load the main template as well as the ones it imports.

Separating templates in multiples files allows reuse and separation of responsibilities/concerns, it's really at the core of the template engine design.

I don't think this has to be addressed in this PR, since we didn't really come to a conclusion concerning how templates will be mapped to kinds and nodes.

Can explain how a template file is mapped to a node in this PR? Is it a file name that gets inherited from kinds/defaults like the other variables?

kellerza:
You can pass a list of templates names. The template names will be combined with the role: `<tname>__<role>.tmpl`

By default the role is the kind, except if you specify it in vars.

hellt:
> You can pass a list of templates names. The template names will be combined with the role: `<tname>__<role>.tmpl`
> 
> By default the role is the kind, except if you specify it in vars.

yes, the idea here is that the template name is not related to a single node, but to a group of them. The default group marker is its `kind`.

So for a lab that has `srl` and `sros` kinds, this will allow to pass a template named `interfaces` files `interfaces__srl.tmpl` and `interfaces__sros.tmpl` will be looked up and applied to the relevant nodes

kellerza:
I would phrase the *group of them* slightly differently:

- The template <name> gets applied to *all* nodes in the topology file.
- The role/kind decides which template (SRL syntax,SROS syntax, etc) get rendered for that node.

Any finer control can be done inside the template with if constructs using even more of the vars on the node/link level.

kellerza:
I would add that the `role` variable creates groups: 
1. firstly each node kind will have its own CLI syntax (default=kind), but
2. nodes can also be grouped into roles (like `spine` or `leaf` roles, or even `leaf_srl` and `leaf_sros`)

karimra:
it seems all this assumes that templates are being run with separate command, not run on deploy ?
can you show a topology file example where the templates are specified and the command to apply them ?

kellerza:
Indeed, config only deploys to running labs.
Vr-05 example folder has 2 examples and the base templates are in the active branch

try: 

`clab config -t topo.yml -n base -p ../templates/config`

add -c 100 to view the rendered templates if you dont have an active lab

karimra:
ok, if this is the case currently, `ParseGlob` works for now. I guess the glob can be adjusted to the pattern that matches the node name/role

hellt:
@karimra do you have anything else you had in mind for this PR?
otherwise we can merge it

karimra:
it's still hard for me to see the whole picture, go ahead. once the other PR is done it will hopefully be clearer.

kellerza:
I'll make some slides to capture the whole picture, I've been asked to present it to an internal community as well. 

This should also give a start for the site/docs.

karimra:
I get how it works and how to use it, what I don't get is the Go code. 
There seems to be a lot of assumptions on the template format and nature.

PR #495, opened 2021-07-06
config: Optional vars & add all nodes / "topology" to vars

kellerza:
Nr 2 & some parts of 3 of #487


templates/base__vr-sros.tmpl
```diff
@@ -1,142 +1,142 @@
-{{ expect .systemip "ip" }}
```
kellerza, line 1: seems I added this with CRLF through the web interface 🤦

clab/config/helpers.go
```diff
@@ -12,30 +12,3 @@ func SplitTrim(s string) []string {
...
-// The new agreed node config
-type NodeSettings struct {
-	Vars      map[string]string
-	Transport string
-	Templates []string
-}
-
-// Temporary function to extract NodeSettings from the Labels
-// In the next phase node settings will be added to the clab file
-func GetNodeConfig(labels map[string]string) NodeSettings {
```
kellerza, line 25: This was a temporary function (before vars was there) and really not used

clab/config/template.go
```diff
@@ -43,21 +47,15 @@ func LoadTemplates(tmpl *template.Template, role string) error {
...
 
-	if len(TemplatePaths) == 0 {
-		return nil, fmt.Errorf("please specify one of more paths with --template-path")
+	if len(TemplatePaths) == 0 { // default is the install path
+		TemplatePaths = []string{"@"}
 	}
 
 	if len(TemplateNames) == 0 {
-		var err error
-		TemplateNames, err = GetTemplateNamesInDirs(TemplatePaths)
```
kellerza, line 54: Thinking about this again, I'm not sure this is the best approach: 
- template paths contain many templates
- templates can be for different roles, so you might miss completely
- templates are supposed to be applied in order

so auto-discovery from file will not be reliable/cause confusion

clab/config/utils.go
```diff
@@ -13,7 +12,16 @@ import (
 )
 
 const (
```
kellerza, line 14: This is a short summary of the keys, that are optionally supplied and in the resulting vars sent to the template

clab/config/template.go
hellt: the idea for `GetTemplateNamesInDirs` is only to kick in if the template names were not supplied by a user

I see this to be a common use cases:
1. you either have just one template in a folder that you want to execute. In that case your template path points to a certain dir and in this dir you have a single template file. No need to ask a user to specify this template name, it will be discovered
2. you have many templates which are stacked and represent a step-by-step configuration process. Since the templates are stacked, I see it is convenient to name them in the order so that you see which one follows the other. For example: 010-ifaces, 020-igp, 030-bgp  
  this will allow the function to follow the order that is assumed by a template designer

kellerza: 👍 ok, the second use case is indeed something that should be very useful. 

For the base templates it could result in a great amount of things being deployed in no order, but luckily its quick to revert anything using config (just start fresh)

kellerza: added this back and updated the function not to include duplicates or blanks

clab/config/utils.go
karimra: why not define these as functions that can be used within the template ?

kellerza: node / nodes / links / role / far really are mechanisms to expose the topology to the template, they have no alternative and should generally not be used in a node's `vars:` (role is probably the exception for more advanced use-caese)

ip / name / linknr are convenience `vars` to render templates, These are obviously optional, but used in most networks (except if you go unnumbered). It would also be really cool to visualize / `graph` them.

Note: The second set can be templated, but it would increase complexity a lot for this first drop.

Also, generators (like IP pools, unique numbers) will probably end up in the same place (while preparing vars) They might not have fixed names, but the point is that I dont see it as part of the template, as they would need to keep state across templates and even across different executions and are used by different templates.

karimra: ok probably not in this PR

clab/config/template.go
```diff
@@ -33,7 +33,11 @@ type NodeConfig struct {
 // Load templates from all paths for the specific role/kind
 func LoadTemplates(tmpl *template.Template, role string) error {
 	for i := range TemplatePaths {
-		fn := filepath.Join(TemplatePaths[i], fmt.Sprintf("*__%s.tmpl", role))
+		p := TemplatePaths[i]
```
hellt, line 36: @kellerza any specific reason for not using

```go
for _, p := range TemplatePaths {
}
```
(eliminating the extra step of `p := TemplatePaths[i]`)

clab/config/template.go
```diff
@@ -43,10 +47,11 @@ func LoadTemplates(tmpl *template.Template, role string) error {
 }
 
 func RenderAll(nodes map[string]nodes.Node, links map[int]*types.Link) (map[string]*NodeConfig, error) {
+	// A map with the ShortName as the key
 	res := make(map[string]*NodeConfig)
 
-	if len(TemplatePaths) == 0 {
-		return nil, fmt.Errorf("please specify one of more paths with --template-path")
+	if len(TemplatePaths) == 0 { // default is the install path
+		TemplatePaths = []string{"@"}
```
hellt, line 54: the `@` char is used here and not the absolute path or a string const because you want users to be able to pass a list of template paths like `@, my-path, my-other-path`?

clab/config/template.go
hellt: I changed that in 16c6b58 to speed up the merging

clab/config/template.go
kellerza: Indeed yes, just a shortcut to add the built-in templates to the path list

kellerza:
Thanks!

PR #498, opened 2021-07-07
remove direct pointer deref in nodecfg creation

hellt:
Hi @karimra 
Do you have any objections to this fix of direct pointer dereference which occurs when the lab is defined like that:

```yaml
name: quickstart

topology:
  defaults:
    kind: srl
    image: nokia/srlinux

  nodes:
    leaf1:
    leaf2:
    spine1:

  links:
    - endpoints: ["leaf1:e1-49", "spine1:e1-1"]
    - endpoints: ["leaf2:e1-49", "spine1:e1-2"]

```


types/node_definition.go
```diff
@@ -65,10 +65,7 @@ func (n *NodeDefinition) GetStartupConfig() string {
...
 
 func (n *NodeDefinition) GetConfigDispatcher() *ConfigDispatcher {
-	if n == nil {
-		return nil
-	}
-	if n.Config == nil {
+	if n == nil || n.Config == nil {
 		return &ConfigDispatcher{}
 	}
 	return n.Config
```
karimra, line 71: why returning an empty struct here? that implies that a config is present. 
nil means it was not configured

karimra:
what was removed here ?

types/node_definition.go
hellt: yes, that is a dirty solution. I did this because we have the following code piece that merges maps of the `*ConfigDispatcher`
https://github.com/srl-labs/containerlab/blob/fix-nilderef/types/topology.go#L138-L151

in order to avoid nil pointer deref in case where I access Vars field like this `ndef.GetConfigDispatcher().Vars` I decided to return an initialized struct

I can rewrite that merge piece to first get the *ConfigDispatcher, and if it is not nil to try and access its Vars

hellt:
> what was removed here ?

direct access to fields of a struct with a pointer receiver. switched to a method based getter, that will check if the receiver is not nil

types/node_definition.go
karimra: maybe add a `GetConfigDispatcherVars` method ?

karimra:
> > what was removed here ?
> 
> direct access to fields of a struct with a pointer receiver. switched to a method based getter, that will check if the receiver is not nil

ok I was confused, this sounds more like an addition

types/node_definition.go
hellt: okay, that sounds fine
like that, I suppose d2f7473?

PR #502, opened 2021-07-08
cleanup unused variables and change Clab.WithRuntime(...) to use *runtime.RuntimeConfig

steiler:
the Clab bound options where not used any more. Hence I've deleted them.
changed Clab.WithRuntime from `(name string, d bool, dur time.Duration, gracefulShutdown bool)` to `(name string, rtconfig *runtime.RuntimeConfig)`.

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -30,9 +30,7 @@ type CLab struct {
 	globalRuntime string
 	Dir           *Directory
 
-	debug            bool
-	timeout          time.Duration
-	gracefulShutdown bool
+	timeout time.Duration
 }
 
 type Directory struct {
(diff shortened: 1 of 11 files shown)


clab/clab.go
```diff
@@ -73,10 +65,7 @@ func WithRuntime(name string, d bool, dur time.Duration, gracefulShutdown bool)
 		if rInit, ok := runtime.ContainerRuntimes[name]; ok {
 			r := rInit()
 			err := r.Init(
-				runtime.WithConfig(&runtime.RuntimeConfig{
-					Timeout: dur,
-					Debug:   d,
-				}),
+				runtime.WithConfig(rtconfig),
```
steiler, line 68: We can discuss if we should use a copy of the runtimeConfig provided to avoid any race conditions.
On the other hand one could use that handle to change runtime config by intention.
Unsure myself.

hellt:
now when I see the full PR I am not sure it is a good idea to import `runtime` package for `cmd` package

karimra:
1. I think the values are still used, they just moved to runtimeConfig
2. I think it's ok, as long as runtime and clab do not import cmd, it should be cycle free

steiler:
Also the debug variable, is just consumed in one case by the docker runtime.
https://github.com/srl-labs/containerlab/blob/fa9c2dae742526716c09de7e2f96a2c142fcdde8/runtime/docker/docker.go#L221-L226
And even this can be left out, since this is omitted via the debug level, which even used the global debug var.
To be honest that flag is not really needed atm.

hellt:
> Also the debug variable, is just consumed in one case by the docker runtime.

the debug is used not only in docker, it is also used in ignite

https://github.com/srl-labs/containerlab/blob/4caa86f05137210a07dce9f588689712683ef244/runtime/ignite/iginite.go#L239

hellt:
but the rest looks ok to me
Building the binary with `-race` doesn't detect any issues with passing runtime config as a pointer, so should be ok as well

PR #506, opened 2021-07-09
New Cumulus Lab

networkop:
Just adding a new lab-example for Cumulus


docs/lab-examples/lab-examples.md
```diff
@@ -56,6 +56,16 @@ You are ready to deploy!
 containerlab deploy -t <topo-file>
 ```
 
+#### SSH access
+For nodes that come up with `sshd` enabled, the following lines can be added to `~/.ssh/config` file to simplify access and prevent future ssh key warnings:
```
hellt, line 60: can we change `sshd` to just SSH? It feels the details are not that important here in this context.

Also can you please add the the ~/.ssh/config is on the containerlab host system? I bet somebody will try to change it in the context of a node...

hellt:
good to merge along with #507 ?

networkop:
yup :+1

hellt:
@networkop there is some issue with links - https://github.com/srl-labs/containerlab/runs/3036128969?check_suite_focus=true#step:4:23

Issue #493, opened 2021-07-06
decouple ignite nodes deployment from other runtimes


hellt:
As per https://github.com/srl-labs/containerlab/pull/482/files#r664336872 that PR uses a sequential deployment of nodes if it happens that any of the nodes requires ignite runtime

The proposal is to deploy ignite nodes sequentially, whereas the rest of the nodes will still benefit from the concurrent deployment

networkop:
I can work on that.

hellt:
done in #508

PR #508, opened 2021-07-09
Try to run as many concurrent workers as possible

networkop:
Addresses #493

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -168,60 +168,69 @@ func (c *CLab) GlobalRuntime() runtime.ContainerRuntime {
 	return c.Runtimes[c.globalRuntime]
 }
 
-func (c *CLab) CreateNodes(ctx context.Context, workerMap map[uint][]nodes.Node) {
-	wgs := make([]*sync.WaitGroup, 0, len(workerMap))
-	for maxWorkers, ns := range workerMap {
+func (c *CLab) CreateNodes(ctx context.Context, maxWorkers uint, serialNodes map[string]struct{}) {
 
-		// Making sure we don't overcommit
-		if maxWorkers > uint(len(ns)) {
-			maxWorkers = uint(len(ns))
-		}
-
-		wg := new(sync.WaitGroup)
-		wgs = append(wgs, wg)
+	wg := new(sync.WaitGroup)
+	if len(serialNodes) > 0 {
+		wg.Add(int(maxWorkers) + 1)
+	} else {
 		wg.Add(int(maxWorkers))
-		nodesChan := make(chan nodes.Node)
-		// start workers
-		for i := uint(0); i < maxWorkers; i++ {
-			go func(i uint, wg *sync.WaitGroup) {
-				defer wg.Done()
-				for {
-					select {
-					case node, ok := <-nodesChan:
-						if node == nil || !ok {
-							log.Debugf("Worker %d terminating...", i)
-							return
-						}
-						log.Debugf("Worker %d received node: %+v", i, node.Config())
-						// PreDeploy
-						err := node.PreDeploy(c.Config.Name, c.Dir.LabCA, c.Dir.LabCARoot)
-						if err != nil {
-							log.Errorf("failed pre-deploy phase for node %q: %v", node.Config().ShortName, err)
-							continue
-						}
-						// Deploy
-						err = node.Deploy(ctx)
-						if err != nil {
-							log.Errorf("failed deploy phase for node %q: %v", node.Config().ShortName, err)
-							continue
-						}
-					case <-ctx.Done():
-						return
-					}
-				}
-			}(i, wg)
-		}
-		for _, n := range ns {
-			nodesChan <- n
-		}
-		// close channel to terminate the workers
-		close(nodesChan)
 	}
 
-	for _, wg := range wgs {
-		// wait for all workers to finish
-		wg.Wait()
+	concurrentChan := make(chan nodes.Node)
+	serialChan := make(chan nodes.Node)
+
+	workerFunc := func(i uint, input chan nodes.Node, wg *sync.WaitGroup) {
+		defer wg.Done()
+		for {
+			select {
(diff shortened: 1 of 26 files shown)


networkop:
@hellt @karimra I've done a first pass of this feature for the deploy action. Let me know what you think before I do the rest.

clab/config.go
```diff
@@ -211,6 +211,7 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 		Sandbox:         c.Config.Topology.GetNodeSandbox(nodeName),
 		Kernel:          c.Config.Topology.GetNodeKernel(nodeName),
 		Runtime:         c.Config.Topology.GetNodeRuntime(nodeName),
+		MaxWorkers:      c.Config.Topology.GetNodeMaxWorkers(nodeName),
```
hellt, line 214: Do we want to make MaxWorkers to be a part of node config?
To me the number of max workers is a runtime's thing, not node's thing. A node is a single thing, it seems it doesn't add up having workers under a node.

If I think about the workers constraints, I'd think along the following lines:

1. build a `map[string][]nodes.Node` of nodes which has two keys
     1. `concurrent` -  this will have a list of nodes which are not artificially constrained (docker, containerd). The max workers will follow the current rule: no constraints, until set via flag
    2. `sequential` - this will have a list of nodes which are associated with runtimes which can't handle concurrent deployment (ignite)
1. then this map can be passed to `c.CreateNodes` and processed there

Maybe I am missing something here which makes this approach not feasible?

networkop: My thinking was to remove any node-specific logic from `clab`. The property of concurrency is node-dependent so it's natural for it to come out of node's config and not get decided in `clab`.
Another issue is that if `cvx` is running in docker runtime, it should have to be serialized, so we can override this default behaviour with `maxworker` field of the node.

hellt: but isn't it a runtime property still? as you mention, the cvx running in docker shouldn't be serialized, so it is the runtime that is a deciding factor here, not the node per-se. I.e. if we say for `cvx` node that the runtime is `docker`, then this is enough to make a decision to not serialize the deployment of the nodes.

another example, by exposing `maxworker` it is possible to do

```yaml
nodes:
  n1:
    kind: linux
    max-workers: 2
  n2:
    kind: linux
    max-workers: 4
```

but this doesn't make sense?

networkop: yeah, your example is ambiguous but can be deterministic, e.g. if we sort all nodes first and say that the first (or last) value always wins. 

So, if I understood you correctly, this is what you're suggesting:
```go
var nodeWorkers map[string][]nodes.Node
for _, n := range c.Nodes {
  if n.GetRuntime() == runtime.IgniteRuntime {
    nodeWorkers[1] = append(nodeWorkers[1], n)
    continue
  }
  nodeWorkers[0] = append(nodeWorkers[0], n)
}
```

steiler: I agree this is runtime dependent hence should not be any kind of Node property.

karimra: I think it doesn't have to be a node param.
a single `map[string]struct{}`, built after nodes are initialized, that contains the nodes names that need to be deployed sequentially is enough.

Deploy the nodes from the built map sequentially, then the ones that are not in the map using the defined number of workers.

networkop:
how does the latest commit look like? @hellt @steiler @karimra

clab/clab.go
```diff
@@ -168,45 +168,67 @@ func (c *CLab) GlobalRuntime() runtime.ContainerRuntime {
...
-					if err != nil {
-						log.Errorf("failed deploy phase for node %q: %v", node.Config().ShortName, err)
-						continue
-					}
-				case <-ctx.Done():
+	if len(serialNodes) > 0 {
+		wg.Add(int(maxWorkers) + 1)
+	} else {
+		wg.Add(int(maxWorkers))
+	}
```
karimra, line 178: I would write this as:
```golang
wg.Add(int(maxWorkers))
if len(serialNodes) > 0 {
  wg.Add(1)
}
```

But I just don't like `else`

lab-examples/cvx03/lab-start.yml
```diff
@@ -15,14 +15,14 @@ topology:
       image: networkop/cx:4.3.0
 
     server01:
-      kind: cvx
+      kind: linux
```
hellt, line 18: Hi @networkop 
what is the role of `kernel` key under the node config for a `linux` kind? is it because `runtime==ignite` is set via cmd flag there?

networkop: actually, it's a leftover, should be removed. it's only relevant to ignite runtime nodes

hellt: but then the image, is it a correct combination:

```yaml
server02:
      kind: linux
      image: networkop/host:ignite
```

hellt: or should the change be `kind: linux->cvx`?

networkop: yes, this image also works for both ignite and docker runtimes. the tag is just poor naming from me. I initially created it to run server nodes with ignite but then realised it works as well for non-ignite usecases.
You can see the buildfiles for it here https://github.com/networkop/cx/tree/main/host

networkop: so there's nothing ignite-specific in it

hellt: ok, will clean this up

clab/clab.go
hellt: did this

hellt:
@networkop I think it looks good

networkop:
> @networkop I think it looks good

cool, i'll add a similar code to the destroy action then.

networkop:
I think it all looks good now. wdyt @hellt ?

hellt:
thanks @networkop

PR #515, opened 2021-07-14
Merging maps

kellerza:
New MergeDicts function with tests. Merges multiple `vars` dicts (`... map[string]interface{}`)

Updated MergeStringMaps to be variadic and accept multiple maps

A small change without impact/needing rebase #495

### DIFF
--- a/utils/env.go
+++ b/utils/env.go
@@ -13,27 +13,50 @@ func ConvertEnvs(m map[string]string) []string {
 	return s
 }
 
-// mergeStringMaps merges map m1 into m2 and return a resulting map as a new map
-// maps that are passed for merging will not be changed
-func MergeStringMaps(m1, m2 map[string]string) map[string]string {
-	if m1 == nil {
-		return m2
-	}
-	if m2 == nil {
-		return m1
-	}
-	// make a copy of a map
-	m := make(map[string]string)
-	for k, v := range m1 {
-		m[k] = v
-	}
+// merge all dictionaries and return a new dictionary
+// recursively if matching keys are both dictionaries
+func MergeDicts(dicts ...map[string]interface{}) map[string]interface{} {
+	res := make(map[string]interface{})
+	for _, m := range dicts {
+		if m == nil {
+			continue
+		}
+		for k, v := range m {
 
-	for k, v := range m2 {
-		m[k] = v
+			if v0, ok := res[k]; ok {
+				// Recursive merging if res[k] exists (and both are dicts)
+				t0, ok0 := v0.(map[string]interface{})
+				t1, ok1 := v.(map[string]interface{})
+				if ok0 && ok1 {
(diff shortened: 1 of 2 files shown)


utils/env_test.go
```diff
@@ -0,0 +1,91 @@
...
+		"r":  "00",
+		"r2": "0",
+	}
+
+	expect := func(m1, m2 map[string]interface{}, v string) {
+		d := MergeDicts(m1, m2)
+		if fmt.Sprintf("%v", d) != v {
+			t.Errorf("err %v, expected %s", d, v)
+		}
+	}
```
karimra, line 59: The maps comparison should be done with `cmp.Equal` from `"github.com/google/go-cmp/cmp"`
Comparing the rendered strings does not guarantee types equality.

utils/env.go
```diff
@@ -13,27 +13,50 @@ func ConvertEnvs(m map[string]string) []string {
...
-	if m2 == nil {
-		return m1
-	}
-	// make a copy of a map
-	m := make(map[string]string)
-	for k, v := range m1 {
-		m[k] = v
+// merge all dictionaries and return a new dictionary
+// recursively if matching keys are both dictionaries
+func MergeDicts(dicts ...map[string]interface{}) map[string]interface{} {
```
karimra, line 18: I think it's better to call the function `MergeMaps`

utils/env_test.go
kellerza: Does it matter?, since MergeMaps cannot return anything except `map[string]interface{}` and all these values & results are manufactured test cases.

utils/env.go
```diff
@@ -13,27 +13,50 @@ func ConvertEnvs(m map[string]string) []string {
...
+	for _, m := range dicts {
+		if m == nil {
+			continue
+		}
+		for k, v := range m {
+
+			if v0, ok := res[k]; ok {
+				// Recursive merging if res[k] exists (and both are dicts)
+				t0, ok0 := v0.(map[string]interface{})
+				t1, ok1 := v.(map[string]interface{})
```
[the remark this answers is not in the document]
kellerza: Ok, I'll add this, but yaml & json both only have string keys, it is also mentioned the unmarshal golang docs - `map[string]interface{}`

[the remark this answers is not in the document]
karimra: yaml can return a `map[interface{}]interface}{}`

utils/env_test.go
karimra: This is Go's testing best practices, and it does matter, 
`map[string]interface{}{"key": int(1)}` is different from `map[string]interface{}{"key": uint(1)}` but their string representation is the same

kellerza:
should be good to merge

hellt:
Thanks @kellerza @karimra !

PR #517, opened 2021-07-16
yaml schema mgmt.bridge and node.runtime added

steiler:


hellt:
@steiler is it ready I suppose?

steiler:
Please take a look if you feel something more should be added for the runtimes or so, since this also includes the doc.
I haven't followed the ignite stuff. Other then that I assume this is OK so far.

Issue #236, opened 2021-01-21
/etc/hosts sometimes doesn't get cleaned up on reconfigure/destroy


hellt:
investigation is needed on how to reproduce it

steiler:
I would like to change the way we are doing this. 
I'd basically like to add a header and a footer comment to the blob we are adding containing the entries like
```
###### CLAB-<deploymentname>-START ######
<HOSTS ENTRIES>
###### CLAB-<deploymentname>-END ######
```
On deletion I'd like to search for these comments and delete them along with the content inbetween.

hellt:
I haven't seen this for a while now, will close this and reopen if this will surface again

PR #520, opened 2021-07-16
embedding host file entries in signature

steiler:
As outlined in #236 /etc/hosts entries are now being embedded in a per Lab (comment based) signature.
This is then used to delete entries on lab destroy.
```
###### CLAB-<LABNAME>-START ######
[<ENTRIES>]
###### CLAB-<LABNAME>-END ######
```

### DIFF
--- a/clab/general.go
+++ b/clab/general.go
@@ -3,11 +3,13 @@ package clab
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"strings"
 
+	"github.com/cloudflare/cfssl/log"
 	"github.com/srl-labs/containerlab/types"
 )
 
@@ -47,9 +49,10 @@ func AppendHostsFileEntries(containers []types.GenericContainer, labname string)
 
 // hostEntries builds an /etc/hosts compliant text blob (as []byte]) for containers ipv4/6 address<->name pairs
 func GenerateHostsEntries(containers []types.GenericContainer, labname string) []byte {
-	buff := bytes.Buffer{}
-	v4buff := bytes.Buffer{}
-	v6buff := bytes.Buffer{}
+	entries := strings.Builder{}
+	v6entries := strings.Builder{}
+
+	entries.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_PREFIX+"\n", labname))
 
 	for _, cont := range containers {
 		if len(cont.Names) == 0 {
@@ -57,25 +60,17 @@ func GenerateHostsEntries(containers []types.GenericContainer, labname string) [
 		}
 		if cont.NetworkSettings.Set {
 			if cont.NetworkSettings.IPv4addr != "" {
-				v4buff.WriteString(cont.NetworkSettings.IPv4addr)
-				v4buff.WriteString("\t")
-				v4buff.WriteString(strings.TrimLeft(cont.Names[0], "/"))
-				v4buff.WriteString("\n")
+				fmt.Fprintf(&entries, "%s\t%s\n", cont.NetworkSettings.IPv4addr, strings.TrimLeft(cont.Names[0], "/"))
 			}
 			if cont.NetworkSettings.IPv6addr != "" {
-				v6buff.WriteString(cont.NetworkSettings.IPv6addr)
-				v6buff.WriteString("\t")
-				v6buff.WriteString(strings.TrimLeft(cont.Names[0], "/"))
-				v6buff.WriteString("\n")
+				fmt.Fprintf(&v6entries, "%s\t%s\n", cont.NetworkSettings.IPv6addr, strings.TrimLeft(cont.Names[0], "/"))
 			}
 		}
 	}
-	// combine stuff
-	buff.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_PREFIX+"\n", labname))
-	buff.WriteString(v4buff.String())
-	buff.WriteString(v6buff.String())
-	buff.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_POSTFIX+"\n", labname))
-	return buff.Bytes()
+
+	entries.WriteString(v6entries.String())
+	entries.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_POSTFIX+"\n", labname))
+	return []byte(entries.String())
 }
 
 func DeleteEntriesFromHostsFile(labname string) error {
@@ -84,7 +79,12 @@ func DeleteEntriesFromHostsFile(labname string) error {
 	}
 	f, err := os.OpenFile("/etc/hosts", os.O_RDWR, 0644) // skipcq: GSC-G302
 	if err != nil {
-		return err
+		if errors.Is(err, os.ErrNotExist) {
+			log.Info("/etc/hosts does not exist")
+			return nil
+		} else {
+			return err
+		}
 	}
 	defer f.Close()
 	reader := bufio.NewReader(f)
--- a/clab/general.go
+++ b/clab/general.go
@@ -1,0 +1,127 @@
+package clab
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/srl-labs/containerlab/types"
+)
+
+var (
+	CLAB_HOSTENTRY_PREFIX  = "###### CLAB-%s-START ######"
+	CLAB_HOSTENTRY_POSTFIX = "###### CLAB-%s-END ######"
+)
+
+func AppendHostsFileEntries(containers []types.GenericContainer, labname string) error {
+	if labname == "" {
+		return fmt.Errorf("missing lab name")
+	}
+	// lets make sure we do not have remaining of a non destroyed run in the hosts file
+	err := DeleteEntriesFromHostsFile(labname)
+	if err != nil {
+		return err
+	}
+	data := GenerateHostsEntries(containers, labname)
+	if len(data) == 0 {
+		return nil
+	}
+	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_WRONLY, os.ModeAppend)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	_, err = f.WriteString("\n")
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(data)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// hostEntries builds an /etc/hosts compliant text blob (as []byte]) for containers ipv4/6 address<->name pairs
+func GenerateHostsEntries(containers []types.GenericContainer, labname string) []byte {
+	buff := bytes.Buffer{}
+	v4buff := bytes.Buffer{}
+	v6buff := bytes.Buffer{}
+
+	for _, cont := range containers {
+		if len(cont.Names) == 0 {
+			continue
+		}
+		if cont.NetworkSettings.Set {
+			if cont.NetworkSettings.IPv4addr != "" {
+				v4buff.WriteString(cont.NetworkSettings.IPv4addr)
+				v4buff.WriteString("\t")
+				v4buff.WriteString(strings.TrimLeft(cont.Names[0], "/"))
+				v4buff.WriteString("\n")
+			}
+			if cont.NetworkSettings.IPv6addr != "" {
+				v6buff.WriteString(cont.NetworkSettings.IPv6addr)
+				v6buff.WriteString("\t")
+				v6buff.WriteString(strings.TrimLeft(cont.Names[0], "/"))
+				v6buff.WriteString("\n")
+			}
+		}
+	}
+	// combine stuff
+	buff.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_PREFIX+"\n", labname))
+	buff.WriteString(v4buff.String())
+	buff.WriteString(v6buff.String())
+	buff.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_POSTFIX+"\n", labname))
+	return buff.Bytes()
+}
+
+func DeleteEntriesFromHostsFile(labname string) error {
+	if labname == "" {
+		return fmt.Errorf("missing containerlab name")
+	}
+	f, err := os.OpenFile("/etc/hosts", os.O_RDWR, 0644) // skipcq: GSC-G302
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	reader := bufio.NewReader(f)
+	skiplines := false
+	output := bytes.Buffer{}
+	prefix := fmt.Sprintf(CLAB_HOSTENTRY_PREFIX, labname)
+	postfix := fmt.Sprintf(CLAB_HOSTENTRY_POSTFIX, labname)
+	for {
+		line, err := reader.ReadString(byte('\n'))
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		if strings.TrimSpace(line) == postfix {
+			skiplines = false
+			continue
+		} else if strings.TrimSpace(line) == prefix || skiplines {
+			skiplines = true
+			continue
+		}
+		output.WriteString(line)
+	}
+	if skiplines {
+		// if skiplines is not false, we did not find the end
+		// so we should not mess with /etc/hosts
+		return fmt.Errorf("issue cleaning up /etc/hosts file. Please do so manually")
+	}
+	err = f.Truncate(0)
+	if err != nil {
+		return err
+	}
+	_, err = f.Seek(0, 0)
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(output.Bytes())
+	return err
+}
--- a/clab/general.go
+++ b/clab/general.go
@@ -14,8 +14,8 @@ import (
 )
 
 const (
-	ClabHostEntryPrefix  = "###### CLAB-%s-START ######"
-	ClabHostEntryPostfix = "###### CLAB-%s-END ######"
+	clabHostEntryPrefix  = "###### CLAB-%s-START ######"
+	clabHostEntryPostfix = "###### CLAB-%s-END ######"
 )
 
 func AppendHostsFileEntries(containers []types.GenericContainer, labname string) error {
@@ -53,7 +53,8 @@ func generateHostsEntries(containers []types.GenericContainer, labname string) [
 	entries := bytes.Buffer{}
 	v6entries := bytes.Buffer{}
 
-	fmt.Fprintf(&entries, ClabHostEntryPrefix+"\n", labname)
+	fmt.Fprintf(&entries, clabHostEntryPrefix, labname)
+	entries.WriteByte('\n')
 
 	for _, cont := range containers {
 		if len(cont.Names) == 0 {
@@ -70,7 +71,8 @@ func generateHostsEntries(containers []types.GenericContainer, labname string) [
 	}
 
 	entries.Write(v6entries.Bytes())
-	fmt.Fprintf(&entries, ClabHostEntryPostfix+"\n", labname)
+	fmt.Fprintf(&entries, clabHostEntryPostfix, labname)
+	entries.WriteByte('\n')
 	return entries.Bytes()
 }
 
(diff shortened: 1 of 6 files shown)


steiler:
Tests will follow.

clab/general.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, cont := range containers {
+		if len(cont.Names) == 0 {
+			continue
+		}
+		if cont.NetworkSettings.Set {
+			if cont.NetworkSettings.IPv4addr != "" {
+				v4buff.WriteString(cont.NetworkSettings.IPv4addr)
+				v4buff.WriteString("\t")
+				v4buff.WriteString(strings.TrimLeft(cont.Names[0], "/"))
+				v4buff.WriteString("\n")
```
kellerza, line 63: Maybe try a `strings.Builder`? it implements io.Writer, so you can do

```go
var buff strings.Builder

fmt.Fprintf(&buff, "%s\t%s\n", cont.NetworkSettings.IPv4addr, strings.Trim(cont.Names[0], "/"))
```

clab/general.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// hostEntries builds an /etc/hosts compliant text blob (as []byte]) for containers ipv4/6 address<->name pairs
+func GenerateHostsEntries(containers []types.GenericContainer, labname string) []byte {
+	buff := bytes.Buffer{}
+	v4buff := bytes.Buffer{}
```
kellerza, line 51: Do you need the `v4buff` if you write the host entry prefix and IPv4 to `buff` and only have one other intermediary `v6buff` to write after the loop?

clab/general.go
```diff
@@ -0,0 +1,127 @@
...
+	buff.WriteString(v6buff.String())
+	buff.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_POSTFIX+"\n", labname))
+	return buff.Bytes()
+}
+
+func DeleteEntriesFromHostsFile(labname string) error {
+	if labname == "" {
+		return fmt.Errorf("missing containerlab name")
+	}
+	f, err := os.OpenFile("/etc/hosts", os.O_RDWR, 0644) // skipcq: GSC-G302
```
kellerza, line 85: You either need to create it (`O_CREATE`) or check the resulting error for `errors.Is(err, os.ErrNotExist)`

clab/general.go
```diff
@@ -0,0 +1,127 @@
...
+	CLAB_HOSTENTRY_POSTFIX = "###### CLAB-%s-END ######"
+)
+
+func AppendHostsFileEntries(containers []types.GenericContainer, labname string) error {
+	if labname == "" {
+		return fmt.Errorf("missing lab name")
+	}
+	// lets make sure we do not have remaining of a non destroyed run in the hosts file
+	err := DeleteEntriesFromHostsFile(labname)
+	if err != nil {
```
kellerza, line 25: If the file does not exist, you should continue here...

hellt:
thanks for your review @kellerza!

hellt:
@karimra 
what do you think about this change? I personally find the general idea quite nice to have a visual separation between the entries and the labs for which they were created

clab/general.go
steiler: thought it might be nice to collect everything up front and finally compose it, just for readability.
However compressed it now.

clab/general.go
steiler: Alright, logging it now and returning with nil.

clab/general.go
steiler: see comment above, so we will now continue here now

clab/general.go
```diff
@@ -0,0 +1,126 @@
...
+			if cont.NetworkSettings.IPv6addr != "" {
+				fmt.Fprintf(&v6entries, "%s\t%s\n", cont.NetworkSettings.IPv6addr, strings.TrimLeft(cont.Names[0], "/"))
+			}
+		}
+	}
+
+	entries.WriteString(v6entries.String())
+	entries.WriteString(fmt.Sprintf(CLAB_HOSTENTRY_POSTFIX+"\n", labname))
+	return []byte(entries.String())
+}
```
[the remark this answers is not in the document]
kellerza: Overall I'd lean toward `*printf`, but might be a personal preference.

See this can likely also be done with `bytes.Buffer` 

The grouping is good, especially since the target here is human readability with the comment block

clab/general.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	_, err = f.Write(data)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// hostEntries builds an /etc/hosts compliant text blob (as []byte]) for containers ipv4/6 address<->name pairs
+func GenerateHostsEntries(containers []types.GenericContainer, labname string) []byte {
```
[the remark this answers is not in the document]
steiler: yes, changed it!

clab/general.go
```diff
@@ -0,0 +1,126 @@
...
+	"strings"
+
+	"github.com/cloudflare/cfssl/log"
+	"github.com/srl-labs/containerlab/types"
+)
+
+var (
+	CLAB_HOSTENTRY_PREFIX  = "###### CLAB-%s-START ######"
+	CLAB_HOSTENTRY_POSTFIX = "###### CLAB-%s-END ######"
+)
```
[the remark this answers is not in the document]
steiler: changed

clab/general.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func DeleteEntriesFromHostsFile(labname string) error {
+	if labname == "" {
+		return fmt.Errorf("missing containerlab name")
+	}
+	f, err := os.OpenFile("/etc/hosts", os.O_RDWR, 0644) // skipcq: GSC-G302
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			log.Info("/etc/hosts does not exist")
```
kellerza, line 83: ?
```suggestion
			log.Info("/etc/hosts does not exist")
			return nil
```

steiler: @hellt how do you want it to be?

clab/general.go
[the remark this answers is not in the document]
karimra: - Iterating twice vs using 2 io.Writer is arguable, but fair enough let's use 2 writers.
- The reason the method returns a []byte is because the output will be written to a file as []byte. 
`strings.Builder` uses a []byte under the hood, it was added to Go as a better alternative to `fmt.Sprintf` if the user knows the items to be written are strings and has the intention of using the result as a `string` (that's why strings.Builder doesn't have/need a `.Bytes()` method but `bytes.Buffer` does).
The use of `strings.Builder` in this case, writes strings to bytes (in the builder) then converts them to a string to then convert the whole thing to []byte to be returned.
If we return a string, the function writing to the file will have to covert to []byte anyways.
- Using string formatters `%s` implies that `fmt.Sprintf` has to do a type reflection on the received `interface{}` to determine its type and find its string representation. which defeats the purpose of buffered writers (`strings.Builder` and/or `bytes.Buffer`)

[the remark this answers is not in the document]
steiler: OK, I think that is all now as discussed separately with you @karimra
I'll set this to resolved

clab/general.go
hellt: if /etc/hosts doesn't exist that looks like an error to me as far as this function is concerned

@kellerza why do you want to return `nil` in that case and not `errors.New("/etc/hosts does not exist")`?

hellt:
what happens if a user has a running lab, they upgrade to a release with this change, then they try to destroy a running lab?

I support the code won't handle the deletion of host entries entered in the previous format?

steiler:
My /etc/hosts was just totally broken ... single characters per line and what not ... might be caused by broken debugging runs or the like... not sure what your hosts file looks like ... maybe a release notes note to double-check and maybe manually clean up?

clab/general.go
kellerza: If you return an error then the calling function (AppenHostEntries) will fail to append.

So we either return a nil here, or we change AppenHostentries to still write if the file does not exist

steiler:
Had to remove the newline since append mode already is on a new line.
otherwise due to the way we delete entries we would add a newline every run which would remain.
https://github.com/srl-labs/containerlab/pull/520/commits/c426be3f2c33a05f14949db71c9b84d6426c5cf7#diff-af304e6c2934686344362db1ac3ffef043cff5e55c3ecf60ba28de37d5f80ce8L39-L42

clab/general.go
hellt: if there is no hosts file, where the `AppendHostEntries` will append to?

kellerza: I would assume it should create the hosts file?

steiler: It does not so far. Is creating the file what we want to happen?

clab/general.go
```diff
@@ -0,0 +1,126 @@
...
+	// lets make sure we do not have remaining of a non destroyed run in the hosts file
+	err := DeleteEntriesFromHostsFile(labname)
+	if err != nil {
+		return err
+	}
+	data := generateHostsEntries(containers, labname)
+	if len(data) == 0 {
+		return nil
+	}
+	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_WRONLY, os.ModeAppend)
```
steiler, line 34: Create if not existing would look like this.

```suggestion
	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_WRONLY, os.ModeAppend)
	
		if err != nil {
		if errors.Is(err, os.ErrNotExist) {
			f, err = os.Create("/etc/hosts")
			if err != nil {
				return err
			}
			err = f.Chmod(0644)
			if err != nil {
				return err
			}
			err = f.Chown(0, 0)
			if err != nil {
				return err
			}
		} else {
			return err
		}
	}
```

kellerza: If AppendHostEntries appends or creates the file., you can just return here and let AppenHostEntries do it's job. No need to create it here. Just retrun `nil` for AppendHostEntries

steiler: a basic "|os.O_CREATE" created the file with a permission set of 000. So the perms should be changed. I only get to know when the perms should be changed if I get the error. Hence I need to separately create the file and set the perms.

tests/01-smoke/01-basic-flow.robot
```diff
@@ -107,12 +107,28 @@ Verify l1 environment has MYVAR variable set
     Should Be Equal As Integers    ${rc}    0
     Should Contain    ${output}    MYVAR is SET
 
+Verify Hosts entries exist
+    [Documentation]     Verification that the expected /etc/hosts entries are created. We are also checking for the HEADER and FOOTER values here, which also contain the lab name.
+    ${rc}    ${output} =    Run And Return Rc And Output
+    ...    cat /etc/hosts | grep "2-linux-nodes" | wc -l
```
hellt, line 113: should user test suite variable instead of string `2-linux-nodes`

clab/general.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/cloudflare/cfssl/log"
```
hellt, line 12: wrong log used here?

clab/general.go
hellt: Let's do the following

1. in AppendHosts we check if the `/etc/hosts` is there. We can use already available func https://github.com/srl-labs/containerlab/blob/7e8ea72915b6566ede6cd93f737cefdcd239c5d8/utils/file.go#L14
2. then if it is not there, we create a file (again with the existing func https://github.com/srl-labs/containerlab/blob/7e8ea72915b6566ede6cd93f737cefdcd239c5d8/utils/file.go#L78), with content `127.0.0.1 localhost`
3. with that, we do not mess too much with permissions and reuse the existing code

thoughts, @kellerza @steiler

steiler: That looks alright to me.

```
root@centos-srlinux:/home/ubuntu/containerlab# go build ; rm /etc/hosts; ls -la /etc/hosts ; ./containerlab  deploy -t tests/01-smoke/01-linux-nodes.clab.yml ; cat /etc/hosts ; ./containerlab  destroy -t tests/01-smoke/01-linux-nodes.clab.yml  ; ls -la /etc/hosts ;cat /etc/hosts
ls: cannot access '/etc/hosts': No such file or directory
INFO[0000] Parsing & checking topology file: 01-linux-nodes.clab.yml
INFO[0000] Creating lab directory: /home/ubuntu/containerlab/clab-2-linux-nodes
INFO[0000] Creating container: l1
INFO[0000] Creating container: l3
INFO[0000] Creating container: l2
INFO[0001] Creating virtual wire: l1:eth1 <--> l2:eth1
INFO[0001] Creating virtual wire: l1:eth2 <--> l2:eth2
INFO[0001] Writing /etc/hosts file
INFO[0001] � New containerlab version 0.15.3 is available! Release notes: https://containerlab.srlinux.dev/rn/0.15/#0.15.3
Run 'containerlab version upgrade' to upgrade or go check other installation options at https://containerlab.srlinux.dev/install/
+---+-----------------------+--------------+---------------------+-------+-------+---------+------------------+------------------------+
| # |         Name          | Container ID |        Image        | Kind  | Group |  State  |   IPv4 Address   |      IPv6 Address      |
+---+-----------------------+--------------+---------------------+-------+-------+---------+------------------+------------------------+
| 1 | clab-2-linux-nodes-l1 | 532fb61a9a53 | alpine:3            | linux |       | running | 172.20.20.2/24   | 2001:172:20:20::2/64   |
| 2 | clab-2-linux-nodes-l2 | 574f0cf19f2a | nginx:stable-alpine | linux |       | running | 172.20.20.100/24 | 2001:172:20:20::100/64 |
| 3 | clab-2-linux-nodes-l3 | 19cbcea3faa7 | nginx:stable-alpine | linux |       | running |                  |                        |
+---+-----------------------+--------------+---------------------+-------+-------+---------+------------------+------------------------+
127.0.0.1       localhost
###### CLAB-2-linux-nodes-START ######
172.20.20.100   clab-2-linux-nodes-l2
172.20.20.2     clab-2-linux-nodes-l1
2001:172:20:20::100     clab-2-linux-nodes-l2
2001:172:20:20::2       clab-2-linux-nodes-l1
###### CLAB-2-linux-nodes-END ######
INFO[0000] Parsing & checking topology file: 01-linux-nodes.clab.yml
INFO[0000] Destroying lab: 2-linux-nodes
INFO[0000] Removed container: clab-2-linux-nodes-l3
INFO[0000] Removed container: clab-2-linux-nodes-l1
INFO[0000] Removed container: clab-2-linux-nodes-l2
INFO[0000] Removing container entries from /etc/hosts file
-rw-r--r-- 1 root root 20 Jul 20 14:36 /etc/hosts
127.0.0.1       localhost
```

steiler:
@hellt, final review on this one?

clab/general.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/srl-labs/containerlab/utils"
+)
+
+const (
+	clabHostEntryPrefix  = "###### CLAB-%s-START ######"
+	clabHostEntryPostfix = "###### CLAB-%s-END ######"
+)
+
+func AppendHostsFileEntries(containers []types.GenericContainer, labname string) error {
+	filename := "/etc/hosts"
```
hellt, line 21: lets make this a constant `clabHostsFilename`?

clab/general.go
```diff
@@ -0,0 +1,130 @@
+package clab
```
hellt, line 1: let's rename this file to `clab/hosts.go` maybe? `general.go` doesn't say anything about the purpose of it

clab/general.go
```diff
@@ -0,0 +1,130 @@
...
+	if labname == "" {
+		return fmt.Errorf("missing lab name")
+	}
+	if !utils.FileExists(filename) {
+		err := utils.CreateFile(filename, "127.0.0.1\tlocalhost")
+		if err != nil {
+			return err
+		}
+	}
+	// lets make sure we do not have remaining of a non destroyed run in the hosts file
```
hellt, line 31: comment: 

```
// lets make sure to remove the entries of a non-properly destroyed lab in the hosts file
```

clab/general.go
```diff
@@ -0,0 +1,130 @@
...
+
+	defer f.Close()
+	_, err = f.Write(data)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// hostEntries builds an /etc/hosts compliant text blob (as []byte]) for containers ipv4/6 address<->name pairs
```
hellt, line 55: comment should start with `generateHostsEntries`

clab/general.go
```diff
@@ -0,0 +1,130 @@
...
+
+	entries.Write(v6entries.Bytes())
+	fmt.Fprintf(&entries, clabHostEntryPostfix, labname)
+	entries.WriteByte('\n')
+	return entries.Bytes()
+}
+
+func DeleteEntriesFromHostsFile(labname string) error {
+	if labname == "" {
+		return fmt.Errorf("missing containerlab name")
```
hellt, line 86: use `errors.New` instead, if no formatting is used

clab/general.go
```diff
@@ -0,0 +1,130 @@
...
+		} else if strings.TrimSpace(line) == prefix || skiplines {
+			skiplines = true
+			continue
+		}
+		output.WriteString(line)
+	}
+	if skiplines {
+		// if skiplines is not false, we did not find the end
+		// so we should not mess with /etc/hosts
+		return fmt.Errorf("issue cleaning up /etc/hosts file. Please do so manually")
```
hellt, line 118:
```
errors.New("failed to cleanup /etc/hosts file: no clab markers found. Please do so manually")
```

clab/general.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+		if cont.NetworkSettings.Set {
+			if cont.NetworkSettings.IPv4addr != "" {
+				fmt.Fprintf(&entries, "%s\t%s\n", cont.NetworkSettings.IPv4addr, strings.TrimLeft(cont.Names[0], "/"))
+			}
+			if cont.NetworkSettings.IPv6addr != "" {
+				fmt.Fprintf(&v6entries, "%s\t%s\n", cont.NetworkSettings.IPv6addr, strings.TrimLeft(cont.Names[0], "/"))
+			}
+		}
+	}
```
hellt, line 76: maybe I am missing something here, but why don't use the same `entries` and write IPv6addr to it in the same loop?

hellt (changes requested):

clab/general.go
steiler: If I'm not mistaking also you liked the idea to have IP versions separated in the block.
It helps a lot with readability should someone take a look at the hosts file for reference.
In the end this was my personal liking to do it that way.

hellt: ah, I see. you group ipv4 and ipv6 separately inside the same lab hosts block

I didn't get this intention from looking at the code.
Do you happen to have the example of the hosts content with such grouping (I have no lab setup with me at the moment)

steiler: Sorry wasn't you it was @kellerza -> https://github.com/srl-labs/containerlab/pull/520#discussion_r672846069

kellerza: Indeed, I like separating them. 

Maybe it's OCD, but the length difference between IPv4&6 messes up the alignment if you don't group them

```
172.20.20.2     clab-tag2-sr2
2001:172:20:20::2       clab-tag2-sr2
172.20.20.3     clab-tag2-sr1
2001:172:20:20::3       clab-tag2-sr1
```

cmd/destroy.go
```diff
@@ -164,7 +164,7 @@ func destroyLab(ctx context.Context, c *clab.CLab) (err error) {
 		}
 	}
 
-	log.Info("Removing container entries from /etc/hosts file")
+	log.Info("Removing containerlab host entries from /etc/hosts file")
```
steiler, line 167: this is prone to error. 
you might want to reference the const with the filename... maybe the log messages should even go into the function?!

hellt: Yes, move plz into the clab hosts

hellt: I decided not to do that
it involves more changes that I wanted, the deletion function is called during the creation as well

PR #525, opened 2021-07-18
Free form vars in topo files

kellerza:
Last step of free form variables, as discussed in #487 & using merge from #515

~`Vars map[string]string`~ --> `Vars map[string]interface{}`

String based variables required comma separated values for both ends of the link, but with free form variables these can now be an array (length=2) of string for the simplest case. More advances cases can contain an array (length=2) of any variable type


hellt:
is it possible to merge maps of vars which have lists inside?

i.e.

```yaml
config:
  vars:
    myfields:
      - var1: val1
      - var2: val2
```

clab/config/utils.go
```diff
@@ -101,35 +102,37 @@ func prepareLinkVars(lIdx int, link *types.Link, varsA, varsB Dict) error {
...
 			return fmt.Errorf("%s: reserved variable name '%s' found", link.String(), k)
 		}
 
-		if k == vkLinkIP && len(r) == 1 {
-			// calc the remote IP
-			ipF, err := ipFarEndS(v)
-			if err != nil {
-				return fmt.Errorf("%s: %s", link.String(), err)
+		vv := reflect.ValueOf(v)
+		if vv.Kind() == reflect.Slice || vv.Kind() == reflect.Array {
```
hellt, line 112: can it possible be `Kind()==Array` here?

kellerza:
> is it possible to merge maps of vars which have lists inside?

The only special case for `MergeMaps` is if we have maps `map[str/int]interface{}`. All other variable types will be overridden by the second param

What do you expect to happen if lists are merged, simply concatenate them? Remove duplicates?

clab/config/utils.go
kellerza: Since the source of the data if from yaml, it could be safe just to use Array. If it was generated by anything else it it could be array-like / or a slice I guess.

Will try it

kellerza: Added a test case for it and Slice seems more preferred...

hellt:
@kellerza I noticed that if there is an error with templation, you're trying to dump the variables for the node, but it doesn't work

```
INFO[0000] srl2 vars =                                  
Error: template: 01-interfaces__srl.tmpl:15:14: executing "01-interfaces__srl.tmpl" at <.mylinks>: can't evaluate field mylinks in type interface {}
```

the problem is that you're ignoring the error during the marshalling https://github.com/kellerza/containerlab/blob/vars_interface/clab/config/template.go#L119
and the marshalling doesn't work, since some of the map keys are now of `interface{}` type

```
json: unsupported type: map[interface {}]interface {}
```

I suggest you change the marshalling from json to yaml, then there is no problem with marshalling, as well as it is easier to read the variables

hellt:
if I have configs defined like this

```yaml
name: srlcfg

topology:
  defaults:
    config:
      vars:
        globmap:
          var1: val1
          var2: val2
    srl2:
      kind: srl
      image: ghcr.io/nokia/srlinux
      config:
        vars:
          globmap:
            var2: rewritten
            newvar: newval
```

it fails to render with a bad error descr `cannot prepare link vars for 0. link [srl1:e1-1, srl2:e1-1]: link [srl1:e1-1, srl2:e1-1]: systemip var required on all nodes`

the expectation is the resulting `globmap` variable should be:

```yaml
globmap:
  var1: val1
  var2: rewritten
  newvar: newval
```

kellerza:
@hellt can you mail me your full yml files referred to above?

I'm guessing you have a map in a list that causes the json unmarshal error. (It should be the only case that does not go through Karim's mapify function) Although your globmap here seems to be a pure map.

> cannot prepare link vars for 0. link [srl1:e1-1, srl2:e1-1]: link [srl1:e1-1, srl2:e1-1]: systemip var required on all nodes

This should only happen if you have `systemip` var defined on one node / one end of a link. Once you have it defined at all the special logic for link IPs kicks in (optional before that). Let me change that to be a warning only

hellt:
@kellerza yes, the `systemip` slipped in one node only, I removed it, and the error is gone.
But, the map merging doesn't look good (at least when marshalled to yaml). Check the full clab file I used

```yaml
name: srlcfg

topology:
  defaults:
    config:
      vars:
        globvar: globval
        globmap:
          var1: val1
          var2: val2
  nodes:
    srl1:
      kind: srl
      image: ghcr.io/nokia/srlinux
    srl2:
      kind: srl
      image: ghcr.io/nokia/srlinux
      config:
        vars:
          # var1: val1
          globmap:
            var2: rewritten
            newvar: newval
          interfaces:
            - name: ethernet-1/1
              description: set in node
            - name: ethernet-1/2

  links:
    - endpoints: ["srl1:e1-1", "srl2:e1-1"]
```

And here is what I get for `vars` for node `srl2` (check how `globmap` at the very beginning of a yaml has unparsed(?) key and missing elements:

```yaml
globmap:
  <interface {} Value>: rewritten
globvar: globval
interfaces:
- description: set in node
  name: ethernet-1/1
- name: ethernet-1/2
links:
- far:
    name: to_srl2
    node: srl1
  name: to_srl1
node: srl2
nodes:
  srl1:
    globmap:
      var1: val1
      var2: val2
    globvar: globval
    links:
    - far:
        name: to_srl1
        node: srl2
      name: to_srl2
    node: srl1
    role: srl
  srl2:
    globmap:
      <interface {} Value>: rewritten
    globvar: globval
    interfaces:
    - description: set in node
      name: ethernet-1/1
    - name: ethernet-1/2
    links:
    - far:
        name: to_srl2
        node: srl1
      name: to_srl1
    node: srl2
    role: srl
role: srl 
```

kellerza:
Thx @hellt, really strange. Will have a look at it later today and add this as a test case for MergeMaps. Likely be a new PR to fix MergeMaps and the logic there. Btw, have you ever looked at adding codecov to see test coverage?

Please let me know if you agree with the concept of #531 .. having all nodes in the debug dumb seems even worse now in yal format

kellerza:
#533 includes the MergeMaps improvements & new tests (specifically the input you gave here).

This should PR be ok to Merge (then I can rebase #531 and work on the check flag there)

hellt:
@kellerza #533 merged

kellerza:
Ok, that was not so good. This is why I hate rebasing

hellt:
@kellerza you can probably checkout to this commit https://github.com/srl-labs/containerlab/pull/525/commits/4e143a0efde87ca89f75f178848d00b27bc8b6fa
then create a new branch out of it and merge master into it

then create a new PR

PR #531, opened 2021-07-20
Supress nodes in config debug output

kellerza:
Debug output of variables is currently very long due to the fact that all nodes variables are included under 'nodes'

This PR compresses other nodes:

The output becomes:

```
INFO[0000] p1 vars = {
      "links": [
            {
                  "far": {
                        "ip": "1.3.8.0/31",
                        "isis_iid": "0",
                        "name": "to_p1",
                        "node": "asbr1",
                        "port": "1/1/c2/1",
                        "sid_idx": "3"
                  },
                  "ip": "1.3.8.1/31",
                  "isis_iid": "0",
                  "name": "to_asbr1",
                  "port": "1/1/c1/1",
                  "sid_idx": "8"
            },
            {
                  "far": {
                        "ip": "1.2.8.0/31",
                        "isis_iid": "0",
                        "name": "to_p1",
                        "node": "abr1",
                        "port": "1/1/c1/1",
                        "sid_idx": "2"
                  },
                  "ip": "1.2.8.1/31",
                  "isis_iid": "0",
                  "name": "to_abr1",
                  "port": "1/1/c2/1",
                  "sid_idx": "8"
            }
      ],
      "node": "p1",
      "nodes": "{p1: {...}, pe1: {...}, pe2: {...}, abr1: {...}, asbr1: {...}, asbr2: {...}, }",
      "r20": "1",
      "role": "vr-sros",
      "systemip": "8.8.8.8/32"
  }
```

The important part here is 
```
      "nodes": "{p1: {...}, pe1: {...}, pe2: {...}, abr1: {...}, asbr1: {...}, asbr2: {...}, }",
```

which makes the total output about 1/7th of what it is today.

Note that with the debug flags `-d -c 1` you do see the other node individually as well

It is also possible to see the full output using a debug count of 3 or higher `-d 3`

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -25,6 +25,7 @@ var configCmd = &cobra.Command{
 		var err error
 
 		transport.DebugCount = debugCount
+		config.DebugCount = debugCount
 
 		c, err := clab.NewContainerLab(
 			clab.WithTimeout(timeout),
@@ -101,7 +102,7 @@ var configCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(configCmd)
 	configCmd.Flags().StringSliceVarP(&config.TemplatePaths, "template-path", "p", []string{}, "comma separated list of paths to search for templates")
-	configCmd.MarkFlagDirname("template-path")
+	_ = configCmd.MarkFlagDirname("template-path")
 	configCmd.Flags().StringSliceVarP(&config.TemplateNames, "template-list", "l", []string{}, "comma separated list of template names to render")
 	configCmd.Flags().IntVarP(&printLines, "check", "c", 0, "render templates in dry-run mode & print N lines of rendered config")
 }
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -51,7 +51,7 @@ var configCmd = &cobra.Command{
 					c.Print(true, true)
 					return nil
 				}
-				log.Warnf("Invalid command line option for check. It needs to be either 'template'(default), 'vars', 'all' or a valid node name")
+				log.Warnf("Invalid command line option for check. Options: 'template'(default), 'vars', 'all' or a valid node name")
 				pt = true
 			}
 			for _, c := range allConfig {
@@ -116,6 +116,6 @@ func init() {
 	configCmd.Flags().StringSliceVarP(&config.TemplatePaths, "template-path", "p", []string{}, "comma separated list of paths to search for templates")
 	_ = configCmd.MarkFlagDirname("template-path")
 	configCmd.Flags().StringSliceVarP(&config.TemplateNames, "template-list", "l", []string{}, "comma separated list of template names to render")
-	configCmd.Flags().StringVarP(&check, "check", "c", "", "render templates in dry-run mode & print N lines of rendered config")
+	configCmd.Flags().StringVarP(&check, "check", "c", "", "render templates in dry-run mode & print either 'template', 'vars', 'all' or a specific node")
 	configCmd.Flags().Lookup("check").NoOptDefVal = "template"
 }
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -11,8 +11,8 @@ import (
 	"github.com/srl-labs/containerlab/nodes"
 )
 
-// Only print config locally, don't send to the node
-var printLines int
+// Dryrun and print config
+var check string
 
 // configCmd represents the config command
 var configCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


hellt:
yes, collapsing the `nodes` of the `vars` is a good choice

I also have the following comment about the `--check` flag. I do not think it makes sense to require a user to specify the number of lines of rendered config they want to see. I think by default we should display the full rendered config when `--check` is supplied.
It would be awesome to allow to dump the rendered configs as files as well as just printing them on the console.

Like I always want to see the full rendered config to make sure that my templates work as expected, and so far I need to artificially crank up the number of displayed lines and doesn't look right

kellerza:
I suggest we keep check as a count, but by default print all lines, so:
- `-c` would be all lines (your use-case & the simplest option)
- `-c 1` would be 1 line of config

So keeping both of our use-cases. I use it a lot to just show the variables that I will have available, like this

`clab config -t topo.clab.yml -c1 -d`

hellt:
> I use it a lot to just show the variables that I will have available

that looks like using abusing the flag  =)

kellerza:
I'm open for suggestions. What about

* -c /-c1 show rendered template
* -cc /-c2 show the vars
* -ccc, where c>=3 show all

hellt:
what if `--check | -c` becomes a stringvar that can take

* -c -- show rendered template (default string value is "template")
* -c vars -- show vars
* -c all -- show vars and full template

cmd/config.go
```diff
@@ -101,7 +114,8 @@ var configCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(configCmd)
 	configCmd.Flags().StringSliceVarP(&config.TemplatePaths, "template-path", "p", []string{}, "comma separated list of paths to search for templates")
-	configCmd.MarkFlagDirname("template-path")
+	_ = configCmd.MarkFlagDirname("template-path")
 	configCmd.Flags().StringSliceVarP(&config.TemplateNames, "template-list", "l", []string{}, "comma separated list of template names to render")
-	configCmd.Flags().IntVarP(&printLines, "check", "c", 0, "render templates in dry-run mode & print N lines of rendered config")
+	configCmd.Flags().StringVarP(&check, "check", "c", "", "render templates in dry-run mode & print either 'template', 'vars', 'all' or a specific node")
```
kellerza, line 119: New check var as discussed

kellerza:
Hi @hellt , I'm having some reservations about the filter I added here and think this should be a dedicated flag

To debug we could rather have an action flag, options here are: vars, template/dry-run, compare, commit. and possibly a show/send (implies no transaction - this is based of the template name today, which is probably not ideal) 

Taking this into account, I believe we need to split the filter and the action, then we end up with something like this:

```
Flags:
  -p, --template-path strings   comma separated list of paths to search for templates
  -a, --action string           specify the configuration action: (default "commit")
                                - vars: only show variables that will be used by the templates
                                - template: show rendered templates
                                - commit: render the template and commit to the node
                                - compare: render the template and compare with the current configuration
                                - send: send the command without any transaction (use for show commands)
  -f, --filter strings          comma separated list of nodes to include
```

What do you think?

hellt:
Maybe better to have subcommands for that? i.e.

* `clab config send`
* `clab config template`
* `clab config compare`

this will be more idiomatic and in line with the way we use Cobra to structure other CLI commands

the flags `-p`, `-l`, `-f` are global to `config` command, so they will propagate to subcommands

I wouldn't make a distinction between `commit` and `send` actions, because in a generic ssh transport (that I plan to add eventually and be based on scrapligo) the commit command will be part of the template to let user decide which commit command they need (commit [confirmed, timeout, now, stay, etc])

kellerza:
Once you move the transaction awareness from clab into the template, I believe you will loose the flexibility to have a template, start playing with compare and when you are happy commit.

You will see that all transactional awareness start with /configure global and discard to ensure you are starting from scratch,- do you want to move all this responsibility to the template user?

hellt:
yes, there soon will be a scrapli-cfg available in Go that will does the smart diff/compare. I am planning to use that, as this will deliver multivendor support.

But now let me check with @carlmontanari if `scrapligo-cfg` intends to make a decision which commit command to use...
@carlmontanari in scrapli-cfg, does a user provide the config snippet with `commit bla` command, or scrapli-cfg appends this?

kellerza:
There is no way for a computer to know your intention, compare/commit

hellt:
the intention is given with a command `clab config compare` vs `clab config send/commit`

at this moment (while scrapligo doesn't have `cfg` subsystem in place) I am more than happy with your approach (just using subcommands instead of a flag)

kellerza:
My understanding of the -cfg submodule is that you will have to pass a parameter to scapligo-cfg's commit/compare/send (send being the default or rather scapligo as it is today, without cfg)

Will look into the subcommands, so we will end up with 5 of these subcommands? vars, template, compare/commit/send? Won't this mean a bit of code duplication?

hellt:
I am thinking about having the following commands

* `config` - this does the `commit` operation
* `config` with `--send` will send the commands in raw mode without entering into any config mode or committin on behalf of a user?
* `config template` - this will render the templates without sending them anywhere. potentially this command will be able to have local flags to save rendered configs under per-node directory for example.
* if `config template` is provided with local `--vars` flag, it will show only the vars used for templation
* config `compare` - does the compare

so in total three commands maybe?

kellerza:
What about 2 only:
1. config (None / compare / send)
2. config template

In this case the `--send` flag wont be carried forward unnecessarily to `config template` and all three variants of 1 has to do with actually transmitting to the node.

2 would require new parameters for saving to files etc

PR #534, opened 2021-07-21
config variables as `interface{}`

kellerza:
Replacement for #525

Thx for the suggestion @hellt 

```bash
git checkout -b vars_interface2 4e143a0
git pull --rebase upstream master
git push origin HEAD
```


clab/config/utils.go
```diff
@@ -13,16 +14,16 @@ import (
...
-	vkLinkIP   = "ip"       // optional, link IP
-	vkLinkName = "name"     // optional, from ShortNames
-	vkLinkNr   = "linknr"   // optional, link number in case you have multiple, used to calculate the name
+	vkNodeName = "node"       // reserved, used for the node's ShortName
+	vkNodes    = "clab_nodes" // reserved, used for all nodes
+	vkLinks    = "clab_links" // reserved, used for all link in a node
+	vkFarEnd   = "clab_far"   // reserved, used for far-end link & node info
+	vkRole     = "clab_role"  // optional, will default to the node's Kind. Used to select the template
+
+	vkSystemIP = "system_ip" // optional, system IP if present could be used to calc link IPs
```
hellt, line 23: why not `clab_system_ip` to indicate this will potentially be used by clab to do stuff?

clab/config/utils.go
```diff
@@ -13,16 +14,16 @@ import (
...
+	vkLinkIP   = "ip"        // optional, link IP
+	vkLinkName = "name"      // optional, from ShortNames
+	vkLinkNo   = "link_no"   // optional, link number in case you have multiple, used to calculate the name
```
hellt, line 26: same thing, let's add `clab_` prefix to indicate the significance of those vars?

and maybe then

* `clab_link_ip`
* `clab_link_name`
* `clab_link_num`
?

clab/config/utils.go
```diff
@@ -101,35 +102,37 @@ func prepareLinkVars(lIdx int, link *types.Link, varsA, varsB Dict) error {
...
-		if k == vkLinkIP && len(r) == 1 {
-			// calc the remote IP
-			ipF, err := ipFarEndS(v)
-			if err != nil {
-				return fmt.Errorf("%s: %s", link.String(), err)
+		vv := reflect.ValueOf(v)
+		if vv.Kind() == reflect.Slice || vv.Kind() == reflect.Array {
+			// Array/slice should contain 2 values, one for each end of the link
+			if vv.Len() != 2 {
+				return fmt.Errorf("%s: var %s should contain 2 elements, found %d: %v", link.String(), k, vv.Len(), v)
```
hellt, line 115: let's expand `var` to `variable`? since this is a user-facing error, I think it is better to spell the var as a variable

clab/config/utils.go
```diff
@@ -168,25 +171,25 @@ func linkIP(link *types.Link) (string, string, error) {
 	_, okA := link.A.Node.Config.Vars[vkSystemIP]
 	_, okB := link.B.Node.Config.Vars[vkSystemIP]
 	if okA != okB {
-		return "", "", fmt.Errorf("%s var required on all nodes", vkSystemIP)
+		log.Warnf("to auto-generate link IPs, a %s var is required on all nodes", vkSystemIP)
```
hellt, line 174: let's expand var to variable? since this is a user-facing error, I think it is better to spell the var as a variable

clab/config/utils.go
kellerza: Ok, will add `clab_`, though intention was to use clab_ for topology file related items and without for more logical related items items

kellerza: A bit of a slippery slope, since `port` will have to be renamed to `clab_port` in the supplied templates once we auto-populate them

kellerza:
Ok, should be done. Not too happy with the result and feel `clab_` is overused on the logical vars, but can live with it. 

The downside is that for every "auto-generation" feature that we add, all supplied templates & all user templates will have to be adapted,instead of just not specifying these in topo files any longer. So ports, and likely number pools, IP pools later to be consistent.

hellt:
> The downside is that for every "auto-generation" feature that we add, all supplied templates & all user templates will have to be adapted,instead of just not specifying these in topo files any longer. So ports, and likely number pools, IP pools later to be consistent.

What if we have a grouping for clab-specific vars instead? Like that

```yaml
config:
  vars:
    clab:
      system_ip: 1.2.3.4
      what_ever: somevalue
```

kellerza:
You still end up having to change templates before being in a position to use new clab functionally, so no gain really. Topo files and debug prints all increase in length.

hellt:
I don't see how you can NOT change the template if you changed any of the variables. One of the points I wanted to make about making a distinction between "free" variable names and the ones which have a special meaning is also applicable to the templates.

If I read the template and I see that it references `clab_port` variable (or `.clab.port`), to me that is a sign that this variable is coming after processing inside the code, and not something that comes from the user-defined variables.

I believe this visual separation is important

kellerza:
### `port` is probably the best example there is.

Today it is called `port`. If you dont specify it you get an error, the `base` template has `{{ expect .port "portregex" }}` so you need to specify it in the topo file

Tomorrow we start auto-generating this `port` becasue we can from the topo (only if it is not there.. all these things are optional) and now you can simply delete port from the topo file and it still works.

### `link_ip` will be inconsistent

Today it is, either:
- specified as IP on one end, auto-gen for the remote
- a mix of system_ip (which is never auto-generated itself). 
- Tomorrow we might have pools. Pools will now be applied to clab_link_ip variables, or even standard non-clab variables if flexible enough

## alternative
If you really want to see these auto-generated vars (of any kind, today's rules, tomorrow based on templates or pools or number ranges, lets rather do a DIFF on the variable show "-c vars".

- You see what you have supplied. 
- You see all the magic
- `clab_` is not overused. Using it for topological things like clab_links/clab_nodes makes a ton of sense

This is very easy to do and adds more value than trying to name them and covers future evolution much better

kellerza:
one last note: even adding this link_name, link_ip, etc could be done in templates. A template to create vars before the real templating starts

It could simplify the code, expose the `magic` and allow even more flexibility in the long run, but clab_ just complicates this

hellt:
Let's maybe set aside the current templates, they will likely be changed; we never reviewed them so far

As to the variables themselves, what I want to achieve is rather simple:

as a user, I want to be able to create a variable `system-ip` under the `.config.vars` and expect it to be available inside the template. It is unnatural for me to reserve common variable names for clab magic.
Hence my idea was to scope those variables so they won't clash with any of the names a regular template designer might choose.
Either `clab_varname` or `clab.varname` will do right by me.

kellerza:
>  I want to be able to create a variable system-ip under the .config.vars and expect it to be available inside the template. It is unnatural for me to reserve common variable names for clab magic.

As agreed, it's all optional. 

Nothing stops you from using `system_ip` after this PR. Same is true for link_ip / link_name or any of these magic vars

kellerza:
And if something stops you from these, I'll consider it a bug & won't mind fixing it

hellt:
Yes, right now it is allowed, but I you mentioned that you are not happy about that change, and I wanted to see if there is a middle ground approach.

But as it is right now, I am OK to merge. If you don't have anything else to add as part of this PR

kellerza:
:+1: Should be ok to merge.

The optional part I agree with 100%

Not happy forcing `clab_` prefix for logical vars that are optional, but can live with it for now. If you can consistently enforce this  in the long run is to be seen.

Already today, if you use link vars, this rule is not consistent for vars inside `clab_far`. Variables in there was either split (user's intention) or duplicated (magic), some have `clab_`, some not.

PR #544, opened 2021-07-23
Add support for additional ansible groups

### LINKED ISSUE #537: Additional Ansible groups
It would be useful to be able to have additional Ansible groups be generated in the ansible inventory file. For example if you have a spine/leaf topology it would be great to be able to split out the two layers into different ansible groups, instead of only having the `ceos` group (if you are using Arista cEOS containers) and having to constantly fiddle with `--limit`s

sc68cal:
Fixes #537

### DIFF
--- a/clab/inventory.go
+++ b/clab/inventory.go
@@ -61,9 +61,9 @@ func (c *CLab) generateAnsibleInventory(w io.Writer) error {
 
 	for _, n := range c.Nodes {
 		i.Nodes[n.Config().Kind] = append(i.Nodes[n.Config().Kind], n.Config())
-                if n.Config().Labels["ansible-group"] != "" {
-		        i.Groups[n.Config().Labels["ansible-group"]] = append(i.Groups[n.Config().Labels["ansible-group"]], n.Config())
-                }
+		if n.Config().Labels["ansible-group"] != "" {
+			i.Groups[n.Config().Labels["ansible-group"]] = append(i.Groups[n.Config().Labels["ansible-group"]], n.Config())
+		}
 	}
 
 	// sort nodes by name as they are not sorted originally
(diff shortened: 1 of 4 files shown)


sc68cal:
Thinks needs a bit more work, so that you could add multiple groups, like for example

```yaml
    node3:
      kind: srl
      license: test_data/node1.lic
      user: custom
      mgmt_ipv4: 172.100.100.13
      labels:
        node-label: value
        ansible-group: 
          - extra_group
          - one_more_group
```

sc68cal:
TODO: add to docs

hellt:
thanks @sc68cal 
very good PR, esp for adding the test case
I will update the docs

hellt:
@sc68cal check out the commit I added for the docs, all looks good?

I also added an ansible_host parameter under the hosts of the user-defined groups which points to ipv4 address of the node.

sc68cal:
> I also added an ansible_host parameter under the hosts of the user-defined groups which points to ipv4 address of the node.

I think that it is not required to add the IPv4 address of the node for the user-defined groups, provided that the IPv4 address is set for the group that matches the `kind` of the node.

clab/inventory.go
```diff
@@ -38,20 +38,33 @@ func (c *CLab) generateAnsibleInventory(w io.Writer) error {
           ansible_host: {{.MgmtIPv4Address}}
 {{- end}}
 {{- end}}
+{{- range $name, $nodes := .Groups}}
+    {{$name}}:
+      hosts:
+      {{- range $nodes}}
+        {{.LongName}}:
+          ansible_host: {{.MgmtIPv4Address}}
```
sc68cal, line 46: I think this can probably be removed

sc68cal:
Docs look good, thank you

sc68cal:
I think actually this is good as is, if having the `ansible_host` set multiple times (but with the same value) is fine. I am 👍  for merge

PR #547, opened 2021-07-24
cvx-bump-ram

networkop:
Slight memory bump for for the latest cvx image. Doing this globally for the entire ignite runtime since a) it's only used by cvx ATM b) `type NodeConfig struct` doesn't have a memory field.

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -211,6 +211,8 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 		Sandbox:         c.Config.Topology.GetNodeSandbox(nodeName),
 		Kernel:          c.Config.Topology.GetNodeKernel(nodeName),
 		Runtime:         c.Config.Topology.GetNodeRuntime(nodeName),
+		CPU:             c.Config.Topology.GetNodeCPU(nodeName),
+		RAM:             c.Config.Topology.GetNodeRAM(nodeName),
 	}
 
 	log.Debugf("node config: %+v", nodeCfg)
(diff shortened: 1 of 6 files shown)


hellt:
I have the same issue with vrnetlab where I need to drive the CPU/mem req for qemu VM. I was thinking about using an env var for that. 

For example if a user sets env var `CLAB_MEM` then this will be picked up by vrnetlab/ignite and will set the spec accordingly.

How about that?

networkop:
in my case it still needs to be part of `NodeConfig` since it's set at `ContainerCreate` stage. I can add the cpu/memory fields to `NodeConfig` to begin with

hellt:
I meant that you could look for env vars defined in the topo before
creating container to set the mem requirements. But now I'm thinking that
having it under a node makes sense. Ideally with that we should also be
able to set cgroups for containers in the future

networkop:
I've added CPU and RAM fields to `nodeConfig` and now they can be unique for each node. how does this look?

hellt:
Would cvx handle the ram set in 0.8G format?

networkop:
yeah, the format is parsed by `https://github.com/c2h5oh/datasize`

hellt:
At this stage the CPU and RAM won't be read from the clab.yml, but that's fine. I can add it later on when will tackle cpu settings for vrnetlab

networkop:
ah yeah, missed that. let me know if you want me to add it

hellt:
If you can add it here that would be great

networkop:
done

hellt:
thanks @networkop

PR #550, opened 2021-07-26
Palo alto

carlmontanari:
Howdy!

This PR adds support for Palo Alto PA-VM instances via vrnetlab. Also added .idea to gitignore (but as its own commit so we can yoink that out if its unwanted).

Initial testing of all this looks to be working as expected/desired!

Carl

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ container-lab
 containerlab
 site/
 .vscode/
+.idea/
 .DS_Store
 __rd*
 tests/out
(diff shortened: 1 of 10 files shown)


hellt:
looks neat, @carlmontanari 
let me know if you want to run some additional testing before I add a few doc items over and merge it

carlmontanari:
I think this is good to go! I will be doing more testing over the next few weeks and can open PR if I run into any issues!

PR #554, opened 2021-07-27
use correct var for container name

karimra:
This fixes `clab exec` when running against multiple nodes.
A wrong variable name was used.


hellt:
thanks!

PR #555, opened 2021-07-27
fix frr01 example lab file and container names

hartleyj:
Updates lab-example/frr01 such that _run.sh_ uses the correct topology file, and _PC-interfaces.sh_ uses the correct container names.


hellt:
thanks!

PR #558, opened 2021-07-27
ipv6 LLA for veth links on linux nodes

hellt:


hellt:
the tests are failing because with all interfaces ipv6 enabled the node fails to start running, it seems
image

PR #564, opened 2021-08-02
refined file bind mounts for srl

### LINKED ISSUE #562: use fine grained mounts for srlinux
we have a too coarse mount for the whole `/etc/opt/srlinux` directory, which may not be needed or even cause unnecessary clutter

hellt:
fix #562 
left only two mounts:

* config.json -> /etc/opt/srlinux/config.json
* topology.yml -> /tmp/topology.yml

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -120,7 +120,10 @@ func NewContainerLab(opts ...ClabOption) (*CLab, error) {
 		o(c)
 	}
 
-	err := c.parseTopology()
+	var err error
+	if c.TopoFile.path != "" {
+		err = c.parseTopology()
+	}
 
 	return c, err
 }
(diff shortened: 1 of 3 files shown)


jbemmel:
Since topology.yml is the same for every node, why not put it under /clab-xxx/ instead of each /clab-xxx/node[x] subdir?
Especially given that it's mounted read-only anyway

hellt:
It is different for every node because base Mac address is part of this
file

jbemmel:
I see, I thought it was the containerlab topology

Since this is being mounted under /tmp, who reads this (within the container)? If it's just nice to have I would suggest to remove it, since it adds a convention that the platforms don't support natively (unlike the config file)

hellt:
It's the srlinux internal scripts which read this file and make the srlinux
process to start with that config

Issue #567, opened 2021-08-04
cannot unmarshal !!str `...` into types.ConfigDispatcher


edvgui:
Hi!

I have been facing a strange issues when trying to deploy labs with `vr-sros` kind and making use of the `config` option.
```console
$ clab version

                           _                   _       _     
                 _        (_)                 | |     | |    
 ____ ___  ____ | |_  ____ _ ____   ____  ____| | ____| | _  
/ ___) _ \|  _ \|  _)/ _  | |  _ \ / _  )/ ___) |/ _  | || \ 
( (__| |_|| | | | |_( ( | | | | | ( (/ /| |   | ( ( | | |_) )
\____)___/|_| |_|\___)_||_|_|_| |_|\____)_|   |_|\_||_|____/ 

    version: 0.16.0
     commit: 4d15005
       date: 2021-08-03T09:03:15Z
     source: https://github.com/srl-labs/containerlab
 rel. notes: https://containerlab.srlinux.dev/rn/0.16/
$ sudo clab deploy --topo example.clab.yml 
FATA[0000] failed to read topology file: yaml: unmarshal errors:
  line 9: cannot unmarshal !!str `./confi...` into types.ConfigDispatcher
  line 19: cannot unmarshal !!str `./confi...` into types.ConfigDispatcher
```

The lab I am trying to deploy is the following:
```yaml
name: nokia
topology:
  nodes:
    nokia-east:
      kind: vr-sros
      type: sr-1s
      image: vrnetlab/vr-sros:20.10.R1
      license: sros-20.7.R1.qcow2.license
      config: ./config/nokia/east.cfg
      ports:
        - 20022:22
        - 20830:830
      mgmt_ipv4: 172.20.20.21
    nokia-west:
      kind: vr-sros
      type: sr-1s
      image: vrnetlab/vr-sros:20.10.R1
      license: sros-20.7.R1.qcow2.license
      config: ./config/nokia/west.cfg
      ports:
        - 21022:22
        - 21830:830
      mgmt_ipv4: 172.20.20.31
  links:
    - endpoints: ["nokia-east:eth1", "nokia-west:eth1"]
```

The strangest thing is that is used to work (at version `0.14.3` at least):
```console
$ clab version

                           _                   _       _     
                 _        (_)                 | |     | |    
 ____ ___  ____ | |_  ____ _ ____   ____  ____| | ____| | _  
[...]

PR #568, opened 2021-08-04
Replacing usage of config in examples by startup-config

edvgui:
I got some issues recently trying to use the `config` property to pass a startup config file as mentioned in #567.

The documentation was only partially updated and some examples where still using the `config` property.

This MR fixes this.


hellt:
thanks @edvgui 
the changes are live on the site

PR #573, opened 2021-08-05
allow custom lab prefix

karimra:
This PR allows setting a custom prefix for the lab, instead of `clab`.
The prefix is configured in the topology file using field `prefix`.
For e.g:
```yaml
name: lab1
prefix: custom-prefix
topology:
  nodes:
    node1:
```
- `node1` longName will be `custom-prefix-lab1-node1`
- the lab directory will be `custom-prefix-lab1`

If not set ( not present ) the default `clab` is used.
If present and set to `""`, the default does not apply. In this case, the `node` longName will be `lab1-node1` and the lab directory will be `lab1`

The prefix value impacts the nodes longName, the hosts entries, the lab directory name,.. but it does not impact the docker labels, those still have prefix `clab-`

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	// prefix is used to distinct containerlab created files/dirs/containers
-	prefix = "clab"
+	defaultPrefix = "clab"
 	// a name of a docker network that nodes management interfaces connect to
 	dockerNetName     = "clab"
 	dockerNetIPv4Addr = "172.20.20.0/24"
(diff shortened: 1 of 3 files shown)


clab/config.go
```diff
@@ -79,12 +80,21 @@ func (c *CLab) parseTopology() error {
 	if c.Config.ConfigPath == "" {
 		c.Config.ConfigPath, _ = filepath.Abs(os.Getenv("PWD"))
 	}
+	if c.Config.Prefix == nil {
+		c.Config.Prefix = new(string)
+		*c.Config.Prefix = defaultPrefix
```
hellt, line 85: can this be 

```go
c.Config.Prefix = &defaultPrefix
```

https://replit.com/join/nwneuvdywh-romandodin

karimra: that's possible only if you change `defaultPrefix` from const to a variable

hellt: ah, roger...

hellt:
thanks, really nice one!

PR #580, opened 2021-08-10
chore: Added a line about UUIDs

cummings-chris:
Very simple documentation update to be explicit about how to add UUIDs that are required by some SR-OS license files.


hellt:
thanks @cummings-chris 
looks very good
I just moved this block under a `license` section so things will stay sorted out nicely

PR #582, opened 2021-08-11
Linux kind documentation fix

GrigoriyMikhalkin:
Linux Kind page in docs uses invalid topology definition as an example(missing `nodes`). Fixed it)


hellt:
thanks! spot on!

PR #587, opened 2021-08-12
Populate /etc/hosts for static IPs

jbemmel:
For demos, it is often handy to refer to nodes by name rather than IP. If a lab contains static IPs, we can add them to /etc/hosts
(ipv4 only for now)

PR became a bit of a mess, changes are only here:
 cmd/deploy.go            | 17 +++++++++++++++++
 runtime/docker/docker.go |  1 +
 types/types.go           |  1 +


jbemmel:
Created a new clean PR

PR #589, opened 2021-08-13
Add optional startup delay (worker sleep)

jbemmel:
Use case: For a demo, Telegraf startup needs to be delayed until all SRL nodes are up and running. Time dependencies are a bit of a hack, but it can help to workaround issues

Explicit depends-on (with optional delay) would be nice too, but this one is an easy add-on


hellt:
Hi @karimra 
I wonder if you have some nice ideas how to overcome the following limitation of nodes/links creation scheduling:

right now we get to links creation only after *all* the nodes have been created, so if we have delayed a creation of a certain node in a lab (as done in this PR), we won't start creating **any** links, even for the nodes that haven't been delayed and ready to accept links

karimra:
It should be doable this way:
- Use two maps of type `map[string]struct{}`, the first tracks created nodes `createdNodes`, the second tracks scheduled links `scheduledLinks`
- The Node deploy method adds the name of the created node to `createdNodes` when it's done.

- A separate goroutine loops through the links( `c.Links` ), checks if both nodes that make up that link are created ( i.e exist in the above map)
- If both exist and the link is not in `scheduledLinks` , add the link name to `scheduledLinks` and send the link to the linkChan to be created
- If one of the above conditions is not met, continue to next link.
- The link scheduling goroutine should return when `len(c.Links) == len(scheduledLinks)`
- Since the two maps will be accessed by multiple goroutines, there should be a mutex to control that

I think this whole thing is too complicated, why not create 2 labs where the second one contains the nodes that need to be delayed, wrap the whole thing in a script with a sleep statement.
For this use case, there is no need to create links to telegraf, only mgmt.

hellt:
@karimra what do you say to the brute force approach I took in ce1fe7e?
the first idea that came to my mind is to return the waitgroups from nodes creation and wait for them after the scheduling of the links

when the links are about to be created, I copy a map of links and go over it checking if the NSPath of both endpoints of the links is not empty (that is denotes if a node was created by that time)

karimra:
Is NSPath always populated? even in the case of mgmt-net and bridge nodes ?

hellt:
Indeed, that may cause trouble as the tests show. 

What if we are a field to the NodeConfig such as "deploymentStatus" and we will set it to "created" in the end of createNodes function?

And then we will check if this node is created when we schedule links to linkschan?

karimra:
> Indeed, that may cause trouble as the tests show.
> 
> What if we add a field to the NodeConfig such as "deploymentStatus" and we will set it to "created" in the end of createNodes function?
> 
> And then we will check if this node is created when we schedule links to linkschan?

that works as well, replacing the `createdNodes` map

hellt:
Looks like it works nicely with docker but something is off with containerd and only with generate command. Need to look deeper :(

hellt:
@steiler as a containerd expert, do you want to have a peek as to why the tests for generate command have an endless loop without break when we check links?

Issue #585, opened 2021-08-11
Proposal: support env variable expansion in topology file


GrigoriyMikhalkin:
Currently(AFAIK) it's only possible to define static environment variables in topology file. But in certain scenarios, support for environment variable expansion would be useful.

For example, user can have a rule in a Makefile, that defines environment variable based on results of some previous action and then tries to deploy topology, expecting that value of this env var will used there.

To be clear, i think environment variable expansion would be useful not only in `env` field, but all over topology definition(i.e. `binds`, `ports`, etc). Here's two examples of how it could be used:

1. 
```
env:
  ENV: $(ENV)
```
2.
```
binds:
  $(PATH_TO)/something:/something/
```

hellt:
hi @GrigoriyMikhalkin 
that makes sense
I think what also makes sense is to make a certain rule to promote all env vars of the shell that runs containerlab to env vars of the node/kind

for example, if in the current shell the following vars are set:

```
ENV_A=valueA
ENV_B=valueB
```

then if we, say, have the following _magic_ variable defined in the topo:

```
env:
  __IMPORT_ENVS: true
```

then we make all the ENV vars of the shell part of the map

GrigoriyMikhalkin:
@hellt Thanks for quick answer! Yep, also regarding precedence -- i guess it would make sense in that case, if variable is defined both in shell and in topo file, prefer value from topo file?

Also, would it be ok if i would take this issue?

hellt:
yes, values from the topo should take precedence over the external env vars
One slight change to your original proposal would be to use shell variable syntax `${VAR_NAME}` if that makes sense

GrigoriyMikhalkin:
Yep, right, `${VAR_NAME}` would be more disambiguous.

hellt:
delivered in #590

PR #590, opened 2021-08-14
Environment variable expansion

GrigoriyMikhalkin:
Implements https://github.com/srl-labs/containerlab/issues/585

Includes both environment variable expansion and importing, when special variable `__IMPORT_ENVS` is set to `true`


hellt:
great stuff, thanks @GrigoriyMikhalkin

PR #591, opened 2021-08-18
Add duplicate mgmt IP check

jbemmel:
When mixing dynamic and static IPs, Docker may assign a duplicate IP. This adds a simple check to print an error message to alert the user


hellt:
Should have been resolved in a merged #592

PR #592, opened 2021-08-19
Schedule nodes with static IPs before the ones with dynamic IPs

karimra:
This PR changes the creation order of nodes by scheduling the ones with static IPs first to avoid dockerd assigning a dynamic IP that is meant for another node (static)

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -165,19 +165,17 @@ func (c *CLab) CreateNodes(ctx context.Context, maxWorkers uint, serialNodes map
 		}
 		dynIPNodes[name] = n
 	}
-	log.Debug("scheduling nodes with static IPs first...")
-	c.createNodes(ctx, int(maxWorkers), serialNodes, staticIPNodes)
-	log.Debug("scheduling nodes with dynamic IPs next...")
-	c.createNodes(ctx, int(maxWorkers), serialNodes, dynIPNodes)
+	if len(staticIPNodes) > 0 {
+		log.Debug("scheduling nodes with static IPs...")
+		c.createNodes(ctx, int(maxWorkers), serialNodes, staticIPNodes)
+	}
+	if len(dynIPNodes) > 0 {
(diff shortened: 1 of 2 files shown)


clab/clab.go
```diff
@@ -155,13 +155,33 @@ func (c *CLab) GlobalRuntime() runtime.ContainerRuntime {
...
+	dynIPNodes := make(map[string]nodes.Node)
 
+	for name, n := range c.Nodes {
+		if n.Config().MgmtIPv4Address != "" || n.Config().MgmtIPv6Address != "" {
+			staticIPNodes[name] = n
+			continue
+		}
+		dynIPNodes[name] = n
+	}
+	log.Debug("scheduling nodes with static IPs first...")
```
hellt, line 168: What if we do this only if static nodes are present? Doing a length check here, as well as keeping the code in createNodes as it is now and just removing the return statement if length is 0?

That way we avoid issuing a debug print here that is not needed if there are now nodes with static IP and won't call a function

karimra: I fail to see the advantage in that change,
what if all the nodes are with static IPs and the dynamic nodes are 0?
should we do a length for the dynamic IPs as well ?

hellt:
It might be more clear flow-wise to not call createNodes knowing that there
is nothing to create beforehand. But that is just my view

That will also disable a log message to be printed that would have no
meaning in case there are no nodes to schedule

Yes, that would apply both to static and dynamic nodes

clab/clab.go
hellt: I merely wanted to prevent a log message from printing 

```
log.Debug("scheduling nodes with static IPs first...")
```

if it is known if there are no nodes to schedule. But can be just me

karimra: I will change it, but I still think it's less optimized.
Scheduling doesn't mean there are node to be scheduled, if there are none, it won't happen.

hellt:
thanks!

PR #599, opened 2021-08-27
Updated vr-ros to support startup configs

nlgotz:
Updated vr-ros to support startup-config option and updated documentation for the feature

This will take the startup-config option and place it into /ftpboot/config.auto.rsc that Docker will then be able to push via FTP. Mikrotik supports automatic config import for files ending in .auto.rsc which this takes advantage of


hellt:
thanks @nlgotz!

PR #600, opened 2021-08-28
making sure scrapli cleanup doesnt break things!

carlmontanari:
howdy folks!

@hellt is already in the loop on some of this, but for anyone else interested -- I've been doing a bit of housekeeping on scrapligo things and made some breaking-ish changes; this draft (until I finalize things ive been working on) is pinned to a dev branch and contains the little fixes that will be necessary if ya'll want to update the scrapligo pin. I think I should be fairly close to being done changing/breaking things and I will update this branch w/ pin to master branch (or perhaps even a real release finally) once I get to that point.

Carl

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,3 +33,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	inet.af/netaddr v0.0.0-20210521171555-9ee55bc0c50b
 )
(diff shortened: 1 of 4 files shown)


hellt:
super dooper @carlmontanari 
thanks

PR #601, opened 2021-08-30
Fix: in filters exists should be used as operator not match value

GrigoriyMikhalkin:
In `exec` cmd, in case if `label` flag was passed without `=` operator, `exists` should be assigned as an operator(as it's expected in runtimes method `buildFilterString`). Instead, it's used as match value(which causes filter misconstruction).


hellt:
nice catch, thanks!

Issue #598, opened 2021-08-26
Proposal: add exec property to nodes component


GrigoriyMikhalkin:
### Problem that this proposal tries to solve
Currently user can specify custom `cmd` for a `node` component. It sets/overrides command specified in container image. But there's no option for the case when user would want to execute additional commands. 

For example, in my lab i use [this image](https://github.com/panubo/docker-sshd/blob/master/Dockerfile) and additionally want to execute `python` installation command. Initial Dockerfile defines `CMD`, so to keep it, i have to run it in new command also. Final cmd looks `bash -c "apk add python3 && /usr/sbin/sshd -D -e -f /etc/ssh/sshd_config"`.

IMO, that's quite inconvenient and it would be great if it be possible to specify additional commands to execute in container, that would not override original image command.

### Proposed solution
Docker has [exec](https://docs.docker.com/engine/reference/commandline/exec/) command that runs command in a running container.  It should be possible to implement similar behavior in containerlab.
 
`exec` list field could be added to a `node` component. Which would look like this:
```
nodes:
  node1:
    exec:
     - bash cmd1.sh
     - bash cmd2.sh
```
Unlike `cmd` field, entries in `exec` field will be executed in running container, without overriding initial command.

hellt:
thanks for the valuable contribution @GrigoriyMikhalkin 
since #602 is now merged, I close this

PR #602, opened 2021-08-31
Added exec property to nodes

GrigoriyMikhalkin:
Attempt to implement https://github.com/srl-labs/containerlab/issues/598.

This PR implements `exec` list field in `node` component and refactors code(mostly in `cmd/exec.go`). Docs not updated(yet).

Here is topo that was used for testing this feature: https://gist.github.com/GrigoriyMikhalkin/23fe014660a17ae45ee5d0c5d180d0a2

### DIFF
--- a/clab/config_test.go
+++ b/clab/config_test.go
@@ -373,55 +373,55 @@ func TestLabelsInit(t *testing.T) {
 			got:  "test_data/topo1.yml",
 			node: "node1",
 			want: map[string]string{
-				"containerlab":      "topo1",
-				"clab-node-name":    "node1",
-				"clab-node-kind":    "srl",
-				"clab-node-type":    "ixr6",
-				"clab-node-group":   "",
-				"clab-node-lab-dir": "./clab-topo1/node1",
-				"clab-topo-file":    "./test_data/topo1.yml",
+				ContainerlabLabel: "topo1",
+				NodeNameLabel:     "node1",
+				NodeKindLabel:     "srl",
+				NodeTypeLabel:     "ixr6",
+				NodeGroupLabel:    "",
+				NodeLabDirLabel:   "./clab-topo1/node1",
+				TopoFileLabel:     "./test_data/topo1.yml",
 			},
 		},
 		"custom_node_label": {
 			got:  "test_data/topo1.yml",
 			node: "node2",
 			want: map[string]string{
-				"containerlab":      "topo1",
-				"clab-node-name":    "node2",
-				"clab-node-kind":    "srl",
-				"clab-node-type":    "ixrd2",
-				"clab-node-group":   "",
-				"clab-node-lab-dir": "./clab-topo1/node2",
-				"clab-topo-file":    "./test_data/topo1.yml",
-				"node-label":        "value",
+				ContainerlabLabel: "topo1",
+				NodeNameLabel:     "node2",
+				NodeKindLabel:     "srl",
+				NodeTypeLabel:     "ixrd2",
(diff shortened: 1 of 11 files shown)


hellt:
thanks @GrigoriyMikhalkin 
I will try to review this either later today or on Friday

GrigoriyMikhalkin:
> is this still a draft because of the docs? or smth else you plan to add code-wise?
Because of docs and general approach.

clab/config_test.go
```diff
@@ -436,8 +436,8 @@ func TestLabelsInit(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			tc.want["clab-node-lab-dir"], _ = resolvePath(tc.want["clab-node-lab-dir"])
-			tc.want["clab-topo-file"], _ = resolvePath(tc.want["clab-topo-file"])
+			tc.want[NodeLabDirLabel], _ = resolvePath(tc.want[NodeLabDirLabel])
+			tc.want[TopoFileLabel], _ = resolvePath(tc.want["clab-topo-file"])
```
hellt, line 440: should be `resolvePath(tc.want[TopoFileLabel])` I presume

types/topology.go
```diff
@@ -285,6 +285,19 @@ func (t *Topology) GetNodeCmd(name string) string {
...
+			return ndef.GetExec()
+		}
+		if t.GetKind(t.GetNodeKind(name)).GetExec() != nil {
+			return t.GetKind(t.GetNodeKind(name)).GetExec()
+		}
+		return t.GetDefaults().GetExec()
+	}
+	return nil
+}
+
```
hellt, line 300: maybe lets do an append from the defaults->kinds->node here?
for example, I might want by default to execute `bash cmd1.sh` on all nodes, then for linux kinds I want to add `bash cmd2.sh` and then for a specific `nodeX` I add `bash cmd3.sh`

In the end, for nodeX the commands will get appended starting from defaults and will comprise the list

- bash cmd1.sh 
- bash cmd2.sh
- bash cmd3.sh

cmd/deploy.go
```diff
@@ -216,6 +218,27 @@ var deployCmd = &cobra.Command{
...
+			}
+		}
+		if format == "json" && (len(execJSONResult) > 0) {
+			result, err := json.Marshal(execJSONResult)
+			if err != nil {
+				log.Errorf("Issue converting exec results to json %v", err)
+			}
+			fmt.Println(string(result))
+		}
+
```
hellt, line 241: every node interface has a `PostDeploy()` method. In this method we run stuff that should happen after the node has been deployed

Maybe it makes sense to call the `execCmds` in that postDeploy phase? In that case all the runtime will be initialised already.

GrigoriyMikhalkin: Yep, i also thought about it. Problem with it is that it looks like it causing unnecessary repetition. Since `execCmds` code, probably will be the same for all types of nodes.

hellt: okay, that is fair
also we might soon have another action that is intended to be run right after the lab is deployed so this block can grow a bit

hellt:
Hi @GrigoriyMikhalkin 
how do you find my recent polishing on this PR?
Basically I just moved the exec printouts before the summary table to keep things nicely timelined and added a doc entry with a runtime test

GrigoriyMikhalkin:
@hellt Looks good to me)

hellt:
awesome, thanks @GrigoriyMikhalkin
definitely a vary valuable contribution that I wanted to have as well

PR #609, opened 2021-09-07
avoid panic when importing Envs if NodeDefinition is nil

karimra:


hellt:
Thanks @karimra !

/cc @GrigoriyMikhalkin

PR #612, opened 2021-09-10
Add 'extras' 'agents' config option for SRL nodes

jbemmel:
SR Linux supports custom agents; this feature copies agent YAML files under /etc/opt/srlinux/appmgr

TODO: Documentation (coming next)


hellt:
@jbemmel can you please check the latest commit from my side

I cleaned it up a bit and made a change to the yaml structure:

`agents -> srl-agents` - this is done to make a visual bind to the srl system. Since `extras` can be really added to any node of the topology, a plain `agents` doesn't say much about which feature is it

PR #613, opened 2021-09-10
Removing the need for srl template config

### LINKED ISSUE #608: rework srlinux config provisioning
at the moment, srlinux configuration is generate before the srlinux node starts
during the config generation the following stuff happens

1. we take the config template that is checkout to this repo https://github.com/srl-labs/containerlab/blob/master/nodes/srl/srl.cfg
2. we generate the tls cert with key
3. we use the generated cert and key and run a templating function over the template in step 1

this results in certs to be already available in the resulting config, as well as the gnmi and json-rpc interfaces enabled.

The problem is that the template we use in step 1 is an srl config from circa v20.3, meaning that the changes made to the default config (like ACLs, etc) are not captured there. This is a big issue for users, since they need to be aware of that so we need to get rid of that.

A solution to this problem can be with pushing only the relevant config changes using scrapligo when the srl node finishes booting. At that time, in the postDeploy phase, we can push the certs, key and enable gnmi and all that good stuff with scrapligo over `docker exec` pty. Pretty much like we currently do for arista to provision its management interface in its postdeploy phase https://github.com/srl-labs/containerlab/blob/947e5b972dfae7a686ba4c94f7ee718ce76700c4/nodes/ceos/ceos.go#L146

hellt:
This PR removes the need to have a default config for SRL. Having the default config poses some problems, as default/factory config grows over time (for example 21.6.2 added ldp and the relevant ACLs) and we need to somehow know which config is relevant for which version.

The approach taken in this PR is as follows:

1. instead of using the default config and template it with TLS/gnmi/json-rpc blobs we let srlinux to start with factory config
2. once it has started, we add the TLS/gnmi/json-rpc config blocks using scrapligo over the factory config that this SRL booted with

@karimra I would like to ask your opinion on this draft approach 

close #608

### DIFF
--- a/nodes/srl/srl.go
+++ b/nodes/srl/srl.go
@@ -14,8 +14,10 @@ import (
 	"path/filepath"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/pkg/errors"
+	"github.com/scrapli/scrapligo/driver/base"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/srl-labs/containerlab/cert"
@@ -27,6 +29,8 @@ import (
 
 const (
 	srlDefaultType = "ixrd2"
+
+	tlsServerProfileName = "clab-profile"
 )
 
 var (
@@ -161,8 +165,13 @@ func (s *srl) Deploy(ctx context.Context) error {
 }
 
 func (s *srl) PostDeploy(ctx context.Context, ns map[string]nodes.Node) error {
+	if s.cfg.StartupConfig == "" {
+		log.Infof("Running postdeploy actions for Nokia SR Linux '%s' node", s.cfg.ShortName)
+		return addDefaultConfig(ctx, s.runtime, s.cfg)
+	}
 	return nil
 }
+
 func (s *srl) Destroy(ctx context.Context) error {
 	// return s.runtime.DeleteContainer(ctx, s.cfg)
 	return nil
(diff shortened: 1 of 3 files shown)


nodes/srl/srl.go
```diff
@@ -161,8 +165,13 @@ func (s *srl) Deploy(ctx context.Context) error {
 }
 
 func (s *srl) PostDeploy(ctx context.Context, ns map[string]nodes.Node) error {
+	if s.cfg.StartupConfig == "" {
```
hellt, line 168: the postdeploy action where we add config on top of the factory one is only needed if no startup config is present.

also we need to make sure that if the config inside the nodeDir is present we also don't call this postdeploy action, as the config is nodeDir should already have all the needed stuff.
**TODO: THIS IS NOT TAKEN INTO ACCOUNT YET**

utils/networkcli.go
```diff
@@ -25,6 +26,9 @@ func SpawnCLIviaExec(platform, contName string) (*network.Driver, error) {
 	var d *network.Driver
 	var err error
 
+	// uncomment to enable logging
+	logging.SetDebugLogger(log.Print)
```
hellt, line 30: TODO: remove when testing is finished

nodes/srl/srl.go
hellt: resolved in eb11522

karimra:
Why not run sr_cli commands using docker exec api? This assumes ssh will always be up at startup in future releases.

hellt:
This code practically executes commands over docker exec sr_cli. Just instead of a single command-by-command execution it spawns a terminal session and sends commands over it. 


In other words ssh is not assumed, and I can add ssh enablement just in case it will ever be disabled in the factory config

karimra:
Ok, that's one way to do it, at least it's not dependent on ssh.
But it's not runtime agnostic, why not use the runtime interface{} Exec method ?

hellt:
Yes, I will need to add runtime specific rules for spawning the CLI. 

I tried to provision a key/cert from the CLI, but I faced an issue with pasting cert string which should have newlines. Didn't spend too much time on it though. Resorted to a spawned CLI session instead

Will add runtime specific opening commands to this method

karimra:
What I mean is why using a cli method while this go method is available with the runtime interface?
https://github.com/srl-labs/containerlab/blob/master/runtime/runtime.go#L47

hellt:
I failed to configure the certs via the exec

karimra:
Because of new lines ? it has to be properly quoted.
I think that approach will be better than writing cli commands for each runtime.

hellt:
@karimra yes, I stopped when hit newline parsing error
ptal here https://github.com/srl-labs/containerlab/pull/616

hellt:
superseded by #616

PR #614, opened 2021-09-11
add newline to the last ignite udev rule

GrigoriyMikhalkin:
Older versions of `systemd-udevd` require new line at the end of the file, otherwise error `invalid key/value pair` is occurring.


hellt:
Cool! Thanks @GrigoriyMikhalkin 

@networkop looks good to you?

PR #615, opened 2021-09-12
Dell FTOS10v Support

log1cb0mb:
- Used `n9kv` kind to test FTOS image and seems to work as expected so used same kind module for `vr_ftosv`.
- Image built with latest hellt/vrnetlab
- SSH works and tested dataplane L3 only so far.


hellt:
Super PR, thank you for taking all the needed steps!

PR #616, opened 2021-09-13
srl config with sr_cli

### LINKED ISSUE #608: rework srlinux config provisioning

hellt:
This PR supersedes #613 
instead of using the spawned PTY and run CLI commands over it, it reuses the existing runtime.Exec method

close #608


hellt:
Looks like this did it https://github.com/srl-labs/containerlab/pull/616/commits/8104041ca7134850e0e514cc6a284605e6e98d50

hellt:
nah, not really

hellt:
@karimra 
seems like I found a way to do this via `sr_cli -ed < file.cfg` approach. Latest commit adds this

PR #617, opened 2021-09-14
added change into the topo dir to resolve paths

### LINKED ISSUE #607: make `destroy` command to use relative paths as referenced in topo file
if a topo file contains a relative path to some bind mount, then when doing `destroy -a` while not being in the same directory where the local path can be resolved, clab will fail.

The destroy needs to somehow change into the directory from which the `deploy` command was called to resolve the paths and destroy the lab succesfully

hellt:
close #607


hellt:
need to think how to deal with the paths...

PR #625, opened 2021-09-20
quick setup clos02

siva19susi:
Adding quick-setup script for deploying underlay connectivity for 5-stage CLOS (clos02).


lab-examples/clos02/connectivity.csv
```diff
@@ -0,0 +1,57 @@
+Source,Interface,Towards,IPv4,IPv4 Prefix,IPv6,IPv6 Prefix
```
hellt, line 1: what is this csv for?

lab-examples/clos02/README.md
```diff
@@ -0,0 +1,10 @@
+# SRL Containerlab (clos02)
```
hellt, line 1: Thanks @siva19susi 

can you please give here a small explanation as to what is being configured with those config files

i.e. which protocols between which DC layers, etc
I will copy it then to the examples description so that it will get rendered in the docs site

lab-examples/clos02/setup.sh
```diff
@@ -0,0 +1,64 @@
...
+  echo $OUT | grep -q -e '\"operation\": \"UPDATE\"'
+  if [ $? -eq 0 ]; then
+    SAVE=$(docker exec -it clab-clos02-$1 sr_cli "save startup" 2>&1)
+  else
+    echo "Error: Unable to push config. Check if NE is reachable with gRPC enabled."
+  fi
+}
+
+if [ $# -eq 0 ]; then
+  echo "Error: No topology file passed."
```
hellt, line 16: just exit here, no need for `else`

lab-examples/clos02/setup.sh
```diff
@@ -0,0 +1,64 @@
...
+else
+  DEPLOY=$(containerlab deploy --topo $1 2>&1)
```
hellt, line 18: let's not hide the deployment process
instead, call the deploy command directly, so that its output is visible as the labs undergoes the deployment process
then you just check that exit status and exit if it is not `0`

lab-examples/clos02/setup.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
hellt, line 1: you may want to step this game a bit and launch configuration steps concurrently, checking that they all succeed in the end

for example like here https://stackoverflow.com/a/29535256/4902910

lab-examples/clos02/setup.sh
```diff
@@ -0,0 +1,64 @@
...
+
+CFG_DIR=./configs
+
+configure_SRL () {
+  OUT=$(gnmic -a clab-clos02-$1 --timeout 30s -u admin -p admin -e json_ietf --skip-verify set --update-path / --update-file $CFG_DIR/$1.yaml 2>&1)
+  echo $OUT | grep -q -e '\"operation\": \"UPDATE\"'
+  if [ $? -eq 0 ]; then
+    SAVE=$(docker exec -it clab-clos02-$1 sr_cli "save startup" 2>&1)
+  else
+    echo "Error: Unable to push config. Check if NE is reachable with gRPC enabled."
```
hellt, line 11: I think here it makes sense to show the `$OUT` as well, as in the case of a failure it is interesting to see what went wrong

hellt (changes requested):

lab-examples/clos02/connectivity.csv
siva19susi: It denotes the complete IP assignment information for clos02 topology.

lab-examples/clos02/README.md
siva19susi: Sure. Will add them through.

lab-examples/clos02/connectivity.csv
siva19susi: Will be added into README itself.

lab-examples/clos02/setup.sh
siva19susi: Removed deploy from setup. Rather kept it as a manual execution.

lab-examples/clos02/setup.sh
siva19susi: Removed.

lab-examples/clos02/setup.sh
siva19susi: $OUT does not have any valid output. So push it into /dev/null.

lab-examples/clos02/setup.sh
siva19susi: Added.

lab-examples/clos02/connectivity.csv
hellt: yes, I think this can be removed from the repo and instead the table in the readme should be used instead

siva19susi: Have already added.

hellt:
Thanks for this great contribution @siva19susi

PR #626, opened 2021-09-20
feat: initial support for vqfx in containerlab

chriscummings-esnet:
This adds initial support to run vQFX images in containerlab. I think that the `vrnetlab` side of things might need some work for supporting the `.qcow2` style images, but this does indeed work using `18.4R1.8` which is based off the `.vmdk` style images. I think that from a clab standpoint not much will need to change though, so it might be worth considering this separately from the `vrnetlab` changes, however, I could see lack of support for `qcow` style causing a support burden.

Also, I have yet to update the docs, but figured I'd throw some code out first before getting to that point.


hellt:
Hi @chriscummings-esnet 
that looks good
let's see through vrnetlab PR completion so that we can merge them simultaneously

chriscummings-esnet:
Also, another thing that will need to be fixed (not sure if it's a requirement for merging, but I want it either way) is getting config loading/saving working with vQFX, as it doesn't do anything currently using this branch.

hellt:
that we can defo tackle in the subsequent PRs

hellt:
merging this

a few todo things that we have left @chriscummings-esnet 

* provide a list of supported/tested version in both vrnetlab readme and containerlab readme for vqfx
* in vrnetlab add config for netconf, snmp (and gnmi if supported)
* update vr-vqfx.md in containerlab to reflect the tested versions and known limitations

PR #632, opened 2021-09-28
Fixed SONiC node kind

marcosfsch:


hellt:
Thanks @marcosfsch! Good catch

PR #636, opened 2021-10-01
docs: veth links MTU fix

GrigoriyMikhalkin:
[In 0.15 release](https://containerlab.srlinux.dev/rn/0.15/#miscellaneous), default MTU for veth links has been [changed to 9500](https://github.com/srl-labs/containerlab/blob/dd122d0510f4545bd185f2bd11acdb497eefa52a/clab/config.go#L37). Docs still mention old value(65000).


hellt:
thanks! good catch

PR #639, opened 2021-10-04
SR Linux ready time out to follow global runtime timeout

hellt:
To make a configurable timeout for sr linux Ready function the timeout is changed from fixed 30s interval to the value of the runtime timeout.

By default runtime timeout as well is set to 30s, but can be changed with a global `--timeout` flag

### DIFF
--- a/nodes/node.go
+++ b/nodes/node.go
@@ -30,7 +30,7 @@ const (
 	NodeKindHOST       = "host"
 	NodeKindLinux      = "linux"
 	NodeKindMySocketIO = "mysocketio"
-	NodeKindOVS        = "bridge-ovs"
+	NodeKindOVS        = "ovs-bridge"
 	NodeKindSonic      = "sonic-vs"
 	NodeKindSRL        = "srl"
 	NodeKindVrCSR      = "vr-csr"
--- a/nodes/srl/srl.go
+++ b/nodes/srl/srl.go
@@ -31,8 +31,7 @@ import (
 const (
 	srlDefaultType = "ixrd2"
 
-	readyTimeout = time.Second * 30 // max wait time for node to boot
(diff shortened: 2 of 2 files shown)


hellt:
wdyt about this change @karimra ?

karimra:
I think this might help, but the issue comes from the initial context created here: https://github.com/srl-labs/containerlab/blob/master/cmd/deploy.go#L67
That context should be created with `context.WithTimeout(context.Background(), timeout)` so that the timeout is propagated to the children contexts created afterwards.
The current change sets a 30s timeout for the ready method alone, not from the start of the deploy cmd.
The 30s default value might need to be increased since we are starting to wait for SRL to come up to push additional configuration.

hellt:
yes, but do we want to have an enforced global timeout in the first place?
This can get tricky as the bigger the lab, the longer overall it takes to deploy it.

By having timeouts for certain operations (container runtime requests, srl ready function) we guard ourselves from hanging or long-executing tasks, but we don't say that the whole deployment should finish in X seconds.

karimra:
That's what the original timeout is supposed to be.
It can be a little confusing for a user trying to troubleshoot and not seeing the lab fail after the expected timeout.
If the timer is to be changed to an individual op timeout, it has to be clarified in the docs together with the number of expected op for each command.

Another option is to leave is as it is and use a specific timeout for SRL ready func, which I think makes more sense since the runtime timeout and SRL boot time are not correlated.

karimra:
it could be a const for now, then wait to see if there are use cases where it needs to be configurable. 30s seems to work fine

hellt:
yes, lemme revert the change to a const with a longer timeout

30s was not enough apparently as some labs were _slow_ to boot srl in that timeframe.
will use 90s just in case

Issue #651, opened 2021-10-13
demand config engine templates only if vars were defined for this node 


hellt:
as reported by a user:

---

Something else I just noticed is that when having different node types the config engine requires me to have a tmpl file for all the kinds in my lab.. I have a lab with a vr-sros and a linux container. The Linux container is ready as it is but for my vr-sros I'd like to leverage the config engine to add some config. 
When trying to render the template it throws an error when I don't have a template for `*__linux.tmpl`. See below:
``` 
[root@comlab-8 sros_tacplus]# clab config -t sros_tacplus.clab.yml template -p . -l tacacs
INFO[0000] Parsing & checking topology file: sros_tacplus.clab.yml
Error: could not load templates from *__linux.tmpl: template: pattern matches no files: `*__linux.tmpl`
```
  
I personally think this is ok. What I believe could be improved is that this error is thrown as well although I specifically define my sros box as a filter for template rendering:
```
[root@comlab-8 sros_tacplus]# clab config -t sros_tacplus.clab.yml template -p . -l tacacs -f sros1
INFO[0000] Parsing & checking topology file: sros_tacplus.clab.yml
Error: could not load templates from *__linux.tmpl: template: pattern matches no files: `*__linux.tmpl`
```

This is the same for sending the config. I helped myself by creating an empty tacacs__linux.tmpl file and using the node filter for config deployment. (as without the node filter it complains about the empty template file as well)

PR #675, opened 2021-11-02
Make "clab configure" to skip nodes without templates or credentials

LimeHat:
Lazy fix for #651. It is still going to connect to all nodes (if we have credentials) but won't send anything without a template.

### DIFF
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -71,19 +71,18 @@ func RenderAll(allnodes map[string]*NodeConfig) error {
 		for _, baseN := range TemplateNames {
 			tmplN := fmt.Sprintf("%s__%s.tmpl", baseN, nc.Vars[vkRole])
 			log.Debugf("Looking up template %v", tmplN)
-			l := tmpl.Lookup(tmplN)
-			if l == nil {
+			if l := tmpl.Lookup(tmplN); l == nil {
 				err := LoadTemplates(tmpl, fmt.Sprintf("%s", nc.Vars[vkRole]))
 				if err != nil {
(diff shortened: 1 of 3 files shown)


clab/config/template.go
```diff
@@ -68,22 +68,25 @@ func RenderAll(allnodes map[string]*NodeConfig) error {
...
+			if l := tmpl.Lookup(tmplN); l == nil {
 				err := LoadTemplates(tmpl, fmt.Sprintf("%s", nc.Vars[vkRole]))
 				if err != nil {
-					return err
+					log.Warnf("Unable to load template %s; skipping", tmplN)
+					continue
 				}
-				if tmpl.Lookup(tmplN) == nil {
-					return fmt.Errorf("template not found %s", tmplN)
+				l = tmpl.Lookup(tmplN)
```
hellt, line 80: is this assignment needed, considering that `l` was assigned on L74?

If it is needed, then assignment on L74 is not?

LimeHat: This one is needed, the logic as follows: 1) check if we have a template in memory for a given node type X (L74); 2) if not - try to load it; (L75) 3) check again (the result) (L80-82). So if 2) gives us non-nil result, we'll be proceeding further, otherwise skipping an iteration based on check 3). We can eliminate the first assignment indeed, and I've considered it, but it will bring some extra work for nodes with templates (available templates will be loaded from a disk for each node vs once per kind). I don't have a strong preference here so kept the current logic with more code / fewer computations. Let me know if you want to change this.

hellt: okay, lets take it one step at a time

PR #678, opened 2021-11-05
Async post-deploy actions for sonic

networkop:
This change is to allow Sonic to boot asynchronously instead of blocking the post-deploy action. The `supervisord` can take quite a bit of time to boot and, when if it fails, it doesn't return, instead it keeps blocking which blocks the `containerlab deploy` action forever. 

Another minor change is the way bgpd is started. Instead of calling `bgpd` directly, we're using `supervisord` to start this process. 

Tested locally, seems to be working fine for me


hellt:
thanks @networkop !

PR #679, opened 2021-11-07
add the option to set nodes memory, CPUs and CPU sets

karimra:


hellt:
is it safe to move it from draft to ready?

hellt:
thanks!

PR #702, opened 2021-11-27
delta-config: Custom CLI config per node, as delta on top of release default

jbemmel:
* Also includes reflect-based generic Node property retrieval method (TODO: refactor others)
* Also auto-enables connected interfaces


hellt:
thanks for the ideas, @jbemmel 

CLI config is enabled in #708 
auto-up of interfaces is added in #709 

the reflection part, as discussed, is not something I am comfortable to take in just yet

PR #706, opened 2021-11-29
Updated endpoint list schema to match the node name schema

### LINKED ISSUE #705: Fix clab.schema.json to allow for periods in device names
Currently, if a device name includes a period in the name, when adding links to a clab.yml file, the schema does not recognize the period and will render an error even though the period should be allowed.

Issue is with the Regex here: https://github.com/srl-labs/containerlab/blob/main/schemas/clab.schema.json#L241

Steps to reproduce:
1. Open VSCode
2. Create lab.clab.yml file
3. Under links, add `- endpoints: ["MY.DEVICE:eth1", "MY-DEVICE2:eth1"]`

nlgotz:
Fixes #705 

This uses the same naming pattern as the nodes (line 301)

### DIFF
--- a/schemas/clab.schema.json
+++ b/schemas/clab.schema.json
@@ -238,7 +238,7 @@
                     "minItems": 2,
                     "items": {
                         "type": "string",
-                        "pattern": "^[\\w\\s-/]+:[\\w\\s-/]+$"
+                        "pattern": "^.*:[\\w\\s-/]+$"
                     },
                     "uniqueItems": true
                 }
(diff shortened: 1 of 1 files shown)


schemas/clab.schema.json
```diff
@@ -238,7 +238,7 @@
                     "minItems": 2,
                     "items": {
                         "type": "string",
-                        "pattern": "^[\\w\\s-/]+:[\\w\\s-/]+$"
```
hellt, line 241: let's maybe use the following 

```
^\\S+:\\S+$
```
which allows any non whitespace char for node and link portions

your current example will not highlight a missing node name, which is not great

nlgotz: Yeah, that looks better than what I tried. I didn't think to check not having a node name

hellt:
thanks @nlgotz !
Changes should be live already

PR #708, opened 2021-12-01
SR Linux `startup-config` in CLI format

hellt:
This PR makes it possible to provide the startup-config for SR Linux in CLI format.

The same `startup-config` directive is used to provide a path to a file with the following contents:

```
set / system information location test123
```

In that case, containerlab will recognize that the file content is not a JSON object and will not mount this file as the `config.json` file.
Instead, it will apply the instructions in that file during the post-deploy stage via `sr_cli`.

In summary, this is the proposed way to provide overlay configurations to SR Linux nodes in CLI format that will get applied over the default config.

### DIFF
--- a/docs/manual/kinds/srl.md
+++ b/docs/manual/kinds/srl.md
@@ -3,7 +3,7 @@
 [Nokia SR Linux](https://www.nokia.com/networks/products/service-router-linux-NOS/) NOS is identified with `srl` kind in the [topology file](../topo-def-file.md). A kind defines a supported feature set and a startup procedure of a node.
 
 ## Managing SR Linux nodes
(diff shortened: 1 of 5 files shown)


hellt:
@jbemmel I propose to reuse the startup-config directive instead of creating another node property as you proposed in #702

As to the reflect-based property retrieval I am hesitating to accept this. I try to avoid reflection whenever possible to keep the compile time type cheking. @karimra may be able to provide his more experienced view on that.

Enabling interfaces defined in the topo file is a good idea, I will likely add it as another PR.

hellt:
Yes, that is the tradeoff I can't make peace with. Go is explicit and wordy (compared to python) in that sense, but that is not without its benefits.

So let's keep it the idiomatic for the time being

PR #711, opened 2021-12-02
Map kernel module path to cvx container

jbemmel:


hellt:
Hi @networkop is it safe to do that in all cases?

networkop:
only for the docker runtime. when running in ignite runtime, the ignite kernel image ships its own `/lib/modules` with stuff needed for CL.

PR #713, opened 2021-12-03
Use deterministic MACs for srl nodes

jbemmel:
By including the node index in the MAC address that starts with "1a:b0:" (LAB, first bit must be clear)

This allows for much easier reading of MAC tables; SRL will fill in the last 3 octets using FF:00:\<port\>


hellt:
thanks! makes lots of sense

PR #726, opened 2021-12-20
Add IXR d2l and d3l device types

jbemmel:


nodes/srl/topology/7220IXRD3L.yml
```diff
@@ -0,0 +1,13 @@
+# Copyright 2020 Nokia
+# Licensed under the BSD 3-Clause License.
+# SPDX-License-Identifier: BSD-3-Clause
+
+chassis_configuration:
+  "chassis_type"  : 73
```
hellt, line 6: Yaml formatting is broken here

hellt: fixed in 0537e25

Issue #668, opened 2021-10-25
swap embedded ssh transport with scrapligo


hellt:
to add support for most popular network oses

steiler:
@hellt what's the actual status?

hellt:
@steiler it is available for implementation.
although I'd say there is a first step that is more crucial to implement, that goes in parallel to what this issue describes.

We need to make sure clab has an option to configure the device(s) with the CLI commands upon the lab deployment.
In essence, this means that a user should be able to provide a file with CLI commands, or a list of such commands in a clab file and those commands need to be "pushed" to the relevant nodes as if they were typed in by an operator.
For this we need to leverage `scrapligo`.

Currently, we have a `.topology.nodes[node].config` structure that was designed to do complicated config templation stuff with clever things embedded, but I feel like it is far more important right now to deliver on the explained use case to make sure we have an easy way to provision network nodes with a config.

I was thinking of having a `clab config push` command that does that, as well as this functionality, can be reused to allow users to create an "over the top" config file that is applied once the lab is deployed automatically.
Think of it as a config that users want to add on top of the clab default config

kellerza:
This should be very close to what we need in containerlab - https://github.com/labctl/labctl/tree/main/utils/tx - it uses scrapligo where the standard driver reads .ssh/config for more advanced configurations (see [here](https://labctl.net/guide/config-engine-tx#transport))

It implements your idea of "config push" as "config send" and on top of that "config commit" (similar to what exists of config engine in containerlab) and "config compare" for SR OS and SRL kinds. You can also override these KindMaps for other kinds

PR #734, opened 2021-12-25
swap poc ssh transport with scrapligo

hellt:

PR #737, opened 2021-12-28
Scrapligo as transport

iprouteget:
This aims to replace the embedded SSH transport used for the `config` command with scrapligo's various network drivers (https://github.com/srl-labs/containerlab/issues/668).

This might have been more overkill, but I thought it would be neat to be able to set various driver options directly in the yaml file (e.g. username, password, ssh config, private key file, etc). At the same time, I wanted to make sure only certain options (e.g. most everything [here](https://github.com/scrapli/scrapligo/blob/0669a885e9a80c52f89fff2d35a10baa5fce1eab/driver/base/base.go#L27)) were available.

I still need to test among the various platforms, but this seems to work on crpd so far:

```
topology:
  kinds:
    crpd:
      enforce-startup-config: true
      startup-config: ./startup.conf
      driver:
        ssh-config-file: <path to my ssh config>
        secondary: clab123
        port: 22
```

I was able to then use the following command to push generated configs from templates using:

`sudo ~/test/containerlab/bin/containerlab config -t junos-routed-fabric.clab.yml template -p ./templates/ -l net-cfg`

This is my first ever PR to an open source project, so I welcome any and all scrutiny!


hellt:
Hey @iprouteget 
Awesome initiative! Without looking at the code yet I have a few suggestions

Let's move what you named a 'driver' config under the 'config.transport.scrapli' struct. 
The idea is to extend the existing config structure that currently has just Vars field with a transport container that will have configurations for various supported transports. Scrapli is one of them, but gnmi or scrapli-netconf will be the others

iprouteget:
Thanks @hellt! It's funny, I actually tried doing something similar the first time. Just to be clear of your plan, the file would look something like (as an example):

```
  nodes:
    sp1:
      kind: crpd
      image: crpd:21.3R1.9
      config:
        vars:
          var1: <value>
          var2: <value>
        transport:
          scrapli:
            ssh-config-file: <file path>
            username: <username>
          gnmi:
```
Which could be populated by at the node level or a higher up (e.g. Kinds) level?

hellt:
Yes, exactly like you showed in the snippet.
The config structure follows the usual descendency chain: defaults -> kind
-> node

Meaning that for example you could have defined a transport for the whole
crpd kind and it would have been inherited by all the crpd nodes.

The most specific config wins

iprouteget:
Sounds good. This latest push does the following:

- Create a struct `ConfigTransport` that includes any and all transports that could be available, starting with Scrapli for now. This is moved within the `ConfigDispatcher` struct so we can unmarshal within the `config` higher level key.
- Moves `transport/driver.go` file to `transport.go`. I think my intent is to allow containerlab to parse all specified transports (scrapli, gnmi, etc) and create the driver needed.
- Updates the json schema to include scrapli and available options.

I mentioned in a comment about setting default values when parsing the yaml file. I don't think the way I did it for scrapli's default port is best; I thought setting a struct tag would work, but I am probably missing something easy here.

kellerza:
I'm not convinced the Transport Interface is the correct solution here, or put differently, it might not be complete in its current form: 

While we can adapt the transport mechanism to connect to the router, we lack any form of transactions (send, commit, compare), and these transactions can be implemented completely differently based on the transport provider.

Taking the existing embedded ssh implementation as an example. It follows Connect, Session_start, Write, Session_end and then Close. Maybe session_start/end can use Transport's Write method (but a generic write is not flexible enough), and in some implementations there is already some transactional functionality built in (at least when I looked at Scrapligo in the past). The transaction and write according to me is too closely coupled to try separate them. I also had a use case to use custom ports, which required influencing the connect. 

For all these cases I simply started using a method to send the config

```go
const (
	ActionSend    = "send"
	ActionCompare = "compare"
	ActionCommit  = "commit"
)

func ConfigSend(c *config.NodeConfig, action string) ([]*transport.SSHReply, error) {
```

hellt:
@kellerza 
I think the transactionality (if that is offered by the transport) will be configured through the transport options. 

Let's take the scrapligo example and the use cases our users might need

### 1 Configuring the device by sending commands
The most common use case, the goal is to get a device to a certain configuration state by sending and applying config commands as if they were entered over the CLI.

Scrapligo offers the following method - [SendConfigs](https://github.com/scrapli/scrapligo/blob/main/driver/network/sendconfigs.go#L14) - which takes a slice of strings representing commands. This method will (in the background) acquire the **configuration level** privilege that is relevant to the specified NOS (like `edit-config exclusive` for SR OS) and will deliver the commands over the SSH channel.

This method does not imply any particular `commit` behaviour, as such users will have to add to the end of their configuration snippet something like `commit now` or similar, depending on what kind of commit sequence is relevant.

### 2 Sending commands which do not require configuration context. Aka `show` commands
In scrapligo this would be carried out with [SendCommans](https://github.com/scrapli/scrapligo/blob/main/driver/network/sendcommands.go#L10) which also takes a slice of commands to send over to the device, but this time the **execution level** privilege is acquired if needed to send `show` commands (think of `enable` on cisco platforms).

Now back to scrapligo transport interface implementation. In my view the default mode for scrapligo transport should be `SendConfigs`. If a user wants to send `show` commands than it will set the relevant transport option. I think what you did was to deduce if a file contains `show` commands by looking at the file template name.

kellerza:
> Now back to scrapligo transport interface implementation. In my view the default mode for scrapligo transport should be SendConfigs. 

This sound more like we need multiple Transport implementations, one for SendConfigs and one for SendCommands.
Otherwise the action (send/commit/compare) is part of the transport options, which seems a bit counter intuitive, since its options for a higher layer, the transaction, rather than the transport.

> If a user wants to send show commands than it will set the relevant transport option. I think what you did was to deduce if a file contains show commands by looking at the file template name

Indeed in the very first iteration, but we have CLI to select the action since a while already (see below)

```
$ clab config --help

configure a lab based on templates and variables from the topology definition file
reference: https://containerlab.srlinux.dev/cmd/config/

Usage:
  containerlab config [flags]
  containerlab config [command]

Aliases:
  config, conf

Available Commands:
  compare     compare configuration to a running lab
  send        send raw configuration to a lab
  template    render a template

Flags:
  -p, --template-path strings   comma separated list of paths to search for templates
  -f, --filter strings          comma separated list of nodes to include
  -h, --help                    help for config
```

btw, in my view even `commit`/SendConfigs should not be default, but rather through an explicit **commit** on the CLI. Once you work with compare & commit during template development, this is a big usability enhancement and prevents you from accidentally comitting.

iprouteget:
Hi all, just catching up here.

I'm still working on merging the transport structs for default/kind/node, but I wanted to submit the first item regarding the move to the Transport interface.

This cleaned up send.go quite a bit. Based on these more recent notes, perhaps I need to look at adding some transport options to incorporate a broader range of functions instead of only committing the configuration. Currently, I'm just using scrapligo's cfg driver, which seems to work for pushing the configurations.

iprouteget:
Phew, it's been a hot minute since I've been able to look at this again. From the notes provided earlier, I tried moving the scrapli code to use the existing transport interface, which seems to be working so far. @hellt, regarding the merging of structs, I just copied the example you provided in discord since it was lightweight and straightforward. I figured once there are more complex types, we'd start testing more mergo?

I still see some things in my code that probably need refactoring, but I want to make sure I'm on the correct path regarding the two main items mentioned.

hellt:
Hey @iprouteget 
thanks for getting back to it
your hot minute was just in time, as we had different things to squash in the meantime.
I will review that in coming weeks!

iprouteget:
Awesome! Much appreciated. I'll add some tests this week as well.

hellt:
@iprouteget do you mind checking if you can port some cleanup changes I started to do in #734 with regards to config engine cleanup?

Then I could retire that PR

iprouteget:
Will do!

iprouteget:
Okay I believe I have the cleanup items from #734, and I also added a test for the MergeStructConfigs function.

hellt:
Hi @iprouteget 
sorry this still drags... There was some house keeping work done that is related to that PR as well

1. scrapilgo was updated to v1, and this has to be addressed here. Here is a PR that updated other parts of containerlab with scrapligo update https://github.com/srl-labs/containerlab/pull/925

iprouteget:
Hi @hellt, all good! I'll update this accordingly.

iprouteget:
@hellt with the amount of time that has passed and the updates to scrapligo, I feel like it would make more sense to close this out and start fresh :) It might allow for breaking up some parts into smaller PRs. Let me know if this works best. Thanks!

hellt:
That makes a alot of sense to me indeed

Issue #739, opened 2022-01-13
rework sonic integration to use two containers


hellt:
It seems sonic-vs container image has been stripped in late 2020 from basic linux utilities such as `iproute2` and a few others. This change led to our current support for sonic-vs to render unusable with recent versions of sonic-vs.

This was identified by @antongisli https://github.com/srl-labs/containerlab/pull/545#issuecomment-1002110724

The initial sonic-vs support in containerlab was not using a pair of containers, as the sonic docs [suggests](https://github.com/Azure/sonic-buildimage/blob/master/platform/vs/README.vsdocker.md). Instead we used a simplified one-container setup with veth pairs representing front ports. This seems to be not possible anymore, and we rather need to fall back to the two-container's setup.

/cc @bortok @antongisli

hellt:
parking this for a while, since single-container mode still works apparently

PR #740, opened 2022-01-13
updated missing info in the sonic lab documentation.

winstonliu1111:
I updated the doc for the current srl-sonic lab to cover missing steps to complete the ping between the loopback addresses.

### DIFF
--- a/docs/lab-examples/srl-sonic.md
+++ b/docs/lab-examples/srl-sonic.md
@@ -24,7 +24,7 @@ This lab demonstrates a simple iBGP peering scenario between Nokia SR Linux and 
 Once the lab is deployed with containerlab, use the following configuration instructions to make interfaces configuration and enable BGP on both nodes.
 
 === "srl"
-    Get into SR Linux CLI with `docker exec -it clab-srlsonic01-srl sr_cli` and start configuration
+    Get into SR Linux CLI with `docker exec -it clab-sonic01-srl sr_cli` and start configuration
     ```bash
     # enter candidate datastore
     enter candidate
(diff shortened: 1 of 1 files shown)


hellt:
thanks @winstonliu1111! I overlooked a lot of it it seems :D

hellt:
which sonic-vs image did you use btw? because it seems the recent ones require a little bit different combination of containers - #739

winstonliu1111:
@hellt. i didn't want to deal with the sonic build, so I tried it against  hairlessalpaca/docker-sonic-vs:202012, as well as an image from @bortok based on head of master from 1/11/2022.  I think @bortok is looking at building another image based on the 202111 tag.  the lab worked okay.

the link you refer to in #739 seems to be just an example on how they use it (I don't think that is strictly required).  they were doing 2 fake servers via netns though they only launched a single server container.  the one end of the veth pairs that sonic sees is in the host network namespace of the server container.  it is compact, but makes it convoluted for people to understand.

bortok:
@winstonliu-keysight could you please change `/etc/frr/daemon` to `/etc/frr/daemons`? I check with my image from 1/11/2022 and it works otherwise.

hellt:
Hi @bortok 
Did you build that image from 2022 or did you fetch it from azure artifacts?
In any case would you mind pushing this image somewhere I can pull it from and test?

@antongisli was trying to use the recent images and it was not working...

bortok:
Hi @hellt 

I've built it from master on 1/11/2022 using @antongisli repo. Just uploaded on dockerhub under `netreplica/docker-sonic-vs:20220111`. I've also tried checking out `202111` branch but it fails to build so far.

hellt:
@antongisli do you want to try this out?

antongisli:
sure I'll try it. BTW I read somewhere, and seem to have found, that BGP can be configured with SONiC config commands rather than using vtysh: [link](https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md#bgp)

antongisli:
`A:srl# ping 192.168.1.2 network-instance default
Using network instance default
PING 192.168.1.2 (192.168.1.2) 56(84) bytes of data.
64 bytes from 192.168.1.2: icmp_seq=1 ttl=64 time=8.86 ms
64 bytes from 192.168.1.2: icmp_seq=2 ttl=64 time=6.51 ms
^C
--- 192.168.1.2 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1002ms
rtt min/avg/max/mdev = 6.513/7.688/8.864/1.178 ms
`

hellt:
@antongisli if you refer to the docs on the web site it will get build once a new release comes out. So far the changes are only on the main branch

PR #755, opened 2022-01-28
Use Docker authentication when needed

### LINKED ISSUE #741: use existing authentication parameters for container registries with docker runtime
When containerlab tries to pull an image, it doesn't use any authentication parameters. As such, if users chose a registry that is not accessible without authentication, containerlab will fail to use such registries.

A proposed enhancement is as follows:

1. in [`PullImageIfRequired`](https://github.com/srl-labs/containerlab/blob/bf3f30b7a055a16c48ce50111dadd9cf8904d66a/runtime/docker/docker.go#L380) func the code determine which registry URL is used first. This might not be trivial, as the default docker hub registry for example is `https://index.docker.io/v1/`
2. then we need to read the JSON file `~/.docker/config.json`
3. then we need to take the auth string from that file that corresponds to a registry URL we identified in 1
4. then we use this auth string in the pullimageoptions as shown here https://docs.docker.com/engine/api/sdk/examples/#pull-an-image-with-authentication

A good first step is to check how docker cli does all of that - https://github.com/docker/cli/blob/6e2838e18645e06f3e4b6c5143898ccc44063e3b/cli/command/image/pull.go#L51 can be that we may reuse whatever docker does there

lbaker-esnet:
This PR enables authenticated image pulls when using the Docker runtime by leveraging an existing Docker configuration file. The authentication mechanism used in this PR assumes that the docker credential store is `$HOME/.docker/config.json`. Closes #741 

**Scenarios where registry credentials are used**:
* If an image must be pulled from a registry, the image domain is used to locate authentication credentials in the docker config. If a match is found the credentials are included in the image pull request.

**Scenarios where registry credentials are not used**:
* If an image must be pulled from a registry and a matching authentication credential is not found in the docker config, credentials will not be included in the image pull request.
* If the docker configuration file is missing, unreadable, or is invalid then the credentials will not be included.
* If the image domain does not match a registry defined in the config file, then the credentials will not be included.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/containernetworking/plugins v0.9.1
 	github.com/containers/podman/v3 v3.4.4
 	github.com/digitalocean/go-openvswitch v0.0.0-20201214180534-ce0f183468d8
+	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v20.10.11+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.4.0
@@ -92,7 +93,6 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/disiqueira/gotree/v3 v3.0.2 // indirect
 	github.com/docker/cli v0.0.0-20200130152716-5d0cf8839492 // indirect
-	github.com/docker/distribution v2.7.1+incompatible // indirect
 	github.com/docker/docker-credential-helpers v0.6.4 // indirect
 	github.com/docker/go-events v0.0.0-20190806004212-e31b211e4f1c // indirect
 	github.com/docker/go-metrics v0.0.1 // indirect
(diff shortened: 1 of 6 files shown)


lbaker-esnet:
@hellt  The code in this PR will use the default docker config file for authentication credentials. Do you see the need to add a configuration variable, environment variable, or command line argument to allow the user to specify a different configuration file?

hellt:
thanks @lbaker-esnet 
I think the default config file location is good. we will see if we (ever) need to configure the location via flags/env vars/etc

I will get to this PR this week

hellt:
Hi @lbaker-esnet 
first of all thank you for working on this! Great aid and a solid improvement

I have provided a few edits on top of you last commit, you can check it out and see if you agree with them.

Mainly I wanted to streamline happy path to be on the first level of indentation, thus refactored the way nesting blocks are used.
Then I used more of table driven tests for GetDockerAuth. So all in all not so many drastic changes :)

lbaker-esnet:
@hellt Excellent, thank you for looking at this. The edits look great (and have provided me some insight into styling for the future).

PR #778, opened 2022-02-14
Adding public keys in ~/.ssh/authorized_keys 

hansthienpondt:
1/ adding ~/.ssh/authorized_keys as valid public keys for the container.
2/ renaming for clarification/ambiguity:
     authzKeysFPath to clabAuthzKeysFPath
     authzFName to clabAuthzFName

### DIFF
--- a/clab/authz_keys.go
+++ b/clab/authz_keys.go
@@ -15,14 +15,14 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


hellt:
thanks @hansthienpondt !

PR #785, opened 2022-02-17
Shared netns support aka network-mode: container 

LimeHat:
As discussed previously, MVP for shared net namespaces. I took a naïve approach here and reused the existing network-mode construct.


hellt:
What would be cool is to have a look at how docker compose implemented
"depends-on" instructions. I suppose it is a waiting loop for running
status of a container.
Maybe we should add this in this (or subsequent PR) to eliminate startup
delay

LimeHat:
Indeed, I thought about that (implementing startup/shutdown control logic); I'd rather do it as a separate PR

PR #787, opened 2022-02-19
add interface mapping file option for cEOS

### LINKED ISSUE #389: Add option to specify the name of the management interface for cEOS
On physical Arista EOS switches, the default management interface is called 'Management1'. cEOS containers are generated with 'Management0' as the management interface name. I'd like to be able to specify this name (or switch it to Management1 in some other way) so that I can run my ansible playbooks for my physical (production) switches on my test containers in containerlab as well without modifying them.

I assume this would be a feature request, but if there's already a way to do this that I missed I'd be more than happy to hear so.

### LINKED ISSUE #786: cEOS - add an option to provide a file for the interface layout
Starting with 4.28.0F there is a possibility to define a mapping between the container ethernet interfaces and the EOS interfaces. With this it is possible to define your own interface layout for the cEOS. Additionally you can decide if the cEOS should use Management0 or Management1. There should be an option in the topology to define which file to be used for a ceos node, similar to the startup-config. 

Example format of a mapping file (EosIntfMapping.json):

```
{
  "ManagementIntf": {
    "eth0": "Management1"
  },
  "EthernetIntf": {
    "eth1": "Ethernet53/1",
    "eth2": "Ethernet54/1",
    "eth3": "Ethernet49/1",
    "eth4": "Ethernet50/1"
  }
}
```

bjmeuer:
close #786 
close #389


hellt:
Hi @bjmeuer 
we try to keep the node parameters generic, meaning that they should be applicable to most nodes. The vast majority of the node level parameters are docker-cli inspired ones. This allows us to have a consistent API surface for nodes definitions.

Keeping that in mind I propose to not introduce a new node parameter `intf-mapping` but instead re-use `binds` parameter that will allow you to map a mapping file to a well known location inside ceos container.

The rest of the logic will check if a bind mount targeting the interface mapping file exists and do the needful

hellt:
Alternatively you can put the `inft-mapping` property under the `extras` field that we have defined for node parameters - https://github.com/srl-labs/containerlab/blob/e81ea32279296c4ed3cb808b1d42c197a8883019/types/node_definition.go#L59

https://github.com/srl-labs/containerlab/blob/e81ea32279296c4ed3cb808b1d42c197a8883019/types/types.go#L240

The extras field can contain extra parameters which are used by some nodes, but not for all of them.

bjmeuer:
> Alternatively you can put the `inft-mapping` property under the `extras` field that we have defined for node parameters -
> 
> https://github.com/srl-labs/containerlab/blob/e81ea32279296c4ed3cb808b1d42c197a8883019/types/node_definition.go#L59
> 
> https://github.com/srl-labs/containerlab/blob/e81ea32279296c4ed3cb808b1d42c197a8883019/types/types.go#L240-L243
> 
> The extras field can contain extra parameters which are used by some nodes, but not for all of them.

Thanks for your suggestions, I will have a look into the bind option first. I will adjust my code and resubmit it.

bjmeuer:
Hi @hellt 

I updated the code using the binds option as you suggested. This works fine. 

Additionally I updated the sections which did statically use "Managment0" as cEOS management interface. If nothing is defined it still uses "Management0", but if in the binds a "EosIntfMapping.json" is mounted, the interface mapped to eth0 in this file will be used. 

Besides that I set ETBA to "1". I talked to a developer and he mentioned that valid values are "0" or "1" and it seems that this is now verified when this Interface Mapping is used. With the current setting of "ETBA=4" the cEOS interface don't work.

hellt:
thanks @bjmeuer I will have a look

@burnyd from Arista suggested changing ETBA to 4 a while back (https://github.com/srl-labs/containerlab/pull/360), with a reasoning that it enables some features that were not possible with ETBA=1 value.
But if `1` is now a more preferable nowadays, thats ok

bjmeuer:
> thanks @bjmeuer I will have a look
> 
> @burnyd from Arista suggested changing ETBA to 4 a while back (#360), with a reasoning that it enables some features that were not possible with ETBA=1 value. But if `1` is now a more preferable nowadays, thats ok

I will talk to burnyd where this is coming from

nodes/ceos/ceos.go
```diff
@@ -118,6 +124,42 @@ func createCEOSFiles(node *types.NodeConfig) error {
 	cfg := filepath.Join(node.LabDir, "flash", "startup-config")
 	node.ResStartupConfig = cfg
 
+	// use interface mapping file to set the Management interface if it is provided in the binds section
+	// default is Management0
+	mgmtInterface = "Management0"
+	for _, bindelement := range node.Binds {
```
hellt, line 130: suggest taking this out as a private function

nodes/ceos/ceos.go
```diff
@@ -118,6 +124,42 @@ func createCEOSFiles(node *types.NodeConfig) error {
 	cfg := filepath.Join(node.LabDir, "flash", "startup-config")
 	node.ResStartupConfig = cfg
 
+	// use interface mapping file to set the Management interface if it is provided in the binds section
+	// default is Management0
+	mgmtInterface = "Management0"
```
hellt, line 129: the default var value has already been set on Ln 45

nodes/ceos/ceos.go
```diff
@@ -118,6 +124,42 @@ func createCEOSFiles(node *types.NodeConfig) error {
...
+				}
+				intfMapping = string(m)
+
+				// Reset management interface if defined in the intfMapping file
+				var intfMappingJson map[string]interface{}
+				err3 := json.Unmarshal([]byte(intfMapping), &intfMappingJson)
+				if err3 != nil {
+					return err3
+				}
+				managementIntfMap, ok := intfMappingJson["ManagementIntf"].(map[string]interface{})
```
hellt, line 146: I think instead of doing unmarshalling to an empty interface you can go safe way and introduce a struct

```go
type IntfMap struct {
	ManagementIntf struct {
		Eth0 string `json:"eth0"`
	} `json:"ManagementIntf"`
	EthernetIntf struct {
		eth map[string]string
	} `json:"EthernetIntf"`
}
```
and use it for unmarshalling

fenner:
> @burnyd from Arista suggested changing ETBA to 4 a while back (#360), with a reasoning that it enables some features that were not possible with ETBA=1 value. But if `1` is now a more preferable nowadays, thats ok

I'm the Arista person in charge of cEOS-lab, and I can say that `1` is the correct value (and that nothing that I'm aware of ever looked for any other value.)

hellt:
> I'm the Arista person in charge of cEOS-lab, and I can say that `1` is the correct value (and that nothing that I'm aware of ever looked for any other value.)

great, thanks. then `1` it is

nodes/ceos/ceos.go
bjmeuer: @hellt Thanks for the comments. I updated my code accordingly.

hellt:
@sulrich @bjmeuer colleagues, can you share the 4.28 version so that I can test this PR?
you can reach directly via discord

hellt:
@bjmeuer I have made slight modifications to the logic you added in this PR
since I don't have 4.28, please run a few tests to check if it works both with and without a mapping file. Thanks!

PR #788, opened 2022-02-20
create `iptables` forwarding rule targeting containerlab nodes

### LINKED ISSUE #781: add iptables `allow` rule for containerlab nodes in the forwarding chain
It is often required to have external connectivity to the containerlab nodes. Most common use cases is to enable external mgmt system to reach containers launched with containerlab.

The default iptables rules on a docker host have a default DENY rule in the FORWARD chain:

```
❯ iptables -vnL | grep -i forward | head -1
Chain FORWARD (policy DROP 0 packets, 0 bytes)
``` 

this blocks external connectivity to the containers and is an expected docker behaviour - https://docs.docker.com/network/bridge/#enable-forwarding-from-docker-containers-to-the-outside-world

The proposal is to add the ALLOW rule for the bridge network that backs up the docker network. For instance:

```
sudo iptables -I DOCKER-USER -o br-73aa9d6ee745 -j ACCEPT
``` 

This will allow any incoming traffic destined to the containerlab management network.

When we destroy a lab, the rule must be deleted only if there are no other links attached to the bridge interface.

ref #780

hellt:
fix #781 

iptables rules management have been added to docker runtime only so far.

Note, apart from the changes done in `iptables.go` file where the funcs to add/remove iptables rules are, the following changes have been made:

1. deploy and destroy cmd functions have been scoped out to `deployFn` and `destroyFn` accordingly.
2. `Mgmt()` method has been added to `Runtime` interface to allow pulling the management struct
3. Runtime's struct `Mgmt` field was changed to  `mgmt`


hellt:
@LimeHat when you create a management network in podman you don't save the bridge interface under c.mgmt.Bridge like we do for docker - https://github.com/srl-labs/containerlab/blob/iptables-fwd/runtime/docker/docker.go#L204?

Is it feasible to add this step so that this PR can have podman support? The idea here is to automatically add iptables rule allowing external traffic to reach the containerlab nodes. This is achieved by installing an allow rule in the FORWARD chain.

hellt:
Let me merge this one, and podman we can address in a separate PR. At this moment the iptables management will only happen if the global runtime is set to docker

PR #789, opened 2022-02-21
Refactor CreateContainer runtime method

### LINKED ISSUE #428: Refactor CreateContainer to not start the container straight away
CreateContainer in docker runtime also starts the container and creates the namespace links.
StartContainer should be called separately, by the core library not implicitly after a create. 

https://github.com/srl-labs/containerlab/blob/4e4df293bd70e4b05cab8b938895e48ace777c37/runtime/docker/docker.go#L251-L261

LimeHat:
Closes #428. 
This PR introduces a few behind-the-scenes changes according to the abovementioned discussion:

1. CreateContainer method will not start containers automatically
- Implemented for Docker and Podman runtimes currently
- Ignite & containerd still use a single function call (either create or start) which performs both actions. Refactoring is doable, but just a bit trickier as I'm not very familiar with the implementations.
2. Deploy method in node definitions now calls CreateContainer() first and then StartContainer()
3. Previous CreateContainer method returned an empty interface{} -> i moved this to the Start method for now (used only by Ignite/cumulus as of today); but we can add it to the new Create if needed.
4. Miscellaneous things like setting up netns symlinks and disabling eth offloads are moved into a separate private method (postStartActions) which is called by StartContainer method [in docker and podman].

The main idea is to get better code readability & manageability and give us more flexibility in the future for cases where we may want to introduce some additional actions between container initialization & start phases.

[a couple of very minor changes are also there like changing %s to %q for quoted strings in a few places & param. naming normalization for docker (containerID & id -> cID)]


hellt:
Hi @LimeHat 
I have just recently merged #788 where I had to change a few small runtime things along the way...
It seems this made a conflict in `docker.go`

hellt:
LGTM, thanks

Issue #649, opened 2021-10-12
propagate config vars to ansible inventory yaml file


hellt:
if a defaults/kinds/node section has a `.config.vars` section it makes sense to copy it to the ansible invenotory, so that ansible scripts could use those vars 

/cc @jrokeach

tobbbles:
🙏 Number 1 use case in my mind for this is to provide a `ansible_docker_host` variable with the container name, to be used with the [Ansible docker connection plugin](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_connection.html#ansible-collections-community-docker-docker-connection)

Actually if this is something that were automatically generated by containerlab, that'd be even better. 

Thanks so much for the great tool!

hellt:
Hi @tobbbles 
If I understand the plugin purpose correctly (run commands inside containers) then I think you don't need this special variable set. The current inventory hostnames are the fully qualified container names, so you should be able to run docker commands with them

tobbbles:
Visiting that now, that's true, however for the hostname to work we actually need to _remove_ the `ansible_host` var, as here https://github.com/srl-labs/containerlab/pulls?q=is%3Apr+ansible+#issuecomment-1047270956

Would we want something to extend this config section of a node to inform containerlab to _not_ generate teh ansible host, and instead leave just the inventory_hostname to take precedent?

hellt:
Config functionality is not maintained at the moment. If/when this is revised with scrapli ssh engine we will go revisit it

PR #793, opened 2022-02-21
Ansible docker connection inventory compatibility

tobbbles:
As per my comment https://github.com/srl-labs/containerlab/issues/649#issuecomment-1047252521 I realized I could just, implement this myself 👯 

Currently `ansible_docker_host` is added as a var to all nodes of `kind: linux`, this is a naive assumption and could probably be more of an opt-in option, but let's see how the PR plays out.

Test case updated, and I _think_ I've managed to document the behaviour in `docs/`, too.

### DIFF
--- a/docs/manual/inventory.md
+++ b/docs/manual/inventory.md
@@ -46,9 +46,28 @@ Lab nodes are grouped under their kinds in the inventory so that the users can s
               ansible_host: <mgmt-ipv4-address>
     ```
 
-When a lab node is of `kind: linux`, container additionally generates a `ansible_docker_host` variable for the inventory. This is useful when using the [Ansible Docker connection](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_connection.html) plugin.
+If you want to use the [Ansible Docker connection](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_connection.html) plugin, you can set the `ansible-no-host-var` label on a node to not generate the `ansible_host` variable in the inventory. This is useful as the connection plugin will now use the `inventory_hostname` when executing via Docker.
 
-
+=== "topology file"
+    ```yaml
+    name: ansible
+    topology:
+    defaults:
+      labels:
+        ansible-no-host-var: "true"
+    nodes:
+      node1:
+      node2:
+    ```
+=== "generated ansible inventory"
+    ```yaml
+    all:
+      children:
+        linux:
+          hosts:
+            clab-ansible-node1:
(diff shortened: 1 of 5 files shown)


tobbbles:
Hrm. Looks like `ansible_host` takes precedent over `ansible_docker_host`, and removing `ansible_host` for `kind: linux` nodes is a bad breaking change.

My main use case is to use the Ansible docker connection plugin to run playbooks against containers running locally. To achieve this, the container name is required instead of an IP. 

Any ideas on how to implement this? Proper implementation with the conditional based on a nodes' configuration variable?

hellt:
Hi @tobbbles 

you can try and add the following logic to remove the ansible_host var

add a label `ansible-no-ssh-host` that if set to true will not generate `ansible_ssh_host` setting in the template https://github.com/srl-labs/containerlab/blob/main/clab/inventory.go#L38

tobbbles:
Aaaand implemented!

I'm not 100% comfortable with the logic of `ansible-no-host-var` being burried away in the template generation logic, but it's added, and it's working. What are your thoughts?

hellt:
thanks @tobbbles! 
Yes, I am not a big fan of programming in templating langs, but here it seems there is only one option left -- create some struct field that will store a bool of `with-ansible-host` and based on that render/not the line in the inventory.

Good thing the currently proposed logic is not the complex one, so...

Do you want to add a small test for that here https://github.com/srl-labs/containerlab/blob/main/clab/inventory_test.go ?

hellt:
ah jeez, the test is already there =) Let me check the docs and then I'll set this off to merging

docs/manual/inventory.md
```diff
@@ -46,6 +46,41 @@ Lab nodes are grouped under their kinds in the inventory so that the users can s
...
+=== "ansible docker host inventory file"
+    ``` yaml
+    all:
+      children:
+        linux:
+          hosts:
+            clab-ansible-linux-host:
+              ansible_host: <mgmt-ipv4-address>
+              ansible_docker_host: clab-ansible-linux-host
+    ```
```
hellt, line 82: @tobbbles what does this example show? Can't quite get my head around it ;)

tobbbles: Ah, that's leftover from the previous implementation that I completely forgot to remove 🤦 Cleaned this up now :)

hellt:
Thanks @tobbbles for this neat PR, tests and docs

The 0.24 release will be available later today

PR #798, opened 2022-02-23
Replace error with a warning when iptables fails

### LINKED ISSUE #797: make iptables rules management fail gracefully
Quite often users install docker via a package manager. Sometimes this results in docker to be installed without docker-managed iptables rules in place.

We need not to stop deployment/destroy when iptables errors occur, instead we should emit a warning error and keep moving

LimeHat:
Fix #797


runtime/docker/docker.go
```diff
@@ -234,8 +234,11 @@ func (d *DockerRuntime) postCreateNetActions() (err error) {
 	if err != nil {
 		log.Warnf("failed to disable TX checksum offloading for the %s bridge interface: %v", d.mgmt.Bridge, err)
 	}
-
-	return d.installIPTablesFwdRule()
+	err = d.installIPTablesFwdRule()
```
hellt, line 237: I was thinking that maybe we should change the `install/deleteIPTablesFwdRule` funcs to return both an error and the stderr.
It seems that with the current way we invoke os/exec the error message from iptables is not captures, since stderr is not captured

https://github.com/LimeHat/containerlab/blob/121059d710bbf00d2616e9cad13d0d2d63f5bb29/runtime/docker/iptables.go#L31

For example, I believe that if we would change the command to this:

```
❯ iptables -I FAKECHAIN -o eth0 -j ACCEPT
iptables: No chain/target/match by that name.
```
we wouldn't display the error message, and just report "exited with status 1", which is fine, but doesn't give a hint about the probable cause.

So maybe what makes sense is to use `.CombinedOutput` for the `os.exec/Cmd` call, this will capture the stderr and stdout (ofc we can just capture the stderr, but more code lines required :D)
Then should the error be non nil, we create an error with the combined output provided for the context.


PS. (same will be needed for delete command as well)

hellt:
looks good @LimeHat 

I had 3 messages around it this morning :D all the cases had one thing in common:
docker was installed via yum on RH 7.9 -> no DOCKER, DOCKER-USER chain whatsoever...

Initially I thought to make another check if `DOCKER-USER` chain exists and only then attempt to manipulate the rules. This will definitely mask any errors on installations which miss DOCKER-USER chain. But then maybe it is best to keep these errors present just to make people consider installing docker properly, so that chains will be there?
What do you think?

LimeHat:
Hmm. Maybe add a check on the existence and a separate log message (perhaps with a link to a FAQ on the issue)?

hellt:
sounds good

how about that?

image

kept under `external-access` chapter of `network.md`?

Source:

```
???error "I receive 'Missing Docker iptables chains' error"
    Containerlab will throw an error "Missing Docker iptables chains" when no `DOCKER-USER` chain is found. This source of this error is likely rooted in the way docker was installed on your system. Quite often, it happens when docker is installed via package managers.

    It is recommended to follow the [official installation procedures](https://docs.docker.com/engine/install/) by selecting "Installation per distro" menu option. When docker is correctly installed on the system, additional iptables chains will be available and the containerlab error will disappear.
```

If you want to finish this off you can take this and change the wording as you see fit (along with the additional check in the install/destroy funcs). Otherwise I can finish this off later

Issue #703, opened 2021-11-29
integrate topology viewer based on NextUI


hellt:
There is another topology visualization framework - NextUI - that can build networking topologies in a nice layout

![image](https://user-images.githubusercontent.com/5679861/143847957-cfca9a51-8293-4e48-b3ef-d474727d9b1c.png)

It has been used in a popular netbox plugin - https://github.com/iDebugAll/nextbox-ui-plugin - to visualise topologies based on data extracted from netbox.
This work is based on a prototype built [here](https://github.com/iDebugAll/devnet_marathon_endgame) and explained in details [here](https://habr.com/ru/post/501164/) (Ru lang)

The way this tool works is that it takes a topology definition structure like pictured below and builds an HTML file with NextUI JS powering the toplogy.

```js
var topologyData = {
    "links": [
        {
            "id": 0,
            "source": 7,
            "srcDevice": "edge-sw01.devnet.lab",
            "srcIfName": "Gi0/2",
            "target": 5,
            "tgtDevice": "core-rtr01.devnet.lab",
            "tgtIfName": "Gi0/0/0/1"
        },
        {
            "id": 1,
            "source": 7,
            "srcDevice": "edge-sw01.devnet.lab",
            "srcIfName": "Gi0/3",
            "target": 3,
            "tgtDevice": "core-rtr02.devnet.lab",
            "tgtIfName": "Gi0/0/0/1"
        },
        {
            "id": 2,
            "source": 4,
            "srcDevice": "dist-rtr01.devnet.lab",
            "srcIfName": "Gi3",
            "target": 3,
            "tgtDevice": "core-rtr02.devnet.lab",
            "tgtIfName": "Gi0/0/0/2"
        },
        {
            "id": 3,
            "source": 4,
            "srcDevice": "dist-rtr01.devnet.lab",
            "srcIfName": "Gi4",
            "target": 1,
            "tgtDevice": "dist-sw01.devnet.lab",
            "tgtIfName": "Eth1/3"
        },
        {
            "id": 4,
            "source": 4,
            "srcDevice": "dist-rtr01.devnet.lab",
            "srcIfName": "Gi6",
            "target": 0,
            "tgtDevice": "dist-rtr02.devnet.lab",
            "tgtIfName": "Gi6"
        },
        {
            "id": 5,
            "source": 4,
            "srcDevice": "dist-rtr01.devnet.lab",
            "srcIfName": "Gi5",
            "target": 2,
[...]

PR #802, opened 2022-02-26
Enable serving local static files

ahmedalodhari:
#703 

This PR adds the option to serve local static files as part of the `graph` CLI command. The directory containing the static files is passed via the new `--static-dir` flag. 

This capability simplifies generating network topological by removing the dependency on using an external web server to host the files required by visualization frameworks such as NextUI.

### DIFF
--- a/cmd/graph.go
+++ b/cmd/graph.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"sort"
 	"strings"
+	"os"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/graph.go
```diff
@@ -143,6 +158,24 @@ var graphCmd = &cobra.Command{
...
+        return
+    }
+
+    if stat.IsDir() {
+        return nil, os.ErrNotExist
+    }
+
+    return f, nil
+}
+
```
hellt, line 178: What is the reason for custom implementation of Open method here?

hellt:
Thanks @ahmedalodhari 
I have added nextui static files to this PR.

Do you mind telling me where is the logic that maps the group name to a certain svg icon?

hellt:
Thanks @ahmedalodhari 
merging this

PR #803, opened 2022-02-26
Update ceos.md

JulioPDX:
Missing syntax highlighting in doc


hellt:
it looks like the indentation in this code block to which you added yaml highlighting needs to be fixed as well

JulioPDX (commented):
Please let me know if there is anything else. Thank you

PR #804, opened 2022-02-26
Nextui graph template

hellt:
needs #802 

This PR embeds the Cisco' nextUI library to render the graph in a nicer template


hellt:
@ahmedalodhari can you please tell me where is the mapping happens between the group name and the nextui icon?

Also I wonder what is the proper UI solution for a case when the group is missing to make node tooltip values aligned? Otherwise the empty line is not rendered
image

ahmedalodhari:
@hellt 
The below code tells NextUI to use the value of the `group` configuration attribute, which is passed by the `{data}` object, to render the icon of the node:
```
        nodeConfig: {
            label: 'model.name',
            iconType: 'model.group',
        },
```

So we have two issues when the group attribute is missing: 
1. missing icon, which can be fixed by having default icon type.
2. unaligned node tooltip, which can be fixed by having an empty string.

I'll fix the code and push the updates later today.

hellt:
closed in favour of #813

Issue #815, opened 2022-03-10
Documentation request: cEOS-lab supports cgroupsv2 starting with 4.28.0


fenner:
Hi!  I notice in https://containerlab.dev/manual/kinds/ceos/#cgroups-v1 you mention that cEOS-lab will not start if your container host is using cgroupsv2.  Starting with EOS-4.28.0, cEOS-lab will automatically detect whether your container host is providing cgroupsv1 or cgroupsv2, and will adapt accordingly.  No configuration is needed.

hellt:
Hi @fenner 
thanks for updating on this, indeed we were waiting for 4.28 to come out before updating the docs.

I'll update the docs, unless you would like to come with a short doc PR

fenner:
I looked for a doc repo, and a doc branch, but I forgot to look for a doc subdirectory!  #816

hellt:
Thanks. Addresses as per your PR

PR #816, opened 2022-03-10
cgroups v2 in cEOS-lab starting with EOS-4.28.0F

fenner:
A small documentation update for cgroups v2 in EOS-4.28.0F, per issue #815


hellt:
Thanks!

PR #819, opened 2022-03-12
adds ceos support for extras

sulrich:
## overview

ceos allows some parameters to be overridden or additional options to be set by placing files within the `flash` directory. this PR adds support for a ceos-specific "extra" to copy user provided files into the flash directory on startup. 

additionally, this PR sets the EBTA environment value to be "1" which is the preferred setting.

### DIFF
--- a/docs/manual/kinds/ceos.md
+++ b/docs/manual/kinds/ceos.md
@@ -215,6 +215,22 @@ clab-srlceos01/ceos
 
 9 directories, 11 files
 ```
+## ceos-specific overrides
+
+If there is a need to provide additional ceos-lab specific configuration or overrides to the ceos node in the topology, these can be added using the `ceos-overrides` extra.  These additional files will be copied to the node's flash directory for evaluation on startup.  
+
+```yaml
+name: ceos_lowrider
+topology:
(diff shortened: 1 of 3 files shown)


hellt:
Hi @sulrich 
since this operation seems to be a generic one, could it be named `copy-to-flash` instead of overrides?

sulrich:
@hellt	 happy to change this to 'copy-to-flash' or something similarly generic. 

wrt this vs. binds, it's really kind of a workflow and packaging thing. i tend to use this to inject the extras and then remove the "extras" from the topology file so that i can package this up for storage and restoration without having to worry about binds lurking around. 

upon further reflection, it might make more sense to make this into something that's only used on initial setup.

hellt:
Hi @sulrich check my changes in the latest commit

I have changed the node name to be `ceos-copy-to-flash` to avoid ambiguity and also added a step to resolve relative paths  relative to the topology file and not the CWD. This is the common logic for paths in clab.

When you confirm that it works as you expected I will merge it

sulrich:
builds clean, works correctly.

i'm good.  thanks!

PR #821, opened 2022-03-15
Check for duplicate management IP addresses

### LINKED ISSUE #812: Duplicate Management Address Checks
This is more to protect the user. I made the mistake of copy pasta and tried deploying a lab with duplicate management IPs. Nodes had static assignments. I think it would be good to protect the user and add a duplicate address check before deployment. Thank you!

bclasse:
Resolve #812

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -378,6 +378,9 @@ func (c *CLab) CheckTopologyDefinition(ctx context.Context) error {
 	if err = c.verifyLinks(); err != nil {
 		return err
 	}
+	if err = c.verifyDuplicateAddresses(); err != nil {
+		return err
+	}
 	if err = c.verifyRootNetnsInterfaceUniqueness(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


clab/config.go
```diff
@@ -433,6 +436,35 @@ func (c *CLab) verifyLinks() error {
 	return nil
 }
 
+// verifyDuplicateAddresses checks that every static IP address in the topology is unique
+func (c *CLab) verifyDuplicateAddresses() error {
+	var ipv4_list []string
+	var ipv6_list []string
+	for _, node := range c.Nodes {
+		ipv4_addr := node.Config().MgmtIPv4Address
+		if ipv4_addr != "" {
```
hellt, line 445: better to keep the happy path on the main line
meaning that 
```
if ipv4_addr == ""
  continue
```
is a better alternative

clab/config.go
```diff
@@ -433,6 +436,35 @@ func (c *CLab) verifyLinks() error {
...
+			for i := range ipv4_list {
+				if ipv4_list[i] == ipv4_addr {
+					return fmt.Errorf("management IPv4 address %q appeared more than once in the topology file", ipv4_addr)
+				}
+			}
+			ipv4_list = append(ipv4_list, ipv4_addr)
```
hellt, line 451: this is a working solution, but not the optimal one
to check for duplicates you rather need to use a map as this will have constant search times, as opposed to the ranging through a list

see example https://replit.com/@RomanDodin/dupsDetection#main.go

PR #825, opened 2022-03-19
Implement docker ulimits

### LINKED ISSUE #824: Pass host's open files ulimit inside a container
The goal is to align max open file limit in the host OS and inside a container.

currently, if host's ulimit is low (e.g. 1024) and container starts with a higher number (e.g. ceos has 1073741816) this may lead to bugs in certain python scripts. For example: 

```bash
$ ulimit -n
1024
$ docker exec -it ceos bash
< inside the container now>
bash-4.2# cat test.sh
#!/usr/bin/python3
import resource
import os

max_fd = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
print(max_fd)
os.closerange(3, max_fd)

bash-4.2# ulimit -n
1073741816

bash-4.2# ./test.sh
1073741816

^Z
[1]+  Stopped                 ./test.sh
bash-4.2# bg
[1]+ ./test.sh &
bash-4.2# top -n1 | grep test
   2400 root      20   0   27504   7192   4500 R 100.0   0.1   0:18.87 test.sh
```

The above python script will block and eat 100% of available CPU core.

LimeHat:
Fix #824 

In this implementation both soft & hard limits will be set to rlimit.max (or 2^20, whichever is lower) because by default unix.rlimit returns the limits of the current process (containerlab binary in this case).
I assume that should be good enough, but if we encounter any issues (someone runs clab on a system with a very low rlimit.max? which is unlikely) we can rework the logic.

### DIFF
--- a/runtime/docker/docker.go
+++ b/runtime/docker/docker.go
@@ -335,11 +335,7 @@ func (d *DockerRuntime) CreateContainer(ctx context.Context, node *types.NodeCon
 	var rlimit unix.Rlimit
 	if err := unix.Getrlimit(unix.RLIMIT_NOFILE, &rlimit); err != nil {
 		log.Warnf("Unable to retrieve rlimit_NOFILE value: %v", err)
-		rlimit.Cur = rLimitMaxValue
 		rlimit.Max = rLimitMaxValue
-	}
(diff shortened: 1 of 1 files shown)


hellt:
thanks @LimeHat !

PR #832, opened 2022-03-21
Better OS detection

janosdebugs:
This PR adds better OS detection using `/etc/os-release` to avoid edge cases where the `rpm` binary is installed on Debian systems.


hellt:
Lgtm, thanks @janosdebugs 
Let's see if users will be happier with that version!

PR #834, opened 2022-03-27
Fix NextUI node tooltip to accommodate long image URLs.

ahmedalodhari:
NextUI has a fixed max-width for the node tooltip. This PR sets the size to be based on the space needed by the content.

nextus-node-tooltip


hellt:
thanks @ahmedalodhari 
that crosses off one of the feedback points https://github.com/srl-labs/containerlab/discussions/818.

PR #843, opened 2022-04-06
Pause / Unpause Containers

### LINKED ISSUE #772: add `pause` method to docker runtime
containers can be paused (https://docs.docker.com/engine/reference/commandline/pause/). we need to add this method to at least docker runtime to make it possible to start a container and immediately put it on pause so that we can do manipulations with interfaces, while container processes are paused.

steiler:
fix #772

The runtime Interface is extended with the following methods:
```golang
// Pause a container identified by its name
PauseContainer(context.Context, string) error
// UnPause / resume a container identified by its name
UnpauseContainer(context.Context, string) error
```
Docker, Podman & containerd do provide Pause and Unpause/Resume calls. So they are used.

Ignite however does not provide this functionality. Hence the implementation uses SIGSTOP and SIGCONT calls on the PGID (ProcessGroupID).

Code is just sparsely tested, and needs wiring up with the higher level functionality mentioned in #772.

### DIFF
--- a/runtime/containerd/containerd.go
+++ b/runtime/containerd/containerd.go
@@ -353,8 +353,7 @@ func (c *ContainerdRuntime) StartContainer(ctx context.Context, _ string, node *
 func (c *ContainerdRuntime) PauseContainer(ctx context.Context, cID string) error {
 	ctask, err := c.getContainerTask(ctx, cID)
 	if err != nil {
-		log.Debugf("container %s: %v", cID, err)
-		return nil
+		return err
 	}
 
 	err = ctask.Pause(ctx)
@@ -364,8 +363,7 @@ func (c *ContainerdRuntime) PauseContainer(ctx context.Context, cID string) erro
 func (c *ContainerdRuntime) UnpauseContainer(ctx context.Context, cID string) error {
 	ctask, err := c.getContainerTask(ctx, cID)
 	if err != nil {
-		log.Debugf("container %s: %v", cID, err)
-		return nil
+		return err
 	}
 
 	err = ctask.Resume(ctx)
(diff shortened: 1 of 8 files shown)


runtime/containerd/containerd.go
```diff
@@ -350,6 +350,28 @@ func (c *ContainerdRuntime) StartContainer(ctx context.Context, _ string, node *
 	return nil, nil
 }
 
+func (c *ContainerdRuntime) PauseContainer(ctx context.Context, cID string) error {
+	ctask, err := c.getContainerTask(ctx, cID)
+	if err != nil {
+		log.Debugf("container %s: %v", cID, err)
+		return nil
```
hellt, line 357: looks like a leftover from the dev time? Should return err?

runtime/containerd/containerd.go
```diff
@@ -350,6 +350,28 @@ func (c *ContainerdRuntime) StartContainer(ctx context.Context, _ string, node *
...
+
+	err = ctask.Pause(ctx)
+	return err
+}
+
+func (c *ContainerdRuntime) UnpauseContainer(ctx context.Context, cID string) error {
+	ctask, err := c.getContainerTask(ctx, cID)
+	if err != nil {
+		log.Debugf("container %s: %v", cID, err)
+		return nil
```
hellt, line 368: looks like a leftover from the dev time? Should return err?

utils/netlink.go
```diff
@@ -41,6 +44,24 @@ func LinkContainerNS(nspath, containerName string) error {
...
+	if len(pathElem) != 4 {
+		return 0, fmt.Errorf("unexpected result looking up container PID")
+	}
+	pid, err := strconv.Atoi(pathElem[1])
+	if err != nil {
+		return 0, fmt.Errorf("error converting the string part of the namespace link to int")
+	}
+	return pid, nil
+}
+
```
hellt, line 64: let's move this to `utils/containers.go`

runtime/containerd/containerd.go
steiler: Done. Removed the log, returning error

runtime/containerd/containerd.go
steiler: Done, removed the log, returning error

utils/netlink.go
steiler: is moved

Issue #801, opened 2022-02-24
support multiple `kind` names in topo file


hellt:
We need to add an ability to defined aliases to kind names that we use in the topology.
I would like to use official `vendor-platform` names instead of the short names we use today.

For instance:

- `srl` -> `nokia-srlinux`
- `ceos` -> `arista-ceos`
- `vr-sros` -> `nokia-sros`

This means that we need to support multiple names for the same `kind`

hellt:
done in #844

PR #844, opened 2022-04-06
allow for multiple names per kind

steiler:
With this PR multiple kind names can be assigned to a kind implementation.

Kinds will register themselves with all their defined `kindnames`.
The default credentials map is now filled from the kind implementation via the `nodes.SetDefaultCredentials(...)` function.
Retrieving the default credentials works via `nodes.GetDefaultCredentialsForKind(<AnyOfTheAssignedKindNames>)`

I want to also get rid of the remaining:
```
const (
	NodeKindBridge          = "bridge"
	NodeKindHOST = "host"
	NodeKindOVS  = "ovs-bridge"
	NodeKindSRL  = "srl"
)
```
however, therefore we need to also work on the clab.config.go since it checks these Kinds. Therefore I had to also come up with a hack in the srl.init() where the kind is overwritten and forced to be nodes.NodeKindSRL.


steiler:
issue for the kind consts created #845

steiler:
Implementing #801

hellt:
crazy week here. will try to review that by the mid next week

nodes/node.go
```diff
@@ -23,35 +24,25 @@ const (
...
-	NodeKindVrXRV            = "vr-xrv"
-	NodeKindVrXRV9K          = "vr-xrv9k"
-	NodeKindVrNXOS           = "vr-nxos"
-	NodeKindIPInfusionOCNOS  = "ipinfusion-ocnos"
-	NodeKindKeysightIxiaCOne = "keysight_ixia-c-one"
+	NodeKindBridge = "bridge"
+
+	NodeKindHOST = "host"
+	NodeKindOVS  = "ovs-bridge"
+	NodeKindSRL  = "srl"
```
hellt, line 31: thanks @steiler this change LGTM!
I think we can chase those dependencies in config.go in a separate PR

hellt (approved):

Issue #837, opened 2022-03-31
add a check for sse3 cpu instruction availability 


hellt:
sse3 instructions must be available for nodes which are based on dpdk
it can be missing if users deploy/use a VM that was created with some old CPU emulated. It is easy to fix with running a VM that has full cpu host support

the check must be done during the topology checks stage, the lab must be halted if sse3 is missing and srl nodes are part of the topology and we must emit the warning message otherwise

nemith:
This doesn't work well for docker in docker or docker in lima as the SSE3 will be done at the docker level.

hellt:
@nemith I'll check if disabling sse3 cpu capability check will work, testing it on srlinux in dind

PR #846, opened 2022-04-07
Check for SSE3 CPU availability

steiler:
Implementing #837
The `NodeConfig` now contains an `SSE3Required bool` flag. 
If SSE3 is required just set this flag to true in the kind `init()`
If a topology is deployed that requires SSE3 it is checked and an exception is thrown if unavailable. If not required by any kind we continue.
Bools are by default initialized with the value false. So this flag needs to only be touched when SSE3 is explicitely required.


hellt:
What if for NodeConfig struct we create a field `HostRequirements` which will be outlined like this

```go
{
  SSE3 bool
}
```

with an idea that eventually we may populate it with more cpu instruction sets or other host related requirements?

hellt:
Hi @steiler 
in [a5c880c](https://github.com/srl-labs/containerlab/pull/846/commits/a5c880c9ac2e6c44833568fbf95e062985404c99) I added what I was referring to as using a struct for host requirements

Check it out, I think it is quite straightforwards and is inline with what you did initially

steiler:
Yes looking good!
whats your opinion on implementing the `MinCPURequired` and alike in the same way?
That would be uniform and keep all kind related difference as a local definition on the kind.

hellt:
Yes. We will do this in a separate PR

PR #847, opened 2022-04-07
adding env-files to kind and node topology definition

### LINKED ISSUE #743: add support for `env-file` node parameter
Currently env vars provided to the nodes can be passed in a form of a `k=v` pairs under the `env` node property.

Add ability to set a path to the env-compatible file to read the variables from it, instead of providing them in the topology. More details about `--env-file` docker run parameter - https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file

steiler:
fix #743
kind and nodes allow for the env-files stanza now.
it expects an array of filenames to load environment variables from.
The base path for relative env-file references is the directory that contains the topology definition .clab.yml


clab/config.go
```diff
@@ -260,9 +260,17 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
...
 	}
+	var err error
+
+	// Load content of the EnvVarFiles
+	envFileContent, err := utils.LoadEnvVarFiles(c.TopoFile.dir, c.Config.Topology.GetNodeEnvFiles(nodeName))
+	if err != nil {
+		return nil, err
+	}
+	// Merge EnvVarFiles content and the existing env variable
+	nodeCfg.Env = utils.MergeStringMaps(envFileContent, nodeCfg.Env)
```
hellt, line 271: Hi @steiler 
let's add a check for env files to be present in a system similar to that code - https://github.com/srl-labs/containerlab/blob/c9f6ddb6c6a04e55ec94b930e970eb67add9bfb2/clab/config.go#L467

There I propose to take it a step further if a path exists, replace the original value of the path with the one that is returned by `utils.ResolvePath` function. That way after doing the check for the path existance you will have your node with the already resolved path. This will remove the need to do the path resolving part that you do in `env.go#L99`.  
Additionally, you won't need to pass c.Topofile.dir to that LoadEnvVarFiles func

steiler: As discussed offline, utilizing the "verify-Stage" is too late, since these tests are performed after the Nodes are created.
But nevertheless did I shift to using the utils.ResolvePath now.
So please go ahead and check I you think this is now ready to be merged.

hellt:
thanks @steiler ! great contribution

PR #850, opened 2022-04-08
topology export

hellt:
copied from #838 because my gh cli seems to be screwed and I couldn't push my fixes directly to the original branch

@bortok please check my latest additions, most important one could be that I removed `config-path` entirely, as we don't use it. I hope graphite doesn't use it either.
The rest of the changes in my latest commits are cosmetic/nits


bortok (changes requested):
Shall we add a tag for json to make sure the mutex won't be marshaled? Removing a tag doesn't prevent it from being exported. Although, I don't see this field in the full Clab export anyway, not sure why.
```Go
	m             *sync.RWMutex `json:"-"`
````

hellt:
Thanks @bortok for a quick review
I am merging this, and we sorted the struct tag privately, so we are good to go

PR #856, opened 2022-04-11
allow for sysctl configuration via the sysctls map in the topology definition yml

### LINKED ISSUE #766: add `sysctl` node option to configure system calls on a per node basis
we do have sysctls exposed in the internal node configuration, but it is not exposed to a user yet.

steiler:
fix #766


hellt:
Hey @steiler 
so I have changed MergeStringMaps to work well with non nil empty maps and thus removed sysctl initialization

looks good so far. I am about to merge this. Thanks again!

PR #862, opened 2022-04-18
Removal of dangling veths

### LINKED ISSUE #842: Delete host interface when container is destroyed
When a lab is destroyed it takes about 2 minutes until the host interfaces of a container are removed from the containerlab host. The removal of the host interfaces of a container should be done immediately with the destroy of the container. 

Example interfaces defined  in the topology file:

```
links:
     - endpoints: ["BorderLeaf_3:eth5", "host:vx102-eth3"]
     - endpoints: ["BorderLeaf_3:eth6", "host:xxx"]
```

LimeHat:
Hopefully this fixes #842

### DIFF
--- a/clab/netlink.go
+++ b/clab/netlink.go
@@ -144,7 +144,6 @@ func (c *CLab) RemoveHostVeth(l *types.Link) (err error) {
 		if err != nil {
 			log.Debugf("Link %q is already gone: %v", l.B.EndpointName, err)
 		}
-	default:
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


clab/netlink.go
```diff
@@ -118,6 +118,37 @@ func (c *CLab) CreateVirtualWiring(l *types.Link) (err error) {
...
+		if err != nil {
+			log.Debugf("Link %q is already gone: %v", l.B.EndpointName, err)
+			break
+		}
+		err = netlink.LinkDel(link)
+		if err != nil {
+			log.Debugf("Link %q is already gone: %v", l.B.EndpointName, err)
+		}
+	default:
+	}
```
hellt, line 148: safe to remove empty default?

PR #875, opened 2022-05-05
Add note for OSX installation

barajus:
Hi, 
Mention that some modules are missing, and provide a link with instruction for recompilation + fix minor typo
Julien


hellt:
thanks @barajus !

PR #881, opened 2022-05-07
add ^et interface pattern matching to if-wait script for arista ceos

### LINKED ISSUE #880: interface name pattern matching for ceos
another one from the discord, interfaces named etXX are not matched by the if-wait script

LimeHat:
Should fix #880


hellt:
good catch
thanks

Issue #634, opened 2021-09-30
Display "Published ports" information when output format is JSON


hellt:
### Discussed in https://github.com/srl-labs/containerlab/discussions/633

<div type='discussions-op-text'>

<sup>Originally posted by **fredsem** September 30, 2021</sup>
Hello,
When I inspect the labs are running, I can see the detail of the published ports in mysocket.io.

```
[root@mylab ~]# containerlab inspect --all
+---+-----------------------------------+----------+-----------------------+--------------+---------------------------------+------------+-------+---------+---------------+--------------+
| # |             Topo Path             | Lab Name |         Name          | Container ID |              Image              |    Kind    | Group |  State  | IPv4 Address  | IPv6 Address |
+---+-----------------------------------+----------+-----------------------+--------------+---------------------------------+------------+-------+---------+---------------+--------------+
| 1 | ../opt/mylab/topologies/lab01.yml | lab01    | clab-lab01-R01        | 911157b0b763 | vrnetlab/vr-sros:20.10.R1       | vr-sros    |       | running | 10.10.10.4/24 | NA           |
| 2 |                                   |          | clab-lab01-R02        | 3b35023c0cbe | vrnetlab/vr-sros:20.10.R1       | vr-sros    |       | running | 10.10.10.3/24 | NA           |
| 3 |                                   |          | clab-lab01-mysocketio | 2e688c170687 | ghcr.io/hellt/mysocketctl:0.4.0 | mysocketio |       | running | 10.10.10.2/24 | NA           |
+---+-----------------------------------+----------+-----------------------+--------------+---------------------------------+------------+-------+---------+---------------+--------------+
**Published ports:
┌──────────────────────────────────────┬─────────────────────────────────────────┬─────────┬──────┬────────────┬─────────────────┐
│ SOCKET ID                            │ DNS NAME                                │ PORT(S) │ TYPE │ CLOUD AUTH │ NAME            │
├──────────────────────────────────────┼─────────────────────────────────────────┼─────────┼──────┼────────────┼─────────────────┤
│ dad6ed7f-f601-4b35-aba1-233711334485 │ lingering-glitter-9346.edge.mysocket.io │ 31181   │ tcp  │ false      │ clab-R01-tcp-22 │
│ e2eccbee-09bb-4c13-bbe6-9df82d8dfd5b │ late-pine-8645.edge.mysocket.io         │ 50297   │ tcp  │ false      │ clab-R02-tcp-22 │
[...]

PR #887, opened 2022-05-11
Display "Published ports" information when output format is JSON

steiler:
This PR is to implement what is mentioned in Issue #634.

The actual implementation is a breaking change, since the up until now top level array of container data moved under the "container_data" entry and at the same time introduces another array called mysocket_data which keeps the mysocketio information.

This still needs to be tested. Actually it seems like my endpoint protection considers mysocketio harmful and blocks it.


steiler:
@hellt: Tested and ready for review now.

hellt:
Hi @steiler 
thanks for that
do you mind checking the failed tests?

steiler:
Alright, all green!
Be aware, that this is a breaking change in inspect commands json output.
it used to output the anonymous array of containers which is now pushed a level down under the key `container_data` and right next to it the `mysocketio_data` key is being introduced. 

https://github.com/srl-labs/containerlab/pull/887/files#diff-f8d2bd35e1a8c9b86c8c4448dde00bc5480652f85ed0630efc012c3906d3a14fR280-R292

clab/file.go
```diff
@@ -35,6 +35,10 @@ type TopoFile struct {
 	name     string // file name without extension
 }
 
+func (tf *TopoFile) GetDir() string {
```
hellt, line 38: just curious why you wanted this function, instead of accessing the field?

cmd/inspect.go
```diff
@@ -76,10 +79,6 @@ var inspectCmd = &cobra.Command{
 			return fmt.Errorf("failed to list containers: %s", err)
 		}
 
-		if len(containers) == 0 {
-			log.Println("no containers found")
-			return nil
-		}
```
hellt, line 82: we need to find a way of not printing an empty table if no containers are found

right now it prints an empty one, and before we used "no containers found"

cmd/inspect.go
```diff
@@ -197,19 +209,97 @@ func printContainerInspect(c *clab.CLab, containers []types.GenericContainer, fo
...
+}
+
+// deduceMySocketIoTokenFileFromBindMounts searches through the topology to find a node of kind mysocketio.
+// if that is found, the bindmounts are searched for ".mysocketio_token" and the path is being converted into an
+// absolute path and returned.
+// If the
+func deduceMySocketIoTokenFileFromBindMounts(nodes map[string]nodes.Node, configPath string) (string, error) {
+	for _, node := range nodes {
+		// search for mysocketio kind in topology config
+		if node.Config().Kind == "mysocketio" {
```
hellt, line 288: lets maitain the happy path here and `continue` if kind is not mysocketio

utils/mysocketio/mysocketio.go
```diff
@@ -0,0 +1,111 @@
+package mysocketio
```
hellt, line 1: lets move this package to upper level, not under utils

clab/file.go
steiler: the issue is, that we put applogic in the cmd package ... what we should do is put all the app logic in the clab package and just do parameter parsing and population in the cmd package... that is not too clean of a structure. As a result of this, from the cmd package we are unable to access the lower case dir variable... instead of opting for making the var public (CamelCase) I introduced this getter.
We could come up with a seperate PR and seperate app logic from parameter paring and initialization if you like, I'd be happy to do so.

cmd/inspect.go
steiler: The issue was, that this was printed even in the json case ... which is bad when json output is expected.
What exactly is the issue with the empty table?

utils/mysocketio/mysocketio.go
steiler: done

cmd/inspect.go
steiler: done

cmd/inspect.go
hellt: it just doesn't make a lot of sense to me printing an empty table. I haven't seen any other tool doing that. Gives an impression that something

we need to make sure it is not printed when format is set to json and printed when default output. I can have a look at it if you have more interesting things to have a crack on

steiler: I'd personally prefere a stable output method. But nevermind.
Check out the new commit. 
https://github.com/srl-labs/containerlab/pull/887/commits/bcc6a2d213394fabb827e14600e593bb199fc4a3

steiler: why revert the test based on json?
lets keep both?! Just to make sure json output is also fine in the zero container case.

cmd/inspect.go
```diff
@@ -161,55 +165,151 @@ func printContainerInspect(c *clab.CLab, containers []types.GenericContainer, fo
...
 	}
+	return result, nil
+}
 
-	fmt.Println("Published ports:")
-	fmt.Println(string(stdout))
+// mySocketIoTokenFileFromBindMounts finds a node of kind mysocketio.
+// if that is found, the bindmounts are searched for ".mysocketio_token" and the path is being converted into an
+// absolute path and returned.
+func mySocketIoTokenFileFromBindMounts(_nodes map[string]nodes.Node, configPath string) (string, error) {
```
hellt, line 292: Hi @steiler 
there is one little thing with this approach which I am ok to let slip

when you do `clab ins -a` this func will not work since we don't have `c.Nodes` populated as we don't read topology files.

One workaround for that is to use GenericContainer type where i added mounts information in this commit. We can rework this function to work on a slice of container info, rather than c.Nodes, and get the binds info from it. But then the problem is what to do if you have >1 mysocketio node, with different tokens...

So for the time being I just made a shunt so that mysocketdata is not fetched if we use `inspect --all`

cmd/inspect.go
hellt: valid point. need to get back

cmd/inspect.go
steiler: oh ok, I see.
What if we then simply print multiple tables indexed with either the containername or maybe better with the lab name, availabel via the container labels?!
In the json case, that would require a change to the output format as well ... so we eiter make the highlevel mysocketio entry a simple list, also carrying the labname as on item of the enclosed entries or a map with the labname as index... 
Thought?

hellt: yes, I think adding lab name as a field for mysocket data is a valid approach. I would go with a list where lab name is a k:v pair inside the list

when `inspect --all` is used and the output is in the table format, then I think it is safe to have a single table that lists set up sockets with a field listing a lab for which these sockets were created.
we would just need to make sure we list through all containers which have `label[kind]=mysocketio`, get their respective source mount path and fetch the sockets.

Quite some work. Maybe better to do that in a separate PR?

hellt:
I think we have discussed this before and wrongfully asked to put info message back when no containers found.
But it makes sense to return an empty json when no containers found, instead of an invalid json as it is today.

steiler:
Alright, @hellt, took me quite a while to get back to this, but take a look at the actual changes.
using the GenericContainer type, thereby allowing for "-a" but also "-t" in both, table as well as json output format.

hellt:
Thanks @steiler !

PR #889, opened 2022-05-13
feat: add support to copy startup-config

crankynetman:
Adds in the required clab support to do the bits in https://github.com/hellt/vrnetlab/pull/78. Basically we just pull in the startup config directive, copy it to the lab directory, and then bind mount it into the container. vrnetlab does the rest from a config load standpoint.


hellt:
thanks @crankynetman 
that is a beatiful PR and a long-awaited feature

Issue #888, opened 2022-05-12
srlinux doesn't have a way to talk to docker's DNS server


hellt:
due to srl nature of using custom netns's, docker' DNS server running on 127.0.0.11:53 is not put in srbase-mgmt netns, but resides in a root netns.

We need to workaround this either by moving docker' dns to srbase-mgmt or by adding static entries to /etc/hosts in srl post-deploy stage

/cc @LimeHat

steiler:
I was thinking of using `socat` to maybe proxy dns requests through fifo pipes. Tried it, but it seems to have the limitations, that the pipe would be shared by the different requests and dig and nslookup do complain about responses from different IP then requested.

```
socat -T15 udp4-recvfrom:53,reuseaddr,fork PIPE:/tmp/req      # namespace not reaching dns server
socat PIPE:/tmp/req UDP:8.8.8.8:53                          # namespace reaching the dns server
```
Maybe someone has a different more clever idea around this, that's why I'm sharing my experience.

hellt:
As @LimeHat found during his walk in the lower docker layers, docker mounts a file to each container at /etc/hosts. The hostpath can be found from container inspect


I propose we add that HostPath to Node definition state 
Then in srl post deploy we need to add the same content we add to systems /etc/hosts today

steiler:
Just found out that powerdns provides a pipe backend (https://doc.powerdns.com/authoritative/backends/pipe.html).
So it should be possible to copy the powerdns binary into the image, start powerdns locally in the srbase-mgmt namespace and get dns requests out via a mounted pipe.

The other end of the pipe would also need to run somewhere, maybe a seperate container, forwaring the requests on to the docker or any other dns server. This would probably provide the most flexibility. 

Here is a simple sample backend implementation I found: https://gist.github.com/sokratisg/10069682

However, if this level of flexibility isn't even needed this is probably to complicated.
So simply consider this as a side note.

LimeHat:
Thanks @steiler .
I think we're looking for something simple for now.
With the proposed 'external dns' approach, for that matter, you don't need to bring any binaries into srlinux containers - you can just configure dns in srlinux. So yes, we can run our own dns-infra container for service discovery (just like k8s does), but I feel like it is a bit more complexity than we need

LimeHat:
So I'm thinking to implement /etc/hosts first (along the lines @hellt summarized above) and then if we encounter any practical limitations or issues with that approach then we can revisit

PR #891, opened 2022-05-25
/etc/hosts population for srl nodes

LimeHat:
Fix for #888 as discussed

### DIFF
--- a/runtime/containerd/containerd.go
+++ b/runtime/containerd/containerd.go
@@ -808,3 +808,9 @@ func (c *ContainerdRuntime) DeleteContainer(ctx context.Context, containerID str
 
 	return nil
 }
+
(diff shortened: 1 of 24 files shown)


runtime/containerd/containerd.go
```diff
@@ -808,3 +808,9 @@ func (c *ContainerdRuntime) DeleteContainer(ctx context.Context, containerID str
 
 	return nil
 }
+
+// GetHostsPath returns fs path to a file which is mounted as /etc/hosts into a given container
+// TODO: do we need it here? currently no-op
```
hellt, line 813: containerd integration is left out. no one is looking into it as of now

hellt:
thanks @LimeHat !

PR #919, opened 2022-06-28
Podman v4 support

LimeHat:
Updated podman runtime to support v4 (exlusively, v3 is now retired).
Finally we can do proper IPv6 support & bridge naming.


LimeHat:
Uh oh CICD build process is complaining

I was using go1.18 on my local setup so I guess there's a difference that matters..  @hellt please help :-)

hellt:
okay, it looks like ubuntu (even on 22.04) comes with podman v3 only
so we have a tough choice

1. stand up a self hosted runner with fedora and deal with all the complications of maintaining a clean testing environment (biggest problem is to allow for concurrent test runs)
2. ditch CI tests 🌚

LimeHat:
Thanks for the fixes, Roman.

> okay, it looks like ubuntu (even on 22.04) comes with podman v3 only so we have a tough choice
> 
> 1. stand up a self hosted runner with fedora and deal with all the complications of maintaining a clean testing environment (biggest problem is to allow for concurrent test runs)
> 2. ditch CI tests 🌚
> 3. try to install podman on ubuntu using unofficial builds (like [this](https://github.com/containers/podman/issues/13302#issuecomment-1050435309)) - this would be my preference

Yeah p.3 sounds fine if we can make it work. I'll look into it.

LimeHat:
@hellt can we try to build it ourselves? i tried the binary package listed in the issues you linked, but it didn't quite work for me (ubuntu has an older glibc version)

But building process seems okayish, I installed three packages
1) make
2) libseccomp-dev
3) libsystemd-dev

And then was able to build the binaries via "make binaries" (which can be combined with "make install.bin && make install.systemd")

I didn't start with a clean ubuntu system, so not sure if its going work on github as well as on my system.

hellt:
That should be possible.
I think to optimize the CI pipeline the build step should start right away
to run in parallel with containerlab build step.
Then the job that tests podman should require 'build-podman' step to
succeed.

In the build step I think some time can be chopped off using env vars to
exclude podman subsystems that we exclude in containerlab...

hellt:
Hi @LimeHat 
in recent commits I have added podman v4 build steps. The build steps includes only `make binaries` stage.
Then I download the `podman` artifact in the `smoke-tests` job to `/usr/local/bin` path in hope that it is all it needs, but apparently it is not, since tests are still failing.

The result of `make binaries` is  these three bins
```
podman  podman-remote  rootlessport
```

maybe we need other bins as well to enable podmanv4? Or maybe we need to use those other make targets to be called...

hellt:
@LimeHat 
ah, okay, solved
the path on ubuntu is `/usr/bin`. Now it seems v4 is working fine, just a couple of tests failing, which has to do with actual podman v4 support, and not the build issue

LimeHat:
Awesome, thank you for working on that Roman!

Hey is there a way to get a detailed output for the failed test?

```
Verify Hosts entries exist :: Verification that the expected /etc/... | FAIL |
'6' does not contain '4'
```

Not entirely clear what has failed there

Also, i think we can enable ipv6 tests for podman now

hellt:
> Hey is there a way to get a detailed output for the failed test?

yeah, if you open the summary of a CI run and scroll at the bottom https://github.com/srl-labs/containerlab/actions/runs/2711962806 you will see archives with detailed logs

once downloaded and unarchived you will get to see the robotframework' output
image

which actually shows that the first failed case is caused by a flawed case condition (should be 6)

LimeHat:
All right, I think things are good now

hellt:
Isn't that great? Thanks @LimeHat, great PR

PR #920, opened 2022-06-28
added v4 gateway functionality

### LINKED ISSUE #796: passing variable of default gateway ipv4-gw (arista cEOS example)
HI,

Could someone add an option to pass variable representing docker network (IP) assigned to bridge (ipv4-gw) in topology YAML file to make it possible to use in startup-config (to pass it as a variable in jinja template)? Or maybe it is easy to extract and pass - please give me a hint :)
For instance, for Arista cEOS it could look just that:

in the topology file we have

name: test
mgmt:
  bridge: oobbridge
  network: oobnetwork
  ipv4_subnet: 172.20.20.0/24
  **ipv4-gw**: 172.20.20.11                 <= variable to exclude from here..

topology:
  kinds:
    ceos:
      image: ceos64-lab:4.27.1.1F
      **startup-config**: vrf-mgmt.cfg   <= and pass it as variable to jinja template in startup config.

The **startup-config** file **vrf-mgmt.cfg** looks like:

hostname {{ .ShortName }}
!
username admin privilege 15 secret admin
!
service routing protocols model multi-agent
!
interface Management0
{{ if .MgmtIPv4Address }}ip address {{ .MgmtIPv4Address }}/{{.MgmtIPv4PrefixLength}}{{end}}
{{ if .MgmtIPv6Address }}ipv6 address {{ .MgmtIPv6Address }}/{{.MgmtIPv6PrefixLength}}{{end}}
!
interface Management1
   vrf mgmt
!
vrf instance mgmt
!
no ip routing vrf mgmt
!
management api gnmi
   transport grpc default
      vrf mgmt
!
management api netconf
   transport ssh default
      vrf mgmt
!
management api http-commands
   vrf mgmt
      no shutdown
!
ip route vrf mgmt 0.0.0.0/0 {{ .**ipv4-gw** }}                        <== I would like to pass information about gw to startupconfig
end

for some reason the  ipv4_subnet would be available as well (i.e. for some ACL)

Thanks in , regards
Marek

networkop:
fixes #796

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -323,6 +323,7 @@ func enrichNodes(containers []types.GenericContainer, nodesMap map[string]nodes.
 				node.Config().MgmtIPv4PrefixLength = c.NetworkSettings.IPv4pLen
 				node.Config().MgmtIPv6Address = c.NetworkSettings.IPv6addr
 				node.Config().MgmtIPv6PrefixLength = c.NetworkSettings.IPv6pLen
+				node.Config().MgmtIPv4Gateway = c.NetworkSettings.IPv4Gw
 			}
 			node.Config().ContainerID = c.ID
 		}
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -307,6 +307,7 @@ func setFlags(conf *clab.Config) {
 	}
 }
 
+// enrichNodes add container runtime assigned information (such as dynamically assigned IP addresses) to the nodes
 func enrichNodes(containers []types.GenericContainer, nodesMap map[string]nodes.Node) {
 	for i := range containers {
 		c := &containers[i]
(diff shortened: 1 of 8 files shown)


hellt:
Hey @networkop 

Wanted to propose a slight change in the logic 
(discard lots of changes in ceos.go, it is due to struct receiver renaming `s ->n`)

I wanted to make IPv4 GW address available earlier, before the node starts. That way we can use this var in ceos default startup config template and with that let users put this route under the VRF should they wish so.

To do that, I added the following to docker's runtime network creation step.

https://github.com/srl-labs/containerlab/pull/920/files#diff-c386fb09f5478e73399310d641d4b25900cb2a240d48a693d6d3d0dfdc646204R230

what do you think?

hellt:
I think I am good with merging it

have a nice one @networkop ,
image

bewing:
Quick question -- was there a technical limitation preventing IPv6Gateway from also being exposed to cEOS, or just a lack of need?

hellt:
> Quick question -- was there a technical limitation preventing IPv6Gateway from also being exposed to cEOS, or just a lack of need?

yes, lack of need I suppose. Don't see why this can't be added

PR #931, opened 2022-07-07
Fix release date for 0.28

toerb:
Hi there,

the release 0.28 is greeting from the future!
Correct date should be 2022-06-27.

BR
Tobias


hellt:
Thanks!

PR #935, opened 2022-07-10
update github.com/docker/distribution to fix checksum error

vincentbernat:
When trying to build, we get this error:

```
verifying github.com/docker/distribution@v2.8.0+incompatible: checksum mismatch
        downloaded: h1:u9vuu6qqG7nN9a735Noed0ahoUm30iipVRlhgh72N0M=
        go.sum:     h1:l9EaZDICImO1ngI+uTifW+ZYvvz7fKISBAKpg+MbWbY=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```

See: https://github.com/distribution/distribution/issues/3590. The tag
has been updated and therefore, the recorded checksum in `go.sum` does
not match what a fresh build may pull from the web.

The easiest fix is just to bump the dependency, as done here.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/containernetworking/plugins v0.9.1
 	github.com/containers/podman/v3 v3.4.4
 	github.com/digitalocean/go-openvswitch v0.0.0-20201214180534-ce0f183468d8
-	github.com/docker/distribution v2.8.0+incompatible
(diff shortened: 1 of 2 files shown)


hellt:
Thanks @vincentbernat !

PR #940, opened 2022-07-11
expose IPv6 gateway and add to cEOS config

bewing:
Add MgmtmIPv6Gateway to NodeConfig and use to configure default IPv6
route for cEOS devices.


hellt:
thansk @bewing 
looks like when the PR is run against our repo, the clab image can be pulled nicely

PR #941, opened 2022-07-11
support cEOS management VRFs

bewing:
Allow specification of a Management VRF name in topology file.

TODO:

- [x] Docs
- [x] Tests
- [x] Suggestions?

### DIFF
--- a/docs/manual/kinds/ceos.md
+++ b/docs/manual/kinds/ceos.md
@@ -204,6 +204,28 @@ cEOS Ma0 interface will be configured with a random MAC address with `00:1c:73` 
 
 A default ipv4 route is also created with a next-hop of the management network to allow for outgoing connections.
 
+#### MGMT VRF
+The default empty configuration supports placing the management interface into a VRF to isolate it from the main device routing table.  Passing the environment variable `CLAB_MGMT_VRF` in either the kind or node definition will activate this behavior, and alter the management services configuration to also reflect the management VRF.  You can duplicate this when using the `startup-config` by starting from the linked template below.
+
+```yaml
+# example topo file with management VRF
+# node1 will have vrf MGMT
+# node2 will have vrf FOO
+name: ceos_vrf
+topology:
+  kinds:
+    ceos:
+      env:
+        CLAB_MGMT_VRF: MGMT
+  nodes:
+    node1:
+      kind: ceos
+    node2:
+      kind: ceos
+      env:
+        CLAB_MGMT_VRF: FOO
+```
+
 #### User defined config
 It is possible to make ceos nodes to boot up with a user-defined config instead of a built-in one. With a [`startup-config`](../nodes.md#startup-config) property a user sets the path to the config file that will be mounted to a container and used as a startup config:
 
--- a/nodes/ceos/ceos.cfg
+++ b/nodes/ceos/ceos.cfg
@@ -7,20 +7,28 @@ username admin privilege 15 secret admin
 !
 service routing protocols model multi-agent
 !
-{{ if .MgmtIPv4Gateway }}ip route 0.0.0.0/0 {{ .MgmtIPv4Gateway }}{{end}}
-{{ if .MgmtIPv6Gateway }}ipv6 route ::0/0 {{ .MgmtIPv6Gateway }}{{end}}
+{{- if .Env.CLAB_MGMT_VRF }}
+vrf instance {{ .Env.CLAB_MGMT_VRF }}
+!
+{{end}}
+{{ if .MgmtIPv4Gateway }}ip route {{ if .Env.CLAB_MGMT_VRF }}vrf {{ .Env.CLAB_MGMT_VRF }} {{end}}0.0.0.0/0 {{ .MgmtIPv4Gateway }}{{end}}
+{{ if .MgmtIPv6Gateway }}ipv6 route {{ if .Env.CLAB_MGMT_VRF }}vrf {{ .Env.CLAB_MGMT_VRF }} {{end}}::0/0 {{ .MgmtIPv6Gateway }}{{end}}
 !
 interface {{ .MgmtIntf }}
+{{ if .Env.CLAB_MGMT_VRF }} vrf {{ .Env.CLAB_MGMT_VRF }}{{end}}
 {{ if .MgmtIPv4Address }}ip address {{ .MgmtIPv4Address }}/{{.MgmtIPv4PrefixLength}}{{end}}
 {{ if .MgmtIPv6Address }}ipv6 address {{ .MgmtIPv6Address }}/{{.MgmtIPv6PrefixLength}}{{end}}
 !
 management api gnmi
    transport grpc default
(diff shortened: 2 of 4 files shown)


bewing:
Primary driver for me is to completely decouple the host/management routing from whatever topology I'm designing.  Not sure what other node types this can be implemented in as well.

types/types.go
```diff
@@ -51,6 +51,7 @@ type MgmtNet struct {
 	IPv6Gw         string `yaml:"ipv6-gw,omitempty" json:"ipv6-gw,omitempty"`
 	MTU            string `yaml:"mtu,omitempty" json:"mtu,omitempty"`
 	ExternalAccess *bool  `yaml:"external-access,omitempty" json:"external-access,omitempty"`
+	VRF            string `yaml:"vrf,omitempty" json:"vrf,omitempty"`
```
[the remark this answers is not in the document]
bewing: `Envs` seems like a loaded term -- might imply actually injecting environment variables into the containers?  Perhaps ` .TmplEnv` instead?

bewing:
New changeset pushed.

Question:  Should the mgmt VRF be part of the default template, or pushed in `ceosPostDeploy` via the CLI?  The latter would allow the feature to be used in conjunction with `startup-config` option.  I'm undecided if that's a good idea or not.

nodes/ceos/ceos.cfg
```diff
@@ -7,20 +7,28 @@ username admin privilege 15 secret admin
...
+{{ if .MgmtIPv6Gateway }}ipv6 route {{ if .Env.MGMT_VRF }}vrf {{ .Env.MGMT_VRF }} {{end}}::0/0 {{ .MgmtIPv6Gateway }}{{end}}
 !
 interface {{ .MgmtIntf }}
+{{ if .Env.MGMT_VRF }} vrf {{ .Env.MGMT_VRF }}{{end}}
 {{ if .MgmtIPv4Address }}ip address {{ .MgmtIPv4Address }}/{{.MgmtIPv4PrefixLength}}{{end}}
 {{ if .MgmtIPv6Address }}ipv6 address {{ .MgmtIPv6Address }}/{{.MgmtIPv6PrefixLength}}{{end}}
 !
 management api gnmi
    transport grpc default
+{{ if .Env.MGMT_VRF }} vrf {{ .Env.MGMT_VRF }}{{end}}
```
hellt, line 24: misaligned spacing in the term `vrf` doesn't cause issues with eos? I think no, but just checking

tests/03-basic-ceos/01-two-ceos.robot
```diff
@@ -46,6 +46,11 @@ Ensure IPv6 default route is in the config file
     Log    ${f}
     Should Contain    ${f}    ipv6 route
 
+Ensure MGMT VRF is in the config file
+    ${f} =    OperatingSystem.Get File    ${EXECDIR}/clab-${lab-name}/${node1-name}/flash/startup-config
+    Log    ${f}
+    Should Contain    ${f}    vrf instance MGMT
+
```
hellt, line 53: would it be more correct to check if router accepts those vrf commands by executing a `show` command against it and checking that the output contains the expected vrf?

docs/manual/kinds/ceos.md
```diff
@@ -204,6 +204,28 @@ cEOS Ma0 interface will be configured with a random MAC address with `00:1c:73`
...
+```yaml
+# example topo file with management VRF
+# node1 will have vrf MGMT
+# node2 will have vrf FOO
+name: ceos_vrf
+topology:
+  kinds:
+    ceos:
+      env:
+        MGMT_VRF: MGMT
```
hellt, line 219: to avoid name collisions, lets use `CLAB_MGMT_VRF` env var name, to clearly say this is containerlab-significant env var.

most vars we introduce in clab follow this prefix principle

hellt (changes requested):

hellt:
> Question: Should the mgmt VRF be part of the default template, or pushed in `ceosPostDeploy` via the CLI?

I would keep it in the default template. This seems non intrusive for labs that don't use VRF based management, and at the same time allows users to intriduce this feature from the get go with a single env var definition

docs/manual/kinds/ceos.md
bewing: Changed in latest changeset

tests/03-basic-ceos/01-two-ceos.robot
bewing: Updated test -- I may need to iterate as I didn't try to test locally.

nodes/ceos/ceos.cfg
bewing: It didn't in my testing, but in the interest of completeness I adjusted the spacing.

tests/03-basic-ceos/01-two-ceos.robot
```diff
@@ -46,6 +46,13 @@ Ensure IPv6 default route is in the config file
     Log    ${f}
     Should Contain    ${f}    ipv6 route
 
+Ensure MGMT VRF is present
+    ${rc}    ${output} =    Run And Return Rc And Output
+    ... sudo docker exec -it clab-${lab-name}-${node1-name} /bin/Cli -p 15 -c "show vrf MGMT"
+    Log    ${output}
+    Should Be Equal As Integers    ${rc}    0
+    Should Contain  ${output}   MGMT
+
```
hellt, line 55: I will fix this test @bewing

hellt: @bewing 
I think the test functionally works now, but I am curious if the output is considered to be healthy?

```
time="2022-07-18T07:32:49Z" level=info msg="Parsing & checking topology file: 03-ceos01-clab.yml"
time="2022-07-18T07:32:50Z" level=info msg="Executed command 'Cli -p 15 -c 'show vrf MGMT'' on clab-03-01-two-ceos-n1. stdout:   VRF        RD              Protocols       State             Interfaces
---------- --------------- --------------- -------------------- -----------
    MGMT       <not set>       ipv4,ipv6       v4:incomplete,    Management0                                               v6:incomplete                                                              
```

i.e. I am not sure if `v4: incomplete` is expected?

Maybe you can try out a binary built with the latest commit and see it in action?

docker run --rm -v $(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 pull ghcr.io/srl-labs/clab-oci:b1bbf898

hellt (approved):

tests/03-basic-ceos/01-two-ceos.robot
bewing: The output from the command on 4.28.1F is
```
$ docker exec -it clab-mgmt-spine1 /bin/Cli -p 15 -c "show vrf MGMT"
   VRF        Protocols       State         Interfaces
---------- --------------- ---------------- ----------
   MGMT       IPv4            no routing    Ma0
   MGMT       IPv6            no routing    Ma0
```

I think `incomplete` just means "routing protocol not enabled".  Since we don't want/need to run a routing protocol in a management VRF by default, I believe this is OK.

hellt:
thanks @bewing for this end-to-end PR

PR #942, opened 2022-07-15
Add Pseudo randomness to SRL chassis mac addresses

zenodhaene:
In Pull request #713 , deterministic MAC address for SRL nodes were introduced. This might give problems when interconnecting different containerlab projects, as the n-th nodes of each containerlab projects have the same chassis mac address. Specifically, FSS will refuse adding two nodes with the same MAC addresses.

This PR adds a pseudo-random, project-specific prefix to the mac address, while ensuring all previous requirements:
- bit 0 : strictly 0 (indicating unicast address)
- bit 1-11 : pseudo-random (avoid mac clashes when interconnecting projects)
- bit 12-24 : index of the node (for labs up to 4096 nodes)
- bit 25-47 : used by SRL: FF:00:<port> (ensures unique port mac addresses)

### DIFF
--- a/nodes/srl/srl.go
+++ b/nodes/srl/srl.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"text/template"
 	"time"
+	"math/rand"
 
 	"github.com/google/shlex"
 	"github.com/hairyhenderson/gomplate/v3"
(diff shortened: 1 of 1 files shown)


hellt:
maybe let's make it easier, @zenodhaene 
I would like to keep the ease of visual identification of SRL macs in the show commands outputs, for that reason it would be great to keep the most significant byte fixed.

Then we can use second byte to encode the index of the node. And third byte for random seed.

Of course, this means that we can only have up to 256 nodes in the same lab, but this seems to be well in the limits of what one server can bear, we didn't have labs bigger than that. If we ever come across those, we then can introduce another flag that will trade mac identification for scale.

Here is what I propose to use
https://go.dev/play/p/XV2JFbf1CQ3

zenodhaene:
@hellt Thanks!

hellt:
@zenodhaene Let me know if this works for your use case and then I will merge this

zenodhaene:
@hellt Just tested it out, it works great for my usecase: FSS seems to treat the mac address as (part of) the primary key, but does not check it afterwards. Destroying and redeploying the lab changes the chassis mac addresses in FSS, but FSS does not double-check this and everything keeps working.

PR #945, opened 2022-07-20
Go 1.18 & net/netip

kellerza:
`inet.af/netaddr` was accepted into the go standard library as `net/netip`

I would like to update kellerza/template to use net/netip (ideally without creating a v2 of the package), which would need go 1.18 in containerlab


hellt:
Thanks @kellerza 
Good stuff.

PR #949, opened 2022-07-21
Restore the slice template function

kellerza:
Gomplate override the built-in slice function from text/template.

This PR deletes the `slice` shortcut provided by gomplate. Use `coll.Slice`

Built-in
```
slice
	slice returns the result of slicing its first argument by the
	remaining arguments. Thus "slice x 1 2" is, in Go syntax, x[1:2],
	while "slice x" is x[:], "slice x 1" is x[1:], and "slice x 1 2 3"
	is x[1:2:3]. The first argument must be a string, slice, or array.
```

gomplate's [slice](https://docs.gomplate.ca/functions/coll/#coll-slice)
```
Creates a slice (like an array or list). Useful when needing to range over a bunch of variables
```

### DIFF
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -70,7 +70,6 @@ func RenderAll(allnodes map[string]*NodeConfig) error {
 
 	// gomplate overrides the built-in *slice* function. Rename it to *list*
 	gfuncs := gomplate.CreateFuncs(context.Background(), new(data.Data))
-	gfuncs["list"] = gfuncs["slice"]
 	delete(gfuncs, "slice")
 
 	tmpl := template.New("").Funcs(gfuncs).Funcs(jT.Funcs)
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -68,7 +68,7 @@ func RenderAll(allnodes map[string]*NodeConfig) error {
 		log.Infof("No template names specified (-l) using: %s", strings.Join(TemplateNames, ", "))
 	}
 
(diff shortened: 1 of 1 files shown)


hellt:
Let's not shadow names of gomplate since this will make it hard to use docs. Instead users of the template authors should use the fqdn name which is coll.Slice

hellt:
Ah, I see. The issue is that text/template slice is shadowed by gomplate... Let me think a bit more

clab/config/template.go
```diff
@@ -68,9 +68,12 @@ func RenderAll(allnodes map[string]*NodeConfig) error {
 		log.Infof("No template names specified (-l) using: %s", strings.Join(TemplateNames, ", "))
 	}
 
-	tmpl := template.New("").
-		Funcs(jT.Funcs).
-		Funcs(gomplate.CreateFuncs(context.Background(), new(data.Data)))
+	// gomplate overrides the built-in *slice* function. Rename it to *list*
+	gfuncs := gomplate.CreateFuncs(context.Background(), new(data.Data))
+	gfuncs["list"] = gfuncs["slice"]
```
hellt, line 73: If we delete the slice key from gfuncs, I suppose we still can use the fqdn version coll.Slice? If yes, then I wouldn't even bother with making a list alias, as I am advocating for using fully scoped names

kellerza: :+1: since the fully scoped name is still available, I'll go for the delete only

PR #958, opened 2022-07-30
log to stderr and configurable levels

### LINKED ISSUE #956: No way to disable logging, especially when using JSON output
Hi,

Trying to use containerlab 0.29 with "inspect -f json", I always get INFO lines in stdout, which prevents it being parseable as JSON, which contradicts the apparent intent of "-f json" option.
And more generally, a way to configure console output to only print errors would be useful, as otherwise they'll get lost among noisy info output, which also trains user to always ignore it (as is sensible, it being noise in 99.99% cases).

- What I did: `clontainerlab i --all -f json` with no labs running.
- Expected output: `[]` or nothing.
- Actual output: `INFO[0000] no containers found`, which is not valid JSON.
- Steps to reproduce: run `clontainerlab inspect -f json` - always prints INFO lines, which are never valid JSON.
- Setup info: containerlab 0.29 (c0055a03) installed via official docker container.

Thanks.

hellt:
fix #956

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


steiler:
Maybe use this logrus function
https://github.com/sirupsen/logrus/blob/master/logrus.go#L25

PR #959, opened 2022-08-01
introducing env var CLAB_VERSION_CHECK

steiler:
CLAB_VERSION_CHECK can be set to 'disable' to prevent the version checks from running and printing a new version notification

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -103,7 +103,16 @@ func deployFn(_ *cobra.Command, _ []string) error {
 
 	// latest version channel
 	vCh := make(chan string)
-	go getLatestVersion(vCh)
+
+	// check if new_version_notification is meant to be disabled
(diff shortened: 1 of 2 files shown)


mk-fg:
Maybe documenting it via something like this - https://github.com/mk-fg/containerlab/commit/c510b5eccd52d64d7b143f300fc8fe837e242d0a - can also be a good idea?
So that it won't be a completely hidden flag.
I think usual place to look for such "Environment Variables" section is the tail of manpage, and that md file seem to be a closest thing to it.

(if it indeed sounds useful, you can probably add this commit here via `git remote add temp-pr959-doc https://github.com/mk-fg/containerlab/ && git fetch temp-pr959-doc && git cherry-pick c510b5ec`)

hellt:
yeah, I am good with adding this env var doc piece at the proposed place

steiler:
Alright, doc addded, ready to merge I guess @hellt

PR #966, opened 2022-08-04
implementing netmode external-container

steiler:
enriched with checks for container existance

PR #969, opened 2022-08-09
support for external containers in `network-mode`

hellt:
supersedes #966

Hi @steiler 

what do you think about this approach, where a single `container:<cont-name>` instruction is used both for clab and non-clab managed containers.

Mainly I didn't want to introduce another string-based config param (external-container), and proposing to use a single `container:<cont-name>` instead, as it makes (I hope) things simplier for users.

One thing I changed here is the way we use container ID in filter string, before it was just `cID`, but Docker does not use exact matches for that value, instead it uses substring matches. To fix that I used a regexp-based string value to make sure that when we use `GetContainer` we are able to have exact match by the provided name

### DIFF
--- a/docs/manual/nodes.md
+++ b/docs/manual/nodes.md
@@ -361,9 +361,14 @@ topology:
 ```
 
 ### network-mode
+
 By default containerlab nodes use bridge-mode driver - nodes are created with their first interface connected to a docker network (management network).
 
-It is possible to override this behavior and set the network mode to the value of `host`.
+It is possible to change this behavior using `network-mode` property of a node.
+
+#### host mode
+
+The `network-mode` configuration option set to `host` will launch the node in the [host networking mode](https://docs.docker.com/network/host/).
 
 ```yaml
 # example node definition with host networking mode
@@ -372,7 +377,7 @@ my-node:
   network-mode: host
 ```
 
-The `network-mode` configuration option set to `host` will launch the node in the [host networking mode](https://docs.docker.com/network/host/).
+#### container mode
 
 Additionally, a node can join network namespace of another container - by referencing the node in the format of `container:parent_node_name`[^2]:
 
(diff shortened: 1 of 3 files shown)


LimeHat:
@hellt what would happen if a user has a container named "lab-1" in a clab topology & as an external entity at the same time?
It seems that the latter will take precedence, but this might not be intended or desired?

hellt:
@LimeHat good call. I have changed the logic to check for internal containers first.
the mandatory prereq for making clab-scheduled container to be picked up is to use a startup-delay for a donor container, e.g.

```yaml
topology:
  nodes:
    donor:
      kind: linux
      image: alpine:3
    node1:
      kind: linux
      network-mode: container:donor
      startup-delay: 2
      image: alpine:3
```

as otherwise donor might not have enough time to deploy

steiler:
What if we would provide, as a pre-step to any deploy action some struct to the runtime, that it knows which containers are part of the topology. This struct also holds a status, which is managed by the runtime itself. And it could go and do dependency checking itself, start the container when all the internal dependencies are finally started and therby you'd get rid of the sleep timer / start delay foo.
first thought about injecting that list into the container start, but there we are too deep already in the deployment of just a single container, but finally pushing that into the runtime seems to even make more sense to me.
Thought?

hellt:
@steiler I like the idea, and had a thought about it when looking at this PR. But I am not so sure it is runtime business.
I think maybe it is better to look at scheduling function that we have rudimentary support for.

For instance, right now we have two queues - static IP and dynamic IP nodes which are scheduled one after another to guarantee, that nodes with static IPs are created first, and then dyn IP nodes go in.

https://github.com/srl-labs/containerlab/blob/main/clab/clab.go#L165-L192

and it looks like it would be great to have a more feature rich scheduler package that will take care of all the dependencies. Maybe even creating a graph of nodes that we can use for deployments...


But we got carried away here :D
@steiler If you don't have objections to this current impl, I am willing to merge it and carve out a release.

steiler:
Interesting, first time I hear of that scheduler.
Yes, it sounds like it belongs even there.

Anyways this PR seems fine so far.

PR #974, opened 2022-08-11
Don't skip all template errors

kellerza:
- Only skip template files when the glob pattern matches no files, else fail on the template error.
- If no template is found (intentional), log as debug level and not as warning

### DIFF
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -40,6 +40,10 @@ func LoadTemplates(tmpl *template.Template, role string) error {
 		fn := filepath.Join(p, fmt.Sprintf("*__%s.tmpl", role))
 		_, err := tmpl.ParseGlob(fn)
 		if err != nil {
+			if strings.Contains(err.Error(), "pattern matches no file") {
+				log.Debug(err)
+				return nil
+			}
 			return fmt.Errorf("could not load templates from %s: %w", fn, err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


clab/config/template.go
```diff
@@ -40,6 +40,10 @@ func LoadTemplates(tmpl *template.Template, role string) error {
 		fn := filepath.Join(p, fmt.Sprintf("*__%s.tmpl", role))
 		_, err := tmpl.ParseGlob(fn)
 		if err != nil {
+			if strings.Contains(err.Error(), "pattern matches no file") {
+				log.Debug(err)
+				return nil
+			}
```
hellt, line 46: Hi @kellerza 
can you remind me what is the idea here? Why we want to let this error slip?

PS> If this is valid reason, lets look if ParseGlob uses a certain error for it, so we can use `errors.Is()` instead of checking on string value of an error

kellerza: > can you remind me what is the idea here? Why we want to let this error slip?

We parse template roles using wildcards (i.e. `*__ce.tmpl` and `*__srl.tmpl`). 
Certain roles might not have templates defined (i.e. `bgp__srl.tmpl`, but there are no `bgp__ce.tmpl` in the directory)

Initially we failed on these errors and empty template files were required.

A feature was added to not require such empty templates [here](https://github.com/srl-labs/containerlab/commit/c008463ae3b201a62449ceb079246a032b191628#diff-ac45aaa99ba2c73d9e9799f88ff0d26d0cde7d54d0fe690e60d2c8f21ebdf90dR78). Unfortunately this also introduced the regressions that you can have a template file present that is unable to load. Example is an invalid template function name (these functions are resolved on loading templates, not Execution).

**In summary this PR fails with Invalid templates and hides unavailable templates**

> PS> If this is valid reason, lets look if ParseGlob uses a certain error for it, so we can use errors.Is() instead of checking on string value of an error

[text/template] only exposes an [evaluation error](https://pkg.go.dev/text/template#ExecError) and the go source is unavailable at present https://cs.opensource.google/go/go

clab/config/template.go
```diff
@@ -79,13 +83,11 @@ func RenderAll(allnodes map[string]*NodeConfig) error {
...
 				if err != nil {
-					log.Warnf("Unable to load template %s (%+v); skipping", tmplN, err)
-					continue
+					return err
 				}
 				l = tmpl.Lookup(tmplN)
-				log.Debugf("Got a lookup result %+v (of type %T)", l, l)
 				if l == nil {
-					log.Warnf("No template found for %s; skipping..", nc.TargetNode.ShortName)
+					log.Debugf("No template found for %s; skipping..", nc.TargetNode.ShortName)
```
kellerza, line 90: If you do want to show unavailable template, we should only do it here. Probably rather with a Infof, since this is very much desired behavior

PR #976, opened 2022-08-12
Scheduler with waitgroups per node for flexible dependency resolution

steiler:
The basic idea I have is, every node has its waitgroup .... if it is dependent on other nodes, these dependent nodes will add to the waitgroup, and this incrementation for that specifc node is stored in an array. 
when th node is done, it will issue a wg.Done() for all depending nodes.  The creation of all nodes in started in parallel, but with a pre-function that waits for the individual waitgroup to be decremented to 0 .... then this is send of to into the channel towards the workfunc  which creates the nodes
static and dynamic ip nodes fit perfectly into this schema ... all every static node increments the waitgroup count of all the dynamic ... and every static node after its creation decrements these counters again...


steiler:
BTW: I might have messed up something with the serialNodes stuff.
However if something should be started just serial, linking one container as a dependency to the next is a peace of cake.
Please someone with deeper understanding of that doublecheck. THX

Issue #973, opened 2022-08-11
Not all VM kinds support startup-config


nlgotz:
From a testing/usability perspective, it would be awesome if all of the VM kinds would support a startup config.

Going through the code briefly, here's where it looks like is currently supported or not supported

| Kind                      | startup-config Support |
|---------------------------|---------|
| Nokia SROS                | Yes     |
| Arista vEOS               | No     |
| Juniper vMX               | Partial     |
| Juniper vQFX              | Yes     |
| Cisco XRv                 | No     |
| Cisco XRv9k             | No |
| Cisco CSR                 | No     |
| Cisco Nexus                | No     |
| Cisco Nexus 9000v         | No     |
| Dell FTOS10v              | No     |
| Mikrotik RouterOS         | Yes     |
| Palo Alto PAN             | No     |
| IPInfusion OcNOS          | No     |
| Checkpoint Cloudguard     | No     |

From looking at the code, it looks like vMX has support in Containerlab (doesn't look like it in hellt/vrnetlab).

Things Needed for Containerlab:

- [ ] Code for all "No" nodes
- [ ] Update Documentation

This won't fully allow all nodes to load a startup config as changes will need to be done on the hellt/vrnetlab and/or boxen end to actually do something with the config file

nlgotz:
This is something I can take on/be assigned.

I see this as being 1 PR to add the code and then separate PRs to update the documentation once the vrnetlab/boxen portion gets updated.

hellt:
Hi @nlgotz 
currently most (like 90%) of non-sros vrnetlab images uses the following technique to deliver the stratup-config functionality:

1. containrelab refers a file which contains CLI based lines
2. vrnetlab code opens this file and pushes it line by line when the node applies it default config (like [here](https://github.com/hellt/vrnetlab/blob/master/vqfx/docker/launch.py#L156-L178))

are you planning on using this one?

nlgotz:
Yep. I was planning on basically copy/pasting how it's done in the vmx/vqfx go files and then in vrnetlab loading it line by line.

I may redo the RouterOS one as well so that it uses the same method and doesn't require tftp or anything like that to load configs. But that'd be lower on my priority list as it's already working

hellt:
ok, cool
on clab side that can be a single PR (with docs) after the relevant parts are merged into vrnetlab
you will be able to create a clab binary with those changes without releasing it, so you can test vrnetlab part

nlgotz:
[...]

PR #977, opened 2022-08-12
#973 startup config

nlgotz:
Fixes most of the nodes in #973 

I don't have access to all the node images, so I don't want to inadvertently break something without being able to test.

This adds support for CSR, NXOS, N9KV, VEOS, XRV, and XRV9K to use and run the startup-config option.


hellt:
thanks @nlgotz ! Great addition

PR #984, opened 2022-08-25
Add support for dependabot

gaby:
* Support for tracking `gomod` updates.
* Support for tracking `github-actions` updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  -   package-ecosystem: "github-actions"
+      directory: "/"
+      schedule:
+          interval: daily
+  -   package-ecosystem: "gomod"
+      directory: "/"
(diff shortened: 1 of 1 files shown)


gaby:
@hellt This should help automate some of the maintenance tasks.

hellt:
Hi @gaby 
I am hesitating...
Although keeping up-to-date deps is generally good, the amount of deps we have is considerable. Having critical deps on something like `docker` may lead to undesired bumps where the client gets updated to a version that requires a newer docker installation, and our testing pipeline may miss this...

Nevertheless, I am curious to try it out, but I would like to ask you to change the checking interval to weekly instead of daily

gaby:
@hellt Changes made. By default dependabot only tracks, and auto-creates a PR for each new update. It does not auto-merge, or auto change your git repo. It still requires the PR to run it's and get accepted by someone.

hellt:
Yes, sure. Let's experiment

gaby:
Awesome! It already created quite a few PR's and is running tests for each one :-)

PR #998, opened 2022-08-27
Document missing --log-level and --vars global options

mk-fg:
Similar to https://github.com/srl-labs/containerlab/pull/986 but for `--vars` and recently added `--log-level` option (in https://github.com/srl-labs/containerlab/pull/958).

Realized that `--log-level` was missing in docs when #986 was merged, and thought to add missing `--vars` too, while at it.
Especially since latter seem to have non-obvious logic for looking up default files, which now can be documented here in the option description.

Same idea as with #986 wrt placement for this documentation - can likely be improved if there'd be significantly more of such global stuff to affect tools' behavior.

Don't think there are any other undocumented global options or CLAB_ env variables for commands.

### DIFF
--- a/docs/cmd/deploy.md
+++ b/docs/cmd/deploy.md
@@ -22,6 +22,15 @@ When the topology file flag is omitted, containerlab will try to find the matchi
 
 With the global `--name | -n` flag a user sets a lab name. This value will override the lab name value passed in the topology definition file.
 
+#### vars
+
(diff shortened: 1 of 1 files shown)


hellt:
🙏 @mk-fg 
That's a good addition. As for the other global flags you can check `clab -h` output

mk-fg:
Yeah, I did check it, meant to say that there doesn't seem to be any other global ones.

PR #1013, opened 2022-09-05
Shared namespace dependency

steiler:
On top of #976 this implements the dependencies for the "network-mode: container:<CONTAINERNAME>" based dependency.


steiler:
@YutaroHayakawa, maybe you want to try this feature and provide feedback.
Prior to starting the containers you should see information about the dependencies and you should be able to get rid of any start-delays or whatever you might have used to make the shared namespace dependency work.
Let us know if this works for you.
Thanks!

YutaroHayakawa:
Hi @steiler, thanks for your heads up. This is a very cool feature!

In our case, containerlab containers depend on the external Kind containers. Thus, having this kind of dependency warning is useful, I think.

If I add one more suggestion, I would say it would be much nicer if we had a feature to "wait" for such a dependency container. So that we don't have to serialize the Kind deployment and Containerlab deployment then it speeds up the total deployment time I think.

YutaroHayakawa:
It detects cyclic dependency. Nice...

```
name: dependency-manager-demo
topology:
  nodes:
    n0:
      kind: linux
      image: nicolaka/netshoot:latest
      network-mode: container:n1
      cmd: tail -f
    n1:
      kind: linux
      image: nicolaka/netshoot:latest
      network-mode: container:n2
      cmd: tail -f
    n2:
      kind: linux
      image: nicolaka/netshoot:latest
      network-mode: container:n0
      cmd: tail -f
  links:
  - endpoints: ["n0:net0", "n1:net0"]
```

```
INFO[0000] Parsing & checking topology file: topo.yaml
INFO[0000] Creating lab directory: /home/yutaro/WorkSpace/containerlab/clab-dependency-manager-demo
Error: cyclic dependencies found!
n0 -> [ n1 ]
n1 -> [ n2 ]
n2 -> [ n0 ]
```

Makefile
```diff
@@ -19,6 +19,17 @@ build-with-podman:
 test:
 	go test -race ./... -v
 
+MOCKDIR = ./mocks
+.PHONY: mocks-gen
+mocks-gen: mocks-rm ## Generate mocks for all the defined interfaces.
+	go install github.com/golang/mock/mockgen@latest
```
hellt, line 25: lets fix the mockgen version to the version of the mock package as recommended by the mock docs

steiler: alright, done!

PR #1029, opened 2022-09-20
docs/cmd/exec - Add example of `--label` flag for exec command

alexandrehassan:
It isn't quickly obvious how to format the --label command to target a single node. The only example was in the robot tests.


hellt:
thanks!

PR #1032, opened 2022-09-21
Docker `overlay` network driver

hellt:
a WIP stage for docker' overlay network driver
everyone is free to check it out and provide feedback

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -139,7 +139,6 @@ func NewContainerLab(opts ...ClabOption) (*CLab, error) {
 func (c *CLab) initMgmtNetwork() error {
 	log.Debugf("method initMgmtNetwork was called mgmt params %+v", c.Config.Mgmt)
 
-	//TODO DUMAIS: this is temporary, we need to make this configurable
 	if c.Config.Mgmt.Type == types.MgmtNet_Type_Undefined {
 		c.Config.Mgmt.Type = types.MgmtNet_Type_Bridge
 	}
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -138,8 +138,18 @@ func NewContainerLab(opts ...ClabOption) (*CLab, error) {
 // initMgmtNetwork sets management network config.
 func (c *CLab) initMgmtNetwork() error {
 	log.Debugf("method initMgmtNetwork was called mgmt params %+v", c.Config.Mgmt)
+
+	//TODO DUMAIS: this is temporary, we need to make this configurable
+	if c.Config.Mgmt.Type == types.MgmtNet_Type_Undefined {
+		c.Config.Mgmt.Type = types.MgmtNet_Type_Bridge
+	}
+
 	if c.Config.Mgmt.Network == "" {
-		c.Config.Mgmt.Network = dockerNetName
+		if c.Config.Mgmt.Type == types.MgmtNet_Type_Overlay {
+			c.Config.Mgmt.Network = dockerNetName+"-overlay"
+		} else {
+			c.Config.Mgmt.Network = dockerNetName
+		}
 	}
 
 	if c.Config.Mgmt.IPv4Subnet == "" && c.Config.Mgmt.IPv6Subnet == "" {
--- a/runtime/docker/docker.go
+++ b/runtime/docker/docker.go
@@ -90,6 +90,7 @@ func (d *DockerRuntime) WithConfig(cfg *runtime.RuntimeConfig) {
 
 func (d *DockerRuntime) WithMgmtNet(n *types.MgmtNet) {
 	d.mgmt = n
+
 	// return if MTU value was set by a user via config file
 	if n.MTU != "" {
 		return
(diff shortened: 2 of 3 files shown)


hellt:
@pdumais can you add some notes as to how do you use this driver with an example?

pdumais:
The use case is this: we need to spin up a lab that might contain up to 1000 routers (or let's stick with 250 for now because this will hit other limitations of clab). So we need to spread this load across multiple host. Although it would be nice if containerlab could play nice with docker swarm or k8s, we can takr care of the orchestration part by deploying multiple labs separately with their own topo file.

But we still need each lab to sit in the same overlay network. So as a PoC, I created a swarm cluster of 2 hosts and spun up 2 clabs on each hosts. The containers are normal containers (not swarm-managed). But having a swarm lets me create "overlay" networks. Those are able to span across both hosts using a vxlan tunnel. I've modified clab to be able to use these overlay networks.

My initial tests indicates that everything is working fine. But I need guidance on what else might be needed since I don't know about other areas in clab that could be affected by this.

Also, I'm not sure how well it plays wit ipv6. Both labs have conflicting ipv6 addresses but I think that might be normal for link-local addresses. I'm definitely not an ipv6 expert.

steiler:
Are we talking pure mgmt network here? Because I seem to only see mgmt network stuff adjusted in this PR.
Not sure If we should really aim for 1000 nodes and more to be part of the same BD...or what am I missing here?

pdumais:
This is where I'm confused @steiler . It seems like containerlab create 1 single docker network and all nodes of all labs are connected to that network (clab). Unless I did it wrong, I noticed that creating 2 labs results in them using the "clab" docker network. That lab seems to be referred to as "mgmt" in the code. So I'm not clear if it has anything to do with networks created by the nodes.

karimra:
@pdumais it's possible to create a different mgmt network for each lab: https://containerlab.dev/manual/network/#configuring-management-network

pdumais:
> @pdumais it's possible to create a different mgmt network for each lab: https://containerlab.dev/manual/network/#configuring-management-network

@karimra Correct. And these changes seem to be compatible.

`host1# cat topo.yml 
name: meow

mgmt:
  network: awesome
  ipv4_subnet: 172.100.100.0/24
  type: overlay

topology:
  nodes:
    srl1:
      kind: srl
      image: ghcr.io/nokia/srlinux
    srl2:
      kind: srl
      image: ghcr.io/nokia/srlinux

  links:
    - endpoints: ["srl1:e1-1", "srl2:e1-1"]
    - endpoints: ["srl1:e1-2", "srl2:e1-2"]
    - endpoints: ["srl1:e1-3", "srl2:e1-3"]
    
host1# containerlab deploy -t topo.yml 
INFO[0000] Parsing & checking topology file: topo.yml   
Error: containers ["clab-srlceos01-srl1" "clab-srlceos01-srl2"] already exist. Add '--reconfigure' flag to the deploy command to first remove the containers and then deploy the lab
root@dumais-cnv-1:~/test2# vi topo.yml 
root@dumais-cnv-1:~/test2# ../test/containerlab/bin/containerlab deploy -t topo.yml 
INFO[0000] Parsing & checking topology file: topo.yml   
WARN[0000] it appears that container host has low memory available: ~0Gi. This might lead to runtime errors. Consider freeing up more memory. 
INFO[0000] Creating lab directory: /root/test2/clab-meow 
INFO[0000] Creating docker network: Name="awesome", IPv4Subnet="172.100.100.0/24", IPv6Subnet="", MTU="1500" 
INFO[0000] Creating container: "srl1"                   
INFO[0000] Creating container: "srl2"                   
INFO[0002] Creating virtual wire: srl1:e1-1 <--> srl2:e1-1 
INFO[0002] Creating virtual wire: srl1:e1-3 <--> srl2:e1-3 
INFO[0002] Creating virtual wire: srl1:e1-2 <--> srl2:e1-2 
INFO[0002] Running postdeploy actions for Nokia SR Linux 'srl2' node 
INFO[0002] Running postdeploy actions for Nokia SR Linux 'srl1' node 
INFO[0030] Adding containerlab host entries to /etc/hosts file 
INFO[0030] 🎉 New containerlab version 0.31.1 is available! Release notes: https://containerlab.dev/rn/0.31/#0311
+---+----------------+--------------+-----------------------+------+---------+------------------+--------------+
| # |      Name      | Container ID |         Image         | Kind |  State  |   IPv4 Address   | IPv6 Address |
+---+----------------+--------------+-----------------------+------+---------+------------------+--------------+
| 1 | clab-meow-srl1 | 231a05809294 | ghcr.io/nokia/srlinux | srl  | running | 172.100.100.2/24 | N/A          |
| 2 | clab-meow-srl2 | c34252920c1e | ghcr.io/nokia/srlinux | srl  | running | 172.100.100.4/24 | N/A          |
+---+----------------+--------------+-----------------------+------+---------+------------------+--------------+

`

On the 2nd host, I can see the new network "magically" appear:
`host2# docker network ls
NETWORK ID     NAME              DRIVER    SCOPE
re807snwyt7o   awesome           overlay   swarm
bbb9598bb12a   bridge            bridge    local
fd9df7d45975   docker_gwbridge   bridge    local
f864f9150de1   host              host      local
0bc8pvimr43i   ingress           overlay   swarm
a6adb479cb34   none              null      local
`

And I can ping across to srl1:
`host2# docker run --rm -ti --network awesome centos bash -c "ping 172.100.100.2"
PING 172.100.100.2 (172.100.100.2) 56(84) bytes of data.
64 bytes from 172.100.100.2: icmp_seq=1 ttl=64 time=29.8 ms
`

pdumais:
I now understand what you mean about the mgmt network.
I realize now that this does not solve the entire problem. I will also need the link to span across hosts. So creating a veth pair between netns won't cut it. I probably need a veth pair that has an end in a bridge on the host. That bridge would need to be vxlan-aware (openvswith bridge) or something similar.  So the solution would be to be able to do 
`
links:
    - endpoints: ["srl1:e1-1", "host-docker-network:my-swarm-overlay-network"]
`
or
`
links:
    - endpoints: ["srl1:e1-1", "host-bridge:my-OVS-vxlan-enabled-bridge"]
`

pdumais:
Everything I need to accomplish is possible with the OVS Kind. I think this PR is not necessary. Although using OVS requires a lot more config outside of containerlab, it would be nice to have all of this managed automatically.

hellt:
yep, I was under the impression that you contain your nodes on a per-node network and only mgmt network needs to be stretched across the nodes.

Let's park this then

PR #1037, opened 2022-09-27
Fixing Ignite issue regarding serial creation

steiler:


hellt:
Hi @networkop 
in case you have time would you be able to have a look at that PR?

For some reason, the cache that ignite consults with seems to be invalid as shown here - https://github.com/srl-labs/containerlab/issues/1039#issue-1388913471

What is even more funny is that it worked prior to 0.31.0, where in 0.31.0 we didn't have a sequential order of ignite containers start. It is reintroduced in this PR, but issues are still seen.

networkop:
Done a little debugging:
```
(dlv) print strictErrs
[]error len: 2, cap: 2, [
        *errors.errorString {
                s: "unknown field \"spec\"",},
        *errors.errorString {
                s: "unknown field \"status\"",},
]
```
Those errors come from k8s.io's apimachinery. The issue is that ignite and libgitops are built against apimachinery v0.18 or smth and clab now depends on 0.25. So as you see from this diff, it seems like the decoder is now a lot stricter and bails out when it sees unexpected fields.
https://github.com/kubernetes/apimachinery/compare/v0.25.0..v0.22.0#diff-620d6f1202d998ffe5b21e700ec1599b12a7364ef2b705b5be8496f1233a9781L163-L187


TBH, I'm not sure what the right way forward from here is. I can raise this issue in ignite repo but it may require a substantial re-work and time investment, which I doubt weave are prepared to make. Also clab may be one of the few projects that uses ignite via API, not via CLI, so don't know if they will even be willing to entertain such idea.

networkop:
I think there's an indirect dependency by other modules. e.g. `github.com/hairyhenderson/gomplate/v3` depends on `client-go  v0.25.2`, which, in turn, depends on `apimachinery 0.25.1`

networkop:
there are a few others like containerd

networkop:
sadly, these things are pretty common for k8s.io packages, e.g. https://github.com/kubernetes/kubernetes/issues/88472

hellt:
Ah, indeed. Well, then let me mark ignite support as "broken" for now in the docs and make it non default for cvx nodes in the next release. Then over time if ignite lives they will probably have a chance to fix it

networkop:
Yep, probably worth mentioning that older version of clab would still work.

steiler:
What do we do about this PR, which already handles the ignite nodes and creates dependencies that make them start one after the other?!
Do we close it for now? Or do we want to merge is, since we feel that this is only temporarely broken and will probably be fixed somewhen soon?

hellt:
> What do we do about this PR, which already handles the ignite nodes and creates dependencies that make them start one after the other?! Do we close it for now? Or do we want to merge is, since we feel that this is only temporarely broken and will probably be fixed somewhen soon?

Let's merge this, as the slight chance that this gets fixed upstream still exists =)

networkop:
there's a bit of cleanup you can do before merging this. In deploy.go and clab.go remove `serialNodes` variable as it's no longer used.

hellt:
@steiler in #1051 I have changed default runtime to docker for cvx nodes

steiler:
I've removed the serialNodes, thanks for pointing that out @networkop
So I think this can be merged now.

I recognized that destroy also contains serialNodes. Question is, if we want to use DM there as well (we should open a new issue for that) or if we leave that part as is.

steiler:
> Done a little debugging:
> 
> ```
> (dlv) print strictErrs
> []error len: 2, cap: 2, [
>         *errors.errorString {
>                 s: "unknown field \"spec\"",},
>         *errors.errorString {
>                 s: "unknown field \"status\"",},
> ]
> ```
> 
> Those errors come from k8s.io's apimachinery. The issue is that ignite and libgitops are built against apimachinery v0.18 or smth and clab now depends on 0.25. So as you see from this diff, it seems like the decoder is now a lot stricter and bails out when it sees unexpected fields (search for `func (s *Serializer) Decode`) https://github.com/kubernetes/apimachinery/compare/v0.25.0..v0.22.0#diff-620d6f1202d998ffe5b21e700ec1599b12a7364ef2b705b5be8496f1233a9781L163-L187
> 
> TBH, I'm not sure what the right way forward from here is. I can raise this issue in ignite repo but it may require a substantial re-work and time investment, which I doubt weave are prepared to make. Also clab may be one of the few projects that uses ignite via API, not via CLI, so don't know if they will even be willing to entertain such idea.

The issue is, that the apimachinery functions are used, but the object handed into the decodeInto function is this
https://github.com/weaveworks/libgitops/blob/v0.0.2/pkg/runtime/meta.go#L16-L19
instead of the runtime.Object as defined in the apimachinery. And obviously the spec and status definitions are missing with this APIType struct.

Here it is created and handed over to the decodeInto function.
https://github.com/weaveworks/libgitops/blob/v0.0.2/pkg/storage/storage.go#L244-L257

Well nothing that will help much now, but the issue seems with weaveworks packages rather then apimachinery.

hellt:
thanks @steiler , @networkop 
I am merging this and hope ignite will get their assessment if this can be fixed

Issue #1055, opened 2022-10-07
Linux containers not cleaned up after VM reboot


jbemmel:
Suspected cause: https://github.com/srl-labs/containerlab/blob/2d6d77baf9b7c9b15cfdc93c19f74bbf37679a20/runtime/docker/docker.go#L411

When the host system reboots, most containers are removed but a set of Linux containers remains running in isolation

hellt:
As you can see, the reason is in the comment section. Some management
containers (like telemetry stack) may fail over time and the automatic
restart beings them back without human intervention.

Rebooting a node with a lab running is considered to be an incident, and a
manual clean up is expected. Luckily it is a single line bash script

jbemmel:
I would still like to be able to disable this through configuration. When building automated topologies with several simple linux host containers, I prefer them to be cleaned up like the other non-linux containers, and not stick around

Suggestion:
 auto-restart: False (default True)

hellt:
thanks

PR #1056, opened 2022-10-07
Add auto-remove=yes option to avoid auto-restart policy and cleanup containers automatically

jbemmel:
See https://github.com/srl-labs/containerlab/issues/1055

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -230,6 +230,7 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 		CPUSet:          c.Config.Topology.GetNodeCPUSet(nodeName),
 		Memory:          c.Config.Topology.GetNodeMemory(nodeName),
 		StartupDelay:    c.Config.Topology.GetNodeStartupDelay(nodeName),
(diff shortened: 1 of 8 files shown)


hellt:
thanks @jbemmel 
let's name it `restart-policy` to align with the terms docker/k8s use for containers https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy

hellt:
thanks @jbemmel 
I have changed the type to bool as this seems to be inline with the flag value of the docker CLI
Will add a test and merge this

jbemmel:
> thanks @jbemmel
> I have changed the type to bool as this seems to be inline with the flag value of the docker CLI
> Will add a test and merge this

I started with a bool, but the problem is that if you set it to ‘false’ the logic drops down to the next level. A bool cannot distinguish between “not set” and “set to false”

hellt:
@jbemmel yes, initially I decided to not cater for the use case where a true set on kind level can be overridden by `false` set on node level, but it makes sense to do it right from the get go

this PR uses pointers to address that https://github.com/srl-labs/containerlab/pull/1065
anyhow, in the today's release 0.32.2 you should be able to start using this feature with a caveat you know about

PR #1061, opened 2022-10-14
Config Engine: Skip template-paths without templates for a role & expose clab_type

kellerza:
When we search for templates in a list of template-paths the current implementation (incorrectly) stops searching when there are no templates available for a role in a path.

This PR allows the code to continue searching through the remainder of the template-paths (the code already prints a debug message)

Today, the workaround is to create empty/dummy templates

The second part of the PR exposes the node type as `clab_type` in the config engine variables


hellt:
thnx merged

PR #1065, opened 2022-10-16
honor auto-remove bool propagation

hellt:


hellt:
@jbemmel would you be willing to test this PR with a use case that you have?

```
docker run --rm -v $(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 pull ghcr.io/srl-labs/clab-oci:8350cf6c
```

PR #1083, opened 2022-11-01
fix ssse3 instruction check

hellt:


types/types.go
```diff
@@ -147,7 +147,7 @@ type NodeConfig struct {
 }
 
 type HostRequirements struct {
-	SSE3         bool `json:"sse3,omitempty"`          // sse3 cpu instruction
+	SSSE3        bool `json:"sse3,omitempty"`          // ssse3 cpu instruction
```
jbemmel, line 150: json attribute left as 'sse3' - on purpose?

hellt: nope. needs to be fixed

PR #1097, opened 2022-11-16
Fix nextui node tooltip without image field

gamerslouis:
For non container nodes, image filed is not set.
Without this, the nx graph node tooltip image field in value colume will be skipped. Cause the filed name and value mismatch.
I set the image field to  N/A to indicate no image property to fix this problem.

Before fix:
image

After fix:
image


hellt:
thanks @gamerslouis ! It is amazing to see you fixing JS code of the graph engine as I am myself has very little knowledge of JS

PR #1098, opened 2022-11-16
Fix non-container nodes display in web graph

### LINKED ISSUE #1078: when define linux bridge in clab.yaml, use graph command, web show is nil.
When I define a bridge simulate a switch like this.
I use `clab graph -t clab.yaml`, but web not show the switch.
Screen Shot 2022-10-24 at 10 39 36

Screen Shot 2022-10-24 at 10 40 41

gamerslouis:
Without offline mode, the graph command only collect container nodes. So nodes such as bridge are not displayed in the graph web view. This commit fix this problem by building non-container node data with offline method

Close #1078


hellt:
great, thanks @gamerslouis! this is awesome

PR #1099, opened 2022-11-18
Default node embedding

steiler:
Using the embedded DefaultNode to implement the std, behaviour which will also allow for specific implementations to overwrite the behaviour.


hellt (approved):
great stuff @steiler 
PRs that have more deleted lines than added ones are X2

PR #1100, opened 2022-11-18
allowing for external containers to be referenced in the link configuration

steiler:


hellt:
can I review this before #1099 ?

steiler:
Dropping this idea of implementing external container links for a seperate external kind.

PR #1108, opened 2022-11-23
Rate-limit polling for node status at link creation time

steiler:
Creating for links busy-waits until the nodes on both sides of a link are created.
This is just to introduce some sleep time.

Idealy we woudl integrate this into the DependencyManager as well. That would need some rework to use general interfaces instead of just working with node names...

Maybe we open a new issue for that!? But go with this workaround for now,

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -501,7 +501,9 @@ func (c *CLab) CreateLinks(ctx context.Context, workers uint) {
 			}
 			c.m.Unlock()
 		}
-		time.Sleep(time.Duration(500) * time.Millisecond)
+
+		// prevent clab from throttle CPU when nodes take considerable time to start
+		time.Sleep(500 * time.Millisecond)
 	}
 
 	// close channel to terminate the workers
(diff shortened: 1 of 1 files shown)


hellt:
How was the busy polling affecting you?

steiler:
Clab is eating a CPU while deploying nodes that take some time, in my case the k8s-kind. It takes 30sek to 90sek with a lot of cpu to bootstrap a kind cluster with some nodes... if clab is already consuming a cpu for itself, for just polling ... thats not nice

clab/clab.go
```diff
@@ -501,6 +501,7 @@ func (c *CLab) CreateLinks(ctx context.Context, workers uint) {
 			}
 			c.m.Unlock()
 		}
+		time.Sleep(time.Duration(500) * time.Millisecond)
```
hellt, line 504:
```suggestion

		// prevent clab from throttle CPU when nodes take considerable time to start
		time.Sleep(500 * time.Millisecond)
```

steiler: yepp looking fine... did not look up how to properly construct it... my bad!

Issue #1114, opened 2022-11-30
CoreOS install script not working


steiler:
Trying to install containerlab via the script at `https://get.containerlab.dev` fails with the following error on CoreOS:

```
[root@localhost core]# bash -c "$(curl -sL https://get.containerlab.dev)"
Downloading https://github.com/srl-labs/containerlab/releases/download/v0.33.0/containerlab_0.33.0_linux_amd64.rpm
Preparing to install containerlab 0.33.0 from package
error: can't create transaction lock on /usr/share/rpm/.rpm.lock (Read-only file system)
Failed to install containerlab
        For support, go to https://github.com/srl-labs/containerlab/issues
```
It seems like on CoreOS it should be installed via `rpm-ostree install <CLAB.RPM>`

hellt:
can you check if coreos reveals itself in `cat /etc/os-release` file or in `uname -a`

steiler:
```
[core@localhost ~]$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="37.20221106.3.0 (CoreOS)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora CoreOS 37.20221106.3.0"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
HOME_URL="https://getfedora.org/coreos/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora-coreos/"
SUPPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
BUG_REPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
VARIANT="CoreOS"
VARIANT_ID=coreos
OSTREE_VERSION='37.20221106.3.0'
[core@localhost ~]$ uname -a
Linux localhost.localdomain 6.0.7-301.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Nov 4 18:35:48 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

[...]

PR #1115, opened 2022-11-30
Adjust installer to Fedora CoreOS

steiler:
As discussed in #1114, adjustments to the `get.sh` installer script to also work on Fedora CoreOS.


hellt:
thanks @steiler

Issue #1060, opened 2022-10-13
Generalize checkIfSignatures to all kinds


jbemmel:
https://github.com/srl-labs/containerlab/blob/1ef7460784a0ee4d5358221bac9c7563da704215/clab/config.go#L681

SRL nodes aren't the only ones with specific naming requirements, vrnetlab based nodes like sros need to use 'eth[x]'

The correct interface naming pattern should be part of the kind/type definition, and (if set) should be checked generically for all node kinds

Note: was discussed in the past (https://github.com/srl-labs/containerlab/issues/377) but only partially implemented. Some users had issues with using invalid interface names, which silently breaks their topology

steiler:
Moved to #1129 now

hellt:
the ground work is done in #1129 
we have to add a regexp for vr-based nodes and then we can close this one

PR #1119, opened 2022-12-06
Extend node interface

steiler:
Delegate node related actions to the nodes.Node interface


steiler:
This allows for the checking of kind based interface naming conventions prior to deploying a topology.
It is a request raised in #1060.

On the different kinds you just need to implement the following function.
```
func (n *NODE_KIND) CheckInterfaceNamingConvention() error{
}
```

hellt:
replaced by #1129

PR #1121, opened 2022-12-07
K8skind

steiler:
implementing k8s kind


steiler:
@aojea so for your reference, the deployment of kind clusters works fine, also the wireing part, for which we use the indirection of ext-container (which can basically be any external container). However we're mainly facing issues when executing containerlab with just a lab name (which is then on an already deployed topology) therefore we're relying on a label we attach to the containers to relate them to the lab with the given name. This is basically where the issue pops up.

nodes/k8s_kind/k8s_kind.go
```diff
@@ -0,0 +1,180 @@
...
+
+// Register registers the node in the global Node map.
+func Register(r *nodes.NodeRegistry) {
+	r.Register(kindnames, func() nodes.Node {
+		return new(k8s_kind)
+	}, nil)
+}
+
+type k8s_kind struct {
+	nodes.DefaultNode
```
aojea, line 34: you can add a field `labels map[string]string` to this object, and use it to store "containerlab" labels

aojea: hmm, I think I've read this wrong, scratch this comment

steiler: containerlab is a one-off tool. so there is no state kept between executions. and it just does what it is told ... e.g. deploy deploys the topology and quits ... exec the same... thats why we need to keep the state we need externally and we choose to do so in the labels of the nodes spawned by containerlab

aojea: Let me see if I understand k8s_kind represent a cluster, i.e. multiple nodes, no?

steiler: yes it does and that is a new thing. until k8s_kind there is a 1to1 mapping between configured node and resulting container

steiler: I have the idea to spawn pause containers as placeholders for external and k8s_kind containers, to attach the labels to and do the necessary stuff or maintain an index of deployed nodes per topology in a seperate file or so all things that would also work but require a quite some change in clab rather then allowing user-defined labels on the kind docker containers.

steiler:
@hellt 
We always need to have a k8s_kind node (for kind cluster spin-up) and all the resulting containers (master / worker) as ext-containers for wireing. This also results in duplication of containers in the output, because they are reported from the k8s_kind node and as the ext-container.
One option would be to perform some form of deduplication and throw the reported ext-containers away?!
Or maybe cleaner allow for node nesting in the topology definition. where a node can now have multiple sub-nodes that maybe default to ext-container somehow. (Not sure if this might also solve some scenarios where a node spins up seperate control-plane and dataplane containers)....

aojea:
... or maybe you reached the point of adding another abstraction layer and have the concept of cluster of nodes 🙃

steiler:
resolved merge conflicts

hellt:
Like a good red wine. Gets only better with time

hellt:
image

rebased on top of the current main

hellt:
thanks @steiler 
this took some time, but it is ready :D

I have only one issue I wanted to log after I merge this.
The issue is `clab ins -a` does not show any k8s-kind or ext-containers defined.
Could be tricky to solve because of labels that we don't control...

aojea:
👏

PR #1122, opened 2022-12-07
Fixing docker image pulling

steiler:
The way we are trying to make sure an image is present is via the 'reference' filter. This is a fuzzy filter that does not do exact match. See on cli:
```
$ docker images --filter=reference="alpine"
REPOSITORY   TAG       IMAGE ID       CREATED        SIZE
alpine       3         49176f190c7e   2 weeks ago    7.05MB
alpine       3.16      9c6f07244728   3 months ago   5.54MB
```
reference rearch for 'alpine' will in this case find alpine:3 and alpine:3.16. However using or pulling 'alpine' as an image will result in 'alpine:latest' which is not available. The deployment will not fail but stale, when the image is not availabe.

Introducing containsImageTagExactMatch(...) in this, to perform anadditional exact match on the returned tags.

### DIFF
--- a/runtime/docker/docker.go
+++ b/runtime/docker/docker.go
@@ -462,13 +462,16 @@ func (d *DockerRuntime) PullImageIfRequired(ctx context.Context, imageName strin
 		return err
 	}
 
+	canonicalImageName := utils.GetCanonicalImageName(imageName)
+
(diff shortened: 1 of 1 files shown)


steiler:
closing in favour of #1123

PR #1123, opened 2022-12-08
fix image pull routine

hellt:
@steiler check this, I think this method is cleaner, as it doesn't need to list all the images and do hairy tag matching.


steiler:
Yes, you're right, this mechanism is nicer I had not expected that you'd be able to query for the `docker.io/library/alpine` image and get the `alpine` image back.
But it seems to work, at least the CLI works that way ... so yes, its better!

PR #1124, opened 2022-12-09
Extend node interface2

steiler:
Moving LicenseCheck and vCPU&FreeMem into node
Plus fixing the Free / Available Memory check

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -6,15 +6,12 @@ package clab
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
-	"runtime"
 	"sort"
 	"strings"
 
-	"github.com/mackerelio/go-osstat/memory"
 	log "github.com/sirupsen/logrus"
 	"github.com/srl-labs/containerlab/labels"
 	"github.com/srl-labs/containerlab/nodes"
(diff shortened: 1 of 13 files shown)


hellt:
@steiler summing is not gonna work. Hosts can compress mem and oversubscribe the CPU. So I am good with having users to deal with that on their own.
Maybe some very coarse warnings can be given by summing the memory required by the total num of nodes and retracting some to cope with mem deduplication.

e.g. you deploy 10 SR Linux, this is 20GB mem requirement. I'd say if a host mem is < 20GB-30% we issue a warning

steiler:
Also, what I did here is, that you can define the behaviour per Node (Kind) if it should fail or just issue a warning ...
https://github.com/srl-labs/containerlab/pull/1124/files#diff-96dc4083e6471d63c23a814646fbc3e336fb1f8569391d7d7d42aae9eb9e4dfbR25-R28

Take a look and maybe adjust thsi case for SRL (which is the only one I implemented this behaviour for so far)
https://github.com/srl-labs/containerlab/pull/1124/files#diff-84e2ae483c1b7bb1fe7b4e080f06f3b468a9ff01968d90da1747b0ce2d8acc84R130-R133

steiler:
just rebased to actual main, so can be reviewed and merged

steiler:
about to be split into two PRs

PR #1125, opened 2022-12-10
Node scoped checks part1

hellt:
proposal for #1119

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -546,16 +546,16 @@ func (c *CLab) ListContainers(ctx context.Context, filter []*types.GenericFilter
 	return containers, nil
 }
 
-// ListContainersClabNodes lists all containers based on the nodes stored in the clab instance.
-func (c *CLab) ListContainersClabNodes(ctx context.Context) ([]types.GenericContainer, error) {
+// ListNodesContainers lists all containers based on the nodes stored in clab instance.
+func (c *CLab) ListNodesContainers(ctx context.Context) ([]types.GenericContainer, error) {
 	var containers []types.GenericContainer
 
 	for _, n := range c.Nodes {
-		ctrs, err := n.GetRuntimeInformation(ctx)
+		ct, err := n.GetContainer(ctx)
 		if err != nil {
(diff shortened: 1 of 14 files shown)


steiler:
@hellt I think my version of the node interface (specifically UpdateConfigWithRuntimeInfo(...)) was better extendable and had a clearer seperation on what is configuration comming from the user vs what is on the fly generated / runtime assigned data.

Further, later on the Kind inplementation will spawn multiple containers from a single node. whilst being able to return multiple GenericRuntimeInformations per node. So my suggestion is to stick with my implementation.

steiler:
https://github.com/srl-labs/containerlab/pull/1125/commits/7f0e1f2ef94eb91946d9744d13451f41588f4124#diff-831ecb8143ceda28e72434a64e3b7f88d15f9492071603608fd0180ae3105e67R558-R564
Here we should return an valid list, without entries (len(list) == 0 ) but then we can spare the extra check.

hellt:
My main concern was this function

https://github.com/srl-labs/containerlab/pull/1119/files#diff-d25cfdb9022b8a280a3cf87aa5a83926d393e4d5d28c550dd57cca87c9130359R111

It is named after something it doesn't do. You don't get runtime informations, you get containers and while doing that you update node configs with runtime informations for network settings and cID. 

This function was a starting point for a refactor. 

Additionally, there was multiple calls for this func which I don't think necessary, you want to update nodes with runtime info single time, once you started them and runtime info is available. 

I can rename GetContainer to GetContainers to support multi-container nodes, but the overwrite of config with runtime generated info was already there. 

I will create a branch out of your PR and base this PR on top of it so that file changes would be against the base branch it will be clear what I modified. 

We can have a call in a couple of hours to sync up on that one

steiler:
Yes you're right, the function does write stuf into the config. I don't like it at all... but I left that for the time being... we need to change that as well. I haven't had the time to do so by now... anyways I do think that it was at least from the post-deploy perspective the right way.
Thats why I also started not using cfg any more when populating e.g. hostsfile. All of this should go into the RuntimeInformation.
Yes, please ping me and we will have the discussion.

PR #1134, opened 2022-12-16
Refactor exec node interface part 1

hellt:
Hi @steiler 

I propose to make the following changes in the part1 of the review.

Here I rename the Node interface methods to `RunExecs` and `RunExec`

`RunExecs` take a list of string commands to work on, whereas `RunExec` take an `ExecCmd` (renamed).

I have changed the commands used for different kinds from `slice` to `string` type, as it is easier for humans to check what the command is about visually.

PS. I plan to do more renaming in the following PRs, but before I wanted to clear this major change with you first

### DIFF
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -16,13 +16,13 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-// templates to execute.
+// TemplateNames is templates to execute.
 var TemplateNames []string
 
-// path to additional templates.
+// TemplatePaths is path to additional templates.
 var TemplatePaths []string
 
-// debug count.
+// DebugCount is a debug verbosity counter.
 var DebugCount int
 
 type NodeConfig struct {
@@ -34,7 +34,7 @@ type NodeConfig struct {
 	Info []string
 }
 
-// Load templates from all paths for the specific role/kind.
+// LoadTemplates loads templates from all paths for the specific role/kind.
 func LoadTemplates(tmpl *template.Template, role string) error {
 	for _, p := range TemplatePaths {
 		fn := filepath.Join(p, fmt.Sprintf("*__%s.tmpl", role))
@@ -107,7 +107,7 @@ func RenderAll(allnodes map[string]*NodeConfig) error {
 	return nil
 }
 
-// Implement stringer for NodeConfig.
+// String implements stringer interface for NodeConfig.
 func (c *NodeConfig) String() string {
 	s := fmt.Sprintf("%s: %v", c.TargetNode.ShortName, c.Info)
 	return s
--- a/clab/config/transport/ssh.go
+++ b/clab/config/transport/ssh.go
@@ -21,7 +21,7 @@ type SSHSession struct {
 
 type SSHTransportOption func(*SSHTransport) error
 
-// The SSH reply, executed command and the prompt.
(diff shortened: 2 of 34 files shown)


steiler:
Beware, that this way you will specifically need to make sure, no one is providing execs to the config of a bridge or ova...
In my implementation these kinds would just overwrite these functions and no special handling would be needed... you are reintroducing specal code for these kinds of nodes by doing the following:

hellt:
@steiler I am not sure I understand your comment

Here is a lab I used to test the implementation

```yaml
name: linux

topology:
  nodes:
    l1:
      kind: linux
      image: alpine
      exec:
        - ash -c 'cat /etc/os-release | grep NAME' # correct cmd
    l2:
      kind: linux
      image: alpine
      exec:
        - 'echo "l2 node"' # correct cmd
    test-bridge: # should exist for this lab to run
      kind: bridge
      exec:
        - "ip link" # should be a noop, since execs have no meaning

  links:
    - endpoints: ["l1:eth1", "l2:eth1"]
```

The `test-bridge` node has `exec` config provided, but since this node has `RunExecs` func, it is being called and the warnings are emitted. Here is the output from this lab run. What am I missing here?

```log
❯ dclab dep -c -t private/exec.clab.yml
INFO[0000] Parsing & checking topology file: exec.clab.yml 
INFO[0000] Destroying lab: linux                        
INFO[0000] Removed container: clab-linux-l2             
INFO[0000] Removed container: clab-linux-l1             
INFO[0000] Removing containerlab host entries from /etc/hosts file 
INFO[0000] Removing /root/container-lab/clab-linux directory... 
INFO[0000] Creating lab directory: /root/container-lab/clab-linux 
INFO[0000] Creating container: "l1"                     
INFO[0000] Creating container: "l2"                     
INFO[0001] Creating virtual wire: l1:eth1 <--> l2:eth1  
INFO[0002] Adding containerlab host entries to /etc/hosts file 
WARN[0002] Exec operation is not implemented for kind "bridge" 
WARN[0002] Failed to exec commands for node "test-bridge" 
INFO[0002] Executed command 'echo l2 node' on l2. stdout:
l2 node 
INFO[0002] Executed command 'ash -c cat /etc/os-release | grep NAME' on l1. stdout:
NAME="Alpine Linux"
PRETTY_NAME="Alpine Linux v3.17" 
INFO[0002] 🎉 New containerlab version 0.33.0 is available! Release notes: https://containerlab.dev/rn/0.33/
+---+---------------+--------------+--------+-------+---------+----------------+----------------------+
| # |     Name      | Container ID | Image  | Kind  |  State  |  IPv4 Address  |     IPv6 Address     |
+---+---------------+--------------+--------+-------+---------+----------------+----------------------+
| 1 | clab-linux-l1 | 8c9d5d506006 | alpine | linux | running | 172.20.20.2/24 | 2001:172:20:20::2/64 |
| 2 | clab-linux-l2 | 53fa5b6e4a96 | alpine | linux | running | 172.20.20.3/24 | 2001:172:20:20::3/64 |
+---+---------------+--------------+--------+-------+---------+----------------+----------------------+
```

steiler:
Sorry, yes you are right! no RunExecConfig will not come with no RunExec... in that way we're back on track...
You're right!

PR #1139, opened 2022-12-19
move exec code to exec package

hellt:
In this review part I move the exec code to its own `exec` package


hellt:
@steiler any objection with moving exec functionality from types to its own package?

steiler:
Yes that's alright.
At some point we might even want to refactor the whole structure a little more and align more to https://github.com/golang-standards/project-layout
So most of the code (not limited to exec) I think should go into the root repo folder of `pkg`... maybe something into `internal`.

PR #1144, opened 2022-12-20
Implementing xrd kind

trustywolf:
This PR is created to include support of [Cisco IOS XRd](https://www.cisco.com/c/en/us/support/routers/ios-xrd/series.html): xrd-control-plane in containerlab

Here are some lab examples for testing(still work in progress):
- https://github.com/trustywolf/xrd-containerlab

Inspired by #1101 #1072 #1059

References:
- [xrd-tools](https://github.com/ios-xr/xrd-tools)
- [XRd tutorials Series](https://xrdocs.io/virtual-routing/tags/#xrd-tutorial-series)

### DIFF
--- a/nodes/xrd/xrd.cfg
+++ b/nodes/xrd/xrd.cfg
@@ -1,6 +1,3 @@
-!! Copyright 2022 Nokia
-!! Licensed under the BSD 3-Clause License.
-!! SPDX-License-Identifier: BSD-3-Clause
 !
 hostname {{ .ShortName }}
 !
--- a/nodes/xrd/xrd.cfg
+++ b/nodes/xrd/xrd.cfg
@@ -45,6 +45,6 @@ router static
 ssh server v2
 {{- if .Env.CLAB_MGMT_VRF }}
 ssh server vrf {{ .Env.CLAB_MGMT_VRF }}
-ssh server netconf vrf {{ .Env.CLAB_MGMT_VRF }}
 {{- end}}
+ssh server netconf {{ if .Env.CLAB_MGMT_VRF }} vrf {{ .Env.CLAB_MGMT_VRF }}{{end}}
 end
--- a/nodes/xrd/xrd.cfg
+++ b/nodes/xrd/xrd.cfg
@@ -8,7 +8,7 @@ username clab
 !
 grpc
 {{ if .Env.CLAB_MGMT_VRF }} vrf {{ .Env.CLAB_MGMT_VRF }}{{end}}
- port 57344
+ port 9339
  no-tls
  address-family dual
 !
--- a/nodes/xrd/xrd.cfg
+++ b/nodes/xrd/xrd.cfg
@@ -1,0 +1,53 @@
+!! Copyright 2022 Nokia
+!! Licensed under the BSD 3-Clause License.
+!! SPDX-License-Identifier: BSD-3-Clause
+!
+hostname {{ .ShortName }}
+!
+username clab
+ group root-lr
+ group cisco-support
+ secret clab@123
+!
+grpc
+{{ if .Env.CLAB_MGMT_VRF }} vrf {{ .Env.CLAB_MGMT_VRF }}{{end}}
+ port 57344
+ no-tls
+ address-family dual
+!
+{{- if .Env.CLAB_MGMT_VRF }}
+vrf {{ .Env.CLAB_MGMT_VRF }}
+ address-family ipv4 unicast
+ !
+ address-family ipv6 unicast
+ !
+{{- end}}
+!
+line default
+  transport input ssh
+!
+netconf-yang agent
+ ssh
+!
+interface MgmtEth0/RP0/CPU0/0
+{{ if .Env.CLAB_MGMT_VRF }} vrf {{ .Env.CLAB_MGMT_VRF }}{{end}}
+!
+router static
+{{ if .Env.CLAB_MGMT_VRF }} vrf {{ .Env.CLAB_MGMT_VRF }}{{end}}
+{{- if .MgmtIPv4Gateway }}
+ address-family ipv4 unicast
+  0.0.0.0/0 MgmtEth0/RP0/CPU0/0 {{ .MgmtIPv4Gateway }}
+  !
+{{- end}}
+{{- if .MgmtIPv6Gateway }}
+ address-family ipv6 unicast
+  ::/0 MgmtEth0/RP0/CPU0/0 {{ .MgmtIPv6Gateway }}
+  !
+{{- end}}
+!
+ssh server v2
+{{- if .Env.CLAB_MGMT_VRF }}
+ssh server vrf {{ .Env.CLAB_MGMT_VRF }}
+ssh server netconf vrf {{ .Env.CLAB_MGMT_VRF }}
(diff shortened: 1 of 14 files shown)


hellt:
Magnificent! Thanks @trustywolf 
I'll see where can I get an image to test it out.

hellt:
@trustywolf I have refactored the env var population code to only create the links that were defined in a topology. I am not sure if this will block users from live-attaching interfaces to xrd, maybe this is not supported at all.

I wonder, do you know what needs to be added to make `vrouter` variant to work? Currently it complaints about the interface definition type:

```
❯ docker logs -f clab-srlxrd01-xrd
[WARNING] XRd is expected to require 5Gi of RAM, only 3.9Gi available
ERROR: Unknown XR_INTERFACES interface type(s):
 * linux:eth1,xr_name=Gi0/0/0/0
[ERROR  ] Platform init failed
[ERROR  ] XRd hit a critical error during initialization and has aborted launch
```

trustywolf:
@hellt thank you for your review and comments :)

XRd vRouter has a data plane(based on DPDK), so host requirement(such as CPU extensions & HugePages settings) and the format of `XR_INTERFACES` is quite different from `control-plane` variant.

You can find an `XR_INTERFACES` example of `vrouter` variant at the end of [this article](https://xrdocs.io/virtual-routing/tutorials/2022-08-23-xrd-with-docker-control-plane-and-vrouter/).

There is a tool made by Cisco guys to determine the suitability of the host machine for `control-plane` and `vrouter` variant, you can find it in the [host-check](https://github.com/ios-xr/xrd-tools/blob/main/scripts/host-check) script from [ios-xr/xrd-tools](https://github.com/ios-xr/xrd-tools) repo. Please refer to [this article](https://xrdocs.io/virtual-routing/tutorials/2022-08-22-setting-up-host-environment-to-run-xrd/), which shows the usage of it.

You can also find all XRd's environment variables from the  [launch-xrd](https://github.com/ios-xr/xrd-tools/blob/main/scripts/launch-xrd) script. The `vrouter` variant have some more environment variables than control-plane. [This article](https://xrdocs.io/virtual-routing/tutorials/2022-08-25-user-interface-and-knobs-for-xrd/) describes the meaning of these environment variables.

Maybe we need to implement `type`(such as `cp` and `vr`) like `srl` kind to support the `vrouter` variant?

hellt:
Yes, using type was exactly my line of thought to get support for vrouter and cp variants. Thanks for the links too the relevant articles

nodes/xrd/xrd.go
```diff
@@ -0,0 +1,139 @@
...
+	return err
+}
+
+// genInterfacesEnv calculates how many interfaces were defined for xrd node
+// and populates the content of a required env var.
+func (n *xrd) genInterfacesEnv() {
+	// xrd (at least control-plane variant) needs XR_INTERFACE ENV var to be populated for all active interface
+	// here we take the number of links users set in the topology to get the right # of links
+	var interfaceEnvVar string
+
```
steiler, line 131: Should this not maybe reference the names from the endpoint config, rather then the index var I?!
This will lead to confusion, not?!

hellt: what do you mean, @steiler?
the `i` is the index of the endpoints slice element

hellt:
@trustywolf looks like vrouter can only work with PCI devices and link them to the interfaces of a vrouter, which is a nice feature, but not a priority for containerab, where virtual topologies are wired using the veth pairs.
So I will leave this feature out, maybe some one will add it later.

I will do the docs part now for this PR as it is now and merge it.
Thanks again for your contribution 🦾

trustywolf:
@hellt I feel the same way, at this time `control-plane` variant is enough for me to test IOS XR's control plane functions(may be the same as `crpd` & `ceos`).

nodes/xrd/xrd.cfg
```diff
@@ -0,0 +1,50 @@
...
+hostname {{ .ShortName }}
+!
+username clab
+ group root-lr
+ group cisco-support
+ secret clab@123
+!
+grpc
+{{ if .Env.CLAB_MGMT_VRF }} vrf {{ .Env.CLAB_MGMT_VRF }}{{end}}
+ port 9339
```
trustywolf, line 11: @hellt The port number range is 57344-57999.
Ref: https://github.com/CiscoDevNet/grpc-getting-started/blob/master/README.md#grpc-server

hellt: this must be an outdated info. IANA assigned port for gnmi/gnoi is 9339 and since it works with xrd I am gonna use it to prevent proliferation of custom port ranges

hellt:
@trustywolf do you know where xrd keeps its config file (not the first-boot one)? For example, when a user commits changes done via SSH, where to grab the file that contains the new config file?

nodes/xrd/xrd.go
steiler: interfaceEnvVar += fmt.Sprintf("linux:eth%d,xr_name=Gi0/0/0/%d;", i+1, i)
Yes, and then the index is used but not the data behind the index...
Or am I missing something here myself?!

hellt: @steiler I think you refer to the usage of `linux:eth%d` whereas it could've been the entire endpoint name listed here instead of an index?
Yes, I think I can do that, to make sure that users might say

```
["xrd:eth5", "srl:e1-1"]
```

and this still would work

hellt: @steiler like this [ecdf202](https://github.com/srl-labs/containerlab/pull/1144/commits/ecdf2021ef2f7b5dda32bbfc48ad25c9fb1d6022)

trustywolf:
@hellt from [xrd-tools](https://github.com/ios-xr/xrd-tools) i can only find the way to persist data is to bind /xr-storage folder. After xrd booted, I can't find any text files which include running config, maybe db file under /xr-storage/config/config/lr/running/commitdb/ ?

nodes/xrd/xrd.go
steiler: Well yes, in parts...
Would you not also want to allow interface numbers on the xrd side that is not consecutive...
I'm thinking someone who tries to mimic the prod network... Has connections on gi0/0/0/24 and does not want to change interface names in the config ... So can't we come up wit a schema that also allows non consecutive xrd interface numbers?
Wouldn't that make sense?!
We also do so in the srl case.

hellt: @steiler makes sense. Let me try that

hellt: @steiler check [e262674](https://github.com/srl-labs/containerlab/pull/1144/commits/e2626748ce30bb35f6b476192fa22f161d956fc6). Sounds about right?

steiler: Yes, I'd say that's in line with my thinking.
Thanks!

hellt: @steiler nice, thanks. I completely missed on this opportunity

hellt:
missing doc changed for interface naming added in #1146

trustywolf:
doc about how to drop directly into XR CLI added in #1147

PR #1147, opened 2022-12-21
xrd kind - docs improvement

trustywolf:
- Add docs about how to drop directly into XR CLI using `docker exec` command


hellt:
Thanks @trustywolf ! Good catch

Issue #1152, opened 2022-12-25
Index out of range error in v0.34.0 when using ext-container


oshothebig:
Index out of range error happens in v0.34.0 when using ext-container. The error doesn't happen in v0.33.0.

The following is a topology file that can cause the error.

```yaml
name: extcont

topology:
  nodes:
    external1:
      kind: ext-container
    external2:
      kind: ext-container
  links:
  - endpoints: ["external1:eth1", "external2:eth1"]
```

Create `external1` and `external2` containers and run `sudo clab deploy`, then we'll see the issue when using v0.34.0.

```
$ docker run -d --name external1 -t ghcr.io/nicolaka/netshoot
8ab771a11317977a78f1e2842a566e0a9eb22f488bb5e49028da6463a7477f60
$ docker run -d --name external2 -t ghcr.io/nicolaka/netshoot
ca35e14b18d72852ff0b334a7ea88745bdb4f078a4b8f8198554a076903ab35a
$ sudo clab deploy
INFO[0000] Containerlab v0.34.0 started
INFO[0000] Parsing & checking topology file: extcont.clab.yaml
INFO[0000] Creating lab directory: /home/ubuntu/extcont/clab-extcont
INFO[0001] Creating virtual wire: external1:eth1 <--> external2:eth1
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/srl-labs/containerlab/nodes.(*DefaultNode).UpdateConfigWithRuntimeInfo(0x4000797a00, {0x2cdbe30?, 0x4000797a80?})
	github.com/srl-labs/containerlab/nodes/default_node.go:144 +0x190
github.com/srl-labs/containerlab/cmd.deployFn(0x4316dc0?, {0x26cb2ef?, 0x0?, 0x0?})
	github.com/srl-labs/containerlab/cmd/deploy.go:206 +0xc40
github.com/spf13/cobra.(*Command).execute(0x4316dc0, {0x4558a88, 0x0, 0x0})
	github.com/spf13/cobra@v1.6.1/command.go:916 +0x60c
github.com/spf13/cobra.(*Command).ExecuteC(0x4315100)
	github.com/spf13/cobra@v1.6.1/command.go:1044 +0x368
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.6.1/command.go:968
github.com/srl-labs/containerlab/cmd.Execute()
[...]

PR #1154, opened 2022-12-26
fixing #1152

steiler:
fixing #1152
GetContainers() in ext-container with query for shortname

### DIFF
--- a/labels/labels.go
+++ b/labels/labels.go
@@ -1,0 +1,13 @@
+package labels
+
+const (
+	// label names.
+	ContainerlabLabel = "containerlab"
+	NodeNameLabel     = "clab-node-name"
+	NodeKindLabel     = "clab-node-kind"
+	NodeTypeLabel     = "clab-node-type"
+	NodeGroupLabel    = "clab-node-group"
+	NodeLabDirLabel   = "clab-node-lab-dir"
+	TopoFileLabel     = "clab-topo-file"
+	NodeMgmtNetBr     = "clab-mgmt-net-bridge"
+)
--- a/labels/labels.go
+++ b/labels/labels.go
@@ -1,13 +1,13 @@
 package labels
 
 const (
-	// label names.
-	ContainerlabLabel = "containerlab"
-	NodeNameLabel     = "clab-node-name"
-	NodeKindLabel     = "clab-node-kind"
-	NodeTypeLabel     = "clab-node-type"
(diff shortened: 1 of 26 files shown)


hellt:
Hi @steiler 
Can you address the deep source comments?

steiler:
To ba able to reuse the container label keys I've moved them now to their own package, otherwise I ended up in an import cycle.
Further I've introduced the `GetRuntimeContainerName()` as part of the `NodeOverwrites` interface.
We had copy-pasted the `GetContainers()` function on e.g. the ext-container whilst just changing the filter name from Cfg.LongName to Cfg.ShortName.
Now Cfg.LongName is the refault response to `GetRuntimeContainerName()` implemented in DefaultNode. ext-container overwrites this returning ShortName. All functions that need the name to query the Runtime are chnaged to use `GetRuntimeContainerName()`

labels/labels.go
```diff
@@ -0,0 +1,13 @@
...
+
+const (
+	// label names.
+	ContainerlabLabel = "containerlab"
+	NodeNameLabel     = "clab-node-name"
+	NodeKindLabel     = "clab-node-kind"
+	NodeTypeLabel     = "clab-node-type"
+	NodeGroupLabel    = "clab-node-group"
+	NodeLabDirLabel   = "clab-node-lab-dir"
+	TopoFileLabel     = "clab-topo-file"
```
oshothebig, line 11: It might be better to remove `Label` suffix from these constants. They look somewhat redundant as the package name is `labels` and they are not consistent with `NodeMgmtNetBr` that doesn't have `Label` suffix.

hellt:
@steiler we should also add a test with ext containers as part of the smoke test suite

labels/labels.go
hellt: I agree, did that in [859f22b](https://github.com/srl-labs/containerlab/pull/1154/commits/859f22be0f93696ff3542dd7b8f2e2135b7934e8)

hellt:
thanks @steiler!
One thing I noticed is that the exec command doesn't honor the filter label

image

Has this been addressed in other PRs?

steiler:
You're right, but the issue was introduced already in v0.34.0 or even v0.33.0.
We're somehow iterating over all the nodes and executing the commands on all of them ... not honoring any label what so ever.
Since this is a new discovery, there is no fix for it so far.
I'll go and fix that in a seperate PR since thats not really related to the initial issue.

PR #1156, opened 2022-12-27
Introducing NodeRegistry

steiler:
The Implementation, using the once construct is ugly and not quite good practice. Also the global var we used as a node registry is non-optimal.
Here is a different approach introducing a NodeRegistry. A seperate object that keeps track of the registered Node types.

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -15,7 +15,6 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/srl-labs/containerlab/nodes"
-	allNodes "github.com/srl-labs/containerlab/nodes/all"
 	"github.com/srl-labs/containerlab/runtime"
 	_ "github.com/srl-labs/containerlab/runtime/all"
 	"github.com/srl-labs/containerlab/runtime/docker"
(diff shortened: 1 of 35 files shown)


hellt:
Hi @steiler 
I have removed what I consider to be unnecessary interfaces. What was the point of adding them?
The rest looks good to me. If you are ok with removed interfaces and me moving the registry to the Clab struct I am good with merging this

steiler:
I like to use interfaces in a lot more places that we use them right now.
interfaces is not just a vehicle to implement different production level behaviours but also helps a lot with testing.
You can generate mocks only from interfaces afaikt and use these to ease testing.
Sure this was not in use just now... but in my view a good practice.

hellt:
I think creating the interface is only viable when you have multiple types conforming to it. Like different nodes and the corresponding node interface. 
An interface that is implemented by a single type doesn't bring any benefit imho (and complicates code by introducing another level of indirection). You still can create a struct  used for testing manually and perform the tests over it -- especially in case of simple structs like registry and credentials

steiler:
It would be used to mock the registry itself not for testing the registry.

PR #1157, opened 2022-12-27
fix-exec-format-and-add-multipe-cmd

steiler:
The format switch for exec was stored in a different variable then was set via cobra...
Also this PR introduces the capability of providing multiple `--cmd "<CONTAINER-COMMAND>"` options to the exec call.


hellt:
lgtm, thanks

Issue #1158, opened 2022-12-28
json output


steiler:
I think we have had implemented json output in a way, that when executing commands in the containers, that produce json output themselves, this was a seemless json struct.
Right now we do produce the stdout field with an escaped json string.
But we should come back with the seemless integration.

Actual:
```
steiler@vbox:~/projects/containerlab$ sudo ./containerlab -f json -t lab-examples/linux/linux.clab.yaml exec --cmd "ip -j a show dev lo" | jq
INFO[0000] Parsing & checking topology file: linux.clab.yaml 
{
  "R1": [
    {
      "cmd": [
        "ip",
        "-j",
        "a",
        "show",
        "dev",
        "lo"
      ],
      "return-code": 0,
      "stdout": "[{\"ifindex\":1,\"ifname\":\"lo\",\"flags\":[\"LOOPBACK\",\"UP\",\"LOWER_UP\"],\"mtu\":65536,\"qdisc\":\"noqueue\",\"operstate\":\"UNKNOWN\",\"group\":\"default\",\"txqlen\":1000,\"link_type\":\"loopback\",\"address\":\"00:00:00:00:00:00\",\"broadcast\":\"00:00:00:00:00:00\",\"addr_info\":[{\"family\":\"inet\",\"local\":\"127.0.0.1\",\"prefixlen\":8,\"scope\":\"host\",\"label\":\"lo\",\"valid_life_time\":4294967295,\"preferred_life_time\":4294967295},{\"family\":\"inet6\",\"local\":\"::1\",\"prefixlen\":128,\"scope\":\"host\",\"valid_life_time\":4294967295,\"preferred_life_time\":4294967295}]}]\n",
      "stderr": ""
    }
  ],
  "R2": [
    {
      "cmd": [
        "ip",
        "-j",
        "a",
        "show",
        "dev",
        "lo"
      ],
      "return-code": 0,
[...]

PR #1160, opened 2022-12-28
jsonOutput embedding

steiler:
This PR contains changes that allow for the exec output to be valid json, embedding the result of in container executions in the json output structure.
See #1158

### DIFF
--- a/clab/exec/exec.go
+++ b/clab/exec/exec.go
@@ -11,23 +11,43 @@ import (
 	"github.com/google/shlex"
 )
 
+type ExecFormat string
+
 const (
-	ExecFormatJSON  string = "json"
-	ExecFormatPlain string = "plain"
+	ExecFormatJSON  ExecFormat = "json"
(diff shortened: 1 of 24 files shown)


steiler:
Thought about different approaches of implementing this. The issue is, that we use the ResultCollection containing the ExecResults and we call for a json.marshal on the ResultCollection.
This approach introduces an additional parameter to the nodes exec function, which is a function pointer to a function that returns a more abstract interface for ExecResult -> ResultHolderSetter. This is then also implemented by the ExecResultJson.
The exec function of node then takes a pointer to the constructor for these ExecResult(Json) structs.
Such that the correct ExecResult Struct is taken in accordance to the requested output format.

steiler:
@hellt in my view an open discussion point is how we want / can handle non json output from the command when json output is requested. Right now we fail with an error, because the unquoted string is not marshallable. but I guess we want this to simply result in a string in the json output. Or are you of different opinion?

hellt:
done in #1188

PR #1162, opened 2022-12-30
Reduce duplicates in Makefile

oshothebig:
This PR reduces duplicates in Makefile by introducing variables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,17 +5,22 @@ MKDOCS_VER = 8.3.9
 # make sure to also change the mkdocs version in actions' cicd.yml and force-build.yml files
 MKDOCS_INS_VER = 8.5.11-insiders-4.27.0-hellt
 
+DATE := $(shell date)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -5,17 +5,25 @@ MKDOCS_VER = 8.3.9
 # make sure to also change the mkdocs version in actions' cicd.yml and force-build.yml files
 MKDOCS_INS_VER = 8.5.11-insiders-4.27.0-hellt
 
+DATE := $(shell date)
+COMMIT_HASH := $(shell git rev-parse --short HEAD)
+
+LDFLAGS := -s -w
+LDFLAGS += -X 'github.com/srl-labs/containerlab/cmd.version=0.0.0'
```
hellt, line 12: Hi @oshothebig 
Why don't we declare this variable once with all the flags in it?

oshothebig: I'll modify to put all the flags into a single variable.

oshothebig: The variable declaration is revised.

PR #1163, opened 2022-12-30
Correct inconsistencies between PHONY declarations and the actual targets

oshothebig:
`serve` and `ttl-push` are declared as `.PHONY`, but they don't seem to be consistent with targets in Makefile. I assume `site` and `oci-push` are what we want to use though I'm not so sure.


hellt:
Thanks @oshothebig 
Looks good

PR #1167, opened 2023-01-01
Remove unnecessary entrypoint option

oshothebig:
`squidfunk/mkdocs-material` container image specifies `mkdocs` as the entrypoint and `--entrypoint mkdocs` is redundant.


hellt:
Thanks @oshothebig 
Looks good

PR #1168, opened 2023-01-05
fixing the exec function

steiler:
- moved GenericContainer into runtime package (avoid import cycle)
- implement exec on GenericContainer
- perform exec on GenericContainer retrieved via listcontainers with filter

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/srl-labs/containerlab/clab"
 	"github.com/srl-labs/containerlab/clab/exec"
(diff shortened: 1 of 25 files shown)


steiler:
@hellt As discussed, ext-containers do not work this way, so we need to fix them. Maybe disable the ext-container test for now

cmd/exec.go
```diff
@@ -56,25 +56,44 @@ var execCmd = &cobra.Command{
...
 				execResult, err := node.RunExec(ctx, execCmd)
 				if err != nil {
 					// skip nodes that do not support exec
 					if err == exec.ErrRunExecNotSupported {
 						continue
 					}
-					return err
 				}
-				resultCollection.Add(node.Config().ShortName, execResult)
+				resultCollection.Add(node.Cfg.ShortName, execResult)
```
hellt, line 96: Hi @steiler 
I propose we put a requirement for users to always specify a topology file as an argument for `clab exec` command. That way we are able to init CLab structure and init all the node.

With that we are able to add labels to nodes which are not orchestrated by clab (i.e. ext-container nodes).

To make filters to work, we have to implement `c.ListNode(ctx, filters)` function that should return a slice of clab Nodes interfaces which satisfy to the provided filter.

Then we can leverage nodes RunExec function without the need to have execs implemented for generic containers, as we will work only with nodes

steiler:
see #1186

PR #1172, opened 2023-01-08
fix ovs bridge check

hellt:
during the refactoring we started to use linux bridge check as part of the `CheckDeploymentConditions`. This function doesn't work for ovs bridges.

In this PR the check is changed to list ovs bridges specifically.

The PR also fixes the exec bug for nodes which do not support it:

```
INFO[0001] Adding containerlab host entries to /etc/hosts file 
WARN[0001] Exec operation is not implemented for kind "ovs-bridge" 
WARN[0001] Failed to exec commands for node "myovs" 
```

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -257,6 +257,11 @@ func deployFn(_ *cobra.Command, _ []string) error {
 	// execute commands specified for nodes with `exec` node parameter
 	execCollection := exec.NewExecCollection()
 	for _, n := range c.Nodes {
+		// do not try to execute commands if the list is empty
+		if len(n.Config().Exec) == 0 {
+			continue
+		}
+
 		execResult, err := n.RunExecs(ctx, n.Config().Exec)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


hellt:
@steiler I think [a6c2615](https://github.com/srl-labs/containerlab/pull/1172/commits/a6c26154129fe5fb2c236b33230a9ad237f38910) fixes the issue of seeing warn logs for nodes which shouldn't support the exec.
We just exit early if nodes have no exec defined

steiler:
Proposal:
we only expose exec functions for either single string or single *exec.ExecCmd.... then the iteration would cut the issue short because len is 0. And multiple command have to be looped through by the caller?!

hellt:
@steiler so basically scrap RunExecs and only do RunExec node function?

steiler:
yes. I'm thinking why should a node be concered with generating ExecCmds from string ... thats not what it is there for. let someone else do it. device and conquer

PR #1188, opened 2023-01-13
Seamless json output in the exec output

hellt:
This PR adds `json` field to the ExecResult to make sure we have both `stdout` which is a string repr of the output and `json` field is populated when stdout contains a valid json

substitutes #1160


clab/exec/exec.go
```diff
@@ -51,23 +51,13 @@ func NewExecCmdFromSlice(cmd []string) *ExecCmd {
...
 // ExecResult represents a result of a command execution.
 type ExecResult struct {
-	Cmd        []string `json:"cmd"`
-	ReturnCode int      `json:"return-code"`
-	Stdout     string   `json:"stdout"`
-	Stderr     string   `json:"stderr"`
+	Cmd        []string        `json:"cmd"`
+	ReturnCode int             `json:"return-code"`
+	Stdout     string          `json:"stdout"`
+	Json       json.RawMessage `json:"json"`
```
hellt, line 59: @steiler I haven't been able to find a way to keep just `stdout` field playing nicely both with underlying valid json and just strings simultaneously.

Maybe some empty interface type for stdout is needed with dynamic type assertions, but I failed to come up with a solution to that.

Therefore, two distinct fields:

- stdout - for string based outputs
- json - for valid json outputs

clab/exec/exec.go
```diff
@@ -185,10 +174,21 @@ func (ec *ExecCollection) Dump(format string) (string, error) {
...
 	case ExecFormatJSON:
+		// when json format is requested, we check if stdout is a valid json
+		// to nicely display native json output that exec might have produced with `json` tag
+		for _, results := range ec.execEntries {
+			for _, r := range results {
+				if json.Valid([]byte(r.Stdout)) {
+					r.Json = json.RawMessage([]byte(r.Stdout))
+				}
+			}
+		}
```
hellt, line 185: the idea here is that we only populate `execResult.json` when we ask for json formatted output. 
In that case we check if the output has a valid json and populate `r.Json`

hellt (commented):
Hi @steiler 
What do you think about this?

clab/exec/exec.go
hellt: discard, I think I found a way

clab/exec/exec.go
hellt: discard, I have another approach

clab/exec/exec.go
```diff
@@ -51,22 +51,27 @@ func NewExecCmdFromSlice(cmd []string) *ExecCmd {
...
 }
 
 // ExecResult represents a result of a command execution.
 type ExecResult struct {
 	Cmd        []string `json:"cmd"`
 	ReturnCode int      `json:"return-code"`
-	Stdout     string   `json:"stdout"`
+	Stdout     Stdout   `json:"stdout"`
 	Stderr     string   `json:"stderr"`
 }
```
hellt, line 76: @steiler the idea here is to use an alias type and implement a custom marshaller for it.
if the stdout contains a valid json the marshaller just return those bytes, as they are already in shape

else, it uses the stdlib json.Marshal over the string type

PR #1205, opened 2023-01-18
enable clab config to use node registry for credentials

hellt:


steiler:
how about having a cred struct that makes username and password explicit fields rather then using a string slice?
if the string slice is required, then lets add a function to the struct that returns a string slice.
Thats would be a clearer, more obvious approach!?

steiler:
should it not be a registry.GetCredentials(KIND).GetSlice() rather then a function on the registry that creates and returns the slice?!
or even registry.GetKind(KIND).GetCredentials().GetSlice()

hellt:
@steiler ideally, credentials should belong to the actual node to allow per node credentials. The default credentials should be set on the kind level and act as a fallback default. 

I agree that creating a credentials struct that is common type makes sense

steiler:
oh yes haven't thought of node specific credentials. That makes sense.
Right now we have nodes that register with nil as *credentials which is not taken care of by this PR so far.
Shall we rather init with a credentials that would then yield ["",""] or what do you think is the right approach?

hellt:
@steiler I had to quickly fix this without doing much of a redesign, but it is now more aligned to what you said in your 1st comment and indeed handles nils in a way that empty credentials are returned.

This is not optimal or even thought off, it just quickly fixes the broken `clab config` functionality which desperately needs a redesign. For now this seems to patch things though.

PR #1206, opened 2023-01-19
Update get.sh

mabra94:
Allowing containerlab downgrades using this script for package installation on CentOS (i.e. rpm based packages). Was only tested on CentOS.

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -34,7 +34,7 @@ detectOS() {
     # Minimalist GNU for Windows
     mingw*) OS='windows' ;;
     esac
-    
+
     if [ -f /etc/os-release ]; then
         OS_ID="$(. /etc/os-release && echo "$ID")"
     fi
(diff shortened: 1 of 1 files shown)


get.sh
```diff
@@ -218,7 +234,11 @@ setPkgInstaller() {
         if [[ -n "$VARIANT_ID" && $VARIANT_ID == "coreos" ]]; then
             PKG_INSTALLER="rpm-ostree install --uninstall=containerlab --idempotent"
         else
-            PKG_INSTALLER="rpm -U"
+          if [ "$(printf '%s\n' "$TAG_WO_VER" "v$version" | sort -V | head -n1)" = "$TAG_WO_VER" ]; then
+              PKG_INSTALLER="rpm -U --oldpackage"
+          else
+              PKG_INSTALLER="rpm -U"
+          fi
```
hellt, line 241: I think that maybe it is acceptable to have `rpm -U --oldpackage` for both upgrade and downgrade procedures? I think you can install a newer package with `--oldpackage` arg present as well.

This way you won't need a check for a version here. But needs to be tested on an rpm-based system

mabra94: Indeed. Just tested it and it works. 
So I agree, this check here you referenced is not needed. I like the version check in "checkInstalledVersion()" though as it improves the echo statements and tells you whether you're upgrading or downgrading. Just cosmetics though.

Issue #845, opened 2022-04-06
clab/config.go remove dependency on Kind Constants


steiler:
In clab/config.go we rely on the Kind consts for bridge ovs, host and srl type.
These should be changed. The functionality should probably be added to the node implementations itself (when it comes to the srl related stuff) or should be indicated via a boolean flag on the Kind, like an "isBridge" flag  for the bridge or ovs-bridge type of kind.
which is then checked in this case:
https://github.com/srl-labs/containerlab/blob/main/clab/config.go#L411-L421

steiler:
in #828 I would rather then from a central spot stating requirements of certain kinds introduce a `MinCPURequired` flag with anything above 0 being a valid requirement. And then loop through all the kinds of that certain topology and check perform the check... thats the better and more generic approach

hellt:
@steiler after the recent PRs do we still have this dependency?

steiler:
Yes, that is resolved.
Just realized that there is some stuff remaining, that can be removed.
see #1227

PR #1207, opened 2023-01-19
introduce kind properties in noderegistry

steiler:


steiler:
Maybe we also convert the `NonDefaultRuntimes` into a KindProperty!?
https://github.com/srl-labs/containerlab/blob/main/nodes/cvx/cvx.go#L31
Then we would not have these maps dangling there in global var space etc.

@hellt opinion?

types/kind_properties.go
```diff
@@ -0,0 +1,13 @@
...
+	// IsRootNamespaceBased flags if the Kind is run in the Hosts NetworkNamespace
+	IsRootNamespaceBased bool
+}
+
+// NewKindProperties constructor for KindProperties
+func NewKindProperties() *KindProperties {
+	return &KindProperties{
+		IsRootNamespaceBased: false,
+	}
+}
```
hellt, line 13: @steiler I suggest we use `types.HostRequirements` for that, instead of creating KindProperties.
I wanted to have a flexibility where nodes conain information about their particular properties and requirements.

i.e. a node of a certain kind can be both rootns bound and not, which will be driven by the node constructor.
Same applies for non default runtimes, as runtime belongs to a node, and, say, `cvx` node can be started both with docker and ignite runtimes.

So I propose we move `IsRootNSBased` field to the `types.HostRequirements`

hellt (commented):
@steiler I would like to propose an alternative

types/kind_properties.go
steiler: _"I wanted to have a flexibility where nodes conain information about their particular properties and requirements."_
That is what we have here...
`HostRequirements` in my view, is the definition of requirements that a certain kind needs the host to meet, such that it can run on that host.
Whilst KindProperties define properties of a specific node type. That is a semantical differentiation I do not want to mix.

Here you see the main purpose of this change. I'm since quite a long time (#845) pushing on getting rid of Kind specific names or constants in the clab code, but rather abstract things in a way, that any node can flag that it has a certain behaviour, or requres such.

Hence this whole PR is about getting rid of the explicit node references in this line.

https://github.com/srl-labs/containerlab/pull/1207/files#diff-24901324638614d6ed95797828a77247d9dad583c83606ec43cce4f7ed2e8269L504

As you can see here any Node is able to express this property by initialising the KindProperties when registering the kind in the registry.
https://github.com/srl-labs/containerlab/pull/1207/files#diff-0ccbceb781798ef8f90f33c1d85c93e16a66e9c1969c9e5913c9035468cb6700L30-L34

If no special Properties need to be set (sane defaults apply) a nil pointer can be provided for the KindProperties.
https://github.com/srl-labs/containerlab/pull/1207/files#diff-8c1f8ab7e152c42d9b6becc8577e4ca9422ad543bc37dd85517a79f3d63b58e7L53-R55

Regarding "... a node of a certain kind can be both rootns bound and not." that is, in the course of the actual change news to me, since the implementation in main looks like this today, which is only simply checking kind types.

hellt: Yes, I understand the initial reasoning to get rid of kind constants, but I do not feel it is necessary to expand kind registry with kind properties to achieve that.

I believe that we should instead have things like "root ns property" defined on a node level and not inflating the kind registry with that information. The reason being, to allow nodes of the same kind to behave differently and be able to override defaults. In other words, I think it is best if the NodeConfig struct contains everything there is about the node, without being need to query kind registry for any properties. 

If not being part of the HostRequirements struct, it can be a field of the NodeConfig, or part of a new struct Properties, or a metadata attached to the node.

steiler:
closing this PR in favour of #1225

PR #1208, opened 2023-01-19
implement VerifyLicenseFiles()

steiler:


steiler:
@hellt, in my opinion the change you did to the function name limits the use of the function.
you are right that it at the moment only checks the existence of a file... but you could also check for the validity of the license and so on... which the old name allowed for ... now the usage is limited via the name...

hellt:
@steiler yes, I did it on purpose. Validity of licenses is in 99% cases can't be checked off- box. As well as license file name almost never matters as we would rename it the node code when we copy it over to the node FS

PR #1209, opened 2023-01-19
implements node based resource requirements

steiler:


nodes/srl/srl.go
```diff
@@ -124,11 +124,12 @@ func (s *srl) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
 	s.DefaultNode = *nodes.NewDefaultNode(s)
 	// set virtualization requirement
 	s.HostRequirements.SSSE3 = true
+	s.HostRequirements.MinVCPU = 2
+	s.HostRequirements.MinVCPUFailAction = types.FailBehaviourLog
+	s.HostRequirements.MinFreeMemoryGb = 2
+	s.HostRequirements.MinFreeMemoryGbFailAction = types.FailBehaviourLog
```
hellt, line 130: @steiler I think in srl case we should error and stop execution if vCPU < 2
we should keep logging action for memory, as a single node might be able to start with 2G

types/host_requirements.go
```diff
@@ -0,0 +1,89 @@
...
+func (h *HostRequirements) verifyMinFreeMemory() (bool, uint64) {
+	// if the MinFreeMemory amount is 0, there is no requirement defined, so result is true
+	// if != 0 then amount of Free Memory must be greater-equal the requirement
+	freeMemG := virt.GetSysMemory(virt.MemoryTypeAvailable) / 1024 / 1024 / 1024
+
+	boolResult := h.MinFreeMemoryGb == 0 || h.MinFreeMemoryGb != 0 && uint64(h.MinFreeMemoryGb) <= freeMemG
+	return boolResult, freeMemG
+}
+
+// verifyMinVCpu verify that the amount of re
```
hellt, line 83: missing comment

types/host_requirements.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// verifyMinFreeMemory verify that the amount of free memory with the requirement
+// it returns a bool indicating if the requirement is satisfied and the amount of free memory in GB
+func (h *HostRequirements) verifyMinFreeMemory() (bool, uint64) {
+	// if the MinFreeMemory amount is 0, there is no requirement defined, so result is true
+	// if != 0 then amount of Free Memory must be greater-equal the requirement
+	freeMemG := virt.GetSysMemory(virt.MemoryTypeAvailable) / 1024 / 1024 / 1024
+
+	boolResult := h.MinFreeMemoryGb == 0 || h.MinFreeMemoryGb != 0 && uint64(h.MinFreeMemoryGb) <= freeMemG
```
hellt, line 79: @steiler this looks hard on my eyes
what if you do an early return for a case then mem constraint is not defined
and then do a simple check:

```go
if h.MinFreeMemoryGb == 0 {
  return true, 0)

  freeMemG := virt.GetSysMemory(virt.MemoryTypeAvailable) / 1024 / 1024 / 1024
  valid := uint64(h.MinFreeMemoryGb) <= freeMemG
  return valid, freeMemG
```

same with minvcpu check

types/host_requirements.go
```diff
@@ -0,0 +1,89 @@
...
+	if h.VirtRequired && !virt.VerifyVirtSupport() {
+		return fmt.Errorf("the CPU virtualization support is required, but not available")
+	}
+	// check SSSE3 support
+	if h.SSSE3 && !virt.VerifySSSE3Support() {
+		return fmt.Errorf("the SSSE3 CPU feature is required, but not available")
+	}
+	// check minimum vCPUs
+	if valid, num := h.verifyMinVCpu(); !valid {
+		message := fmt.Sprintf("the defined minimum vCPU amount based on the nodes in your topology is %d whilst only %d vCPUs are available", h.MinVCPU, num)
```
hellt, line 46: I had a thought before, that the verify function should have access to the Node struct, so that we can say which node/kind puts this requirement.
But so far we can't do that, when the method belongs to the HostRequirements

```suggestion
		message := fmt.Sprintf("the topology requires minimum %d vCPUs, but the host has %d vCPUs", h.MinVCPU, num)
```

types/host_requirements.go
```diff
@@ -0,0 +1,89 @@
...
+		switch h.MinFreeMemoryGbFailAction {
+		case FailBehaviourError:
+			return fmt.Errorf(message)
+		case FailBehaviourLog:
+			log.Error(message)
+		default:
+			log.Error(message)
+		}
```
hellt, line 54: looks like this is not needed, since we default to FailBehaviorLog in the constructor

types/host_requirements.go
```diff
@@ -0,0 +1,89 @@
...
+	if valid, num := h.verifyMinFreeMemory(); !valid {
+		message := fmt.Sprintf("the defined minimum free memory based on the nodes in your topology is %d GB whilst only %d GB memory is free", h.MinFreeMemoryGb, num)
+		switch h.MinFreeMemoryGbFailAction {
+		case FailBehaviourError:
+			return fmt.Errorf(message)
+		case FailBehaviourLog:
+			log.Error(message)
+		default:
+			log.Error(message)
+		}
```
hellt, line 66: same as https://github.com/srl-labs/containerlab/pull/1209/files#r1088767680

virt/memory.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	return 0
+}
+
+type MemoryType int
+
+const (
+	MemoryTypeTotal MemoryType = iota
+	MemoryTypeAvailable
+)
```
hellt, line 29: lets move this block before the func that uses the type

types/host_requirements.go
hellt, line 83:
```suggestion
// verifyMinVCpu verifies that the node requirement for minimum vCPU count is met. 
```

types/host_requirements.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	return nil
+}
+
+// verifyMinFreeMemory verify that the amount of free memory with the requirement
+// it returns a bool indicating if the requirement is satisfied and the amount of free memory in GB
```
hellt, line 73: I think let's call the func `verifyMinAvailMemory` as it is what we return
```suggestion
// verifyMinAvailMemory verifies that the node requirement for minimum free memory is met. 
// It returns a bool indicating if the requirement is met and the amount of available memory in GB
```

types/host_requirements.go
steiler: done

types/host_requirements.go
steiler: done

types/host_requirements.go
steiler: done

PR #1216, opened 2023-01-20
Cisco 8000 simulator support

rskorka:
Add 'c8000' node kind to support Cisco 8000 simulator docker images.


hellt:
thanks @rskorka 
great contribution

Issue #1211, opened 2023-01-19
Certificate file creation on host changed (?)


jbemmel:
It used to be that containerlab would create certificates for all containers, leaving a copy on the host.
With the recent refactoring, it looks like this has been removed?

If so, would be good to mention that as a change in behavior. In some cases (e.g. connecting to gNMI over TLS) these certificates are needed, would be nice to have an option to keep them around

Issue #746, opened 2022-01-25
Create TLS certificates with stdlib only


hellt:
Remove extra cfssl dependency by reworking the code to use standard crypto lib 
Check gnmic code base for implementation reference
https://github.com/karimra/gnmic/blob/main/utils/tls.go

Issue #150, opened 2020-11-18
delay certificates creation


karimra:
Right now, we generate certificates before the containers deployments, 
This means the mgmt IP addresses, cannot be part of the cert SAN.

What about delaying the certificates generation till after the deployment and use SRL's json-rpc to set the certificates and enable the gnmi-server ?
This also solves the issue of overwriting certificates on disk even if they are present in config.

For other nodes, we can check how certificates can be set other than at boot

hellt:
since now we provision additional config for srlinux _after_ the node has been started, it is possible to get its IP address and create the certs for IP as well

hellt:
done in #1345

PR #1219, opened 2023-01-23
introduce CertificateAuthority and CertificateStore

steiler:
This PR introduces two interfaces and a single respective implementation.
- CertificateAuthority
- CertificateStore

Certificates will for now be created for all of the referenced nodes for a topology. This is meant to change once the KindProperties (#1207) are available. Then it will be again restricted to SRL nodes.
As a next step also the introduction of a seperate config knob is planned to enable cert creation for individual nodes.


steiler:
Introducign the config knob will solve this issue #1211

steiler:
This PR will allow an easy implementation of #746 since we have proper interfaces, which the new implementation simply needs to satisfy.

steiler:
We should maybe consider pushing the certificate mangement and creation down to the node instead of trying to manage this centrally. This would provide the nodes with the ability to e.g. delay cert creation, to then request the cert also with the mgmt IP as a SAN value #150

hellt (approved):

steiler:
looking good

Issue #1221, opened 2023-01-24
Introduce `image-pull-policy` for node specification


hellt:
Introduce per-node property (with kind and defaults levels) that drives the image-pull-policy behavior

steiler:
today we are looking up the image locally and if it exists we use it. If not we pull it.
What behaviour are you after? preventing pulling or forcing the pulling? Or do we need both?

hellt:
@steiler current behavior stays (equals to IfNotPresent), we have to add Always and Never

https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

PR #1223, opened 2023-01-26
implement image-pull-policy

steiler:
implements #1221

### DIFF
--- a/types/node_definition.go
+++ b/types/node_definition.go
@@ -20,6 +20,7 @@ type NodeDefinition struct {
 	AutoRemove           *bool             `yaml:"auto-remove,omitempty"`
 	Config               *ConfigDispatcher `yaml:"config,omitempty"`
 	Image                string            `yaml:"image,omitempty"`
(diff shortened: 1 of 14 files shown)


types/node_definition.go
```diff
@@ -132,6 +133,13 @@ func (n *NodeDefinition) GetImage() string {
 	return n.Image
 }
 
+func (n *NodeDefinition) GetImagePullPolicy() string {
+	if n == nil {
+		return ""
+	}
+	return n.ImagePullPolicy
+}
+
```
hellt, line 142: @steiler in [fe6d9fd](https://github.com/srl-labs/containerlab/pull/1223/commits/fe6d9fd0dbfb259c5ce0869c4aed55e9b4d66c37) I have changed this piece to return a basic string type, since I think *NodeDefinition methods such as this one, should return the raw values users provided in the topology file, without implying defaults.

Defaults, are handled in the Topology.GetNodeImagePullPolicy instead, where we typically do that.

Check if that makes sense to you as well.

steiler: yes, looks alright!

PR #1224, opened 2023-01-27
fixing cvx CheckInterfaceName()

### LINKED ISSUE #1222: Not able to define a link with interfaces prefixed with swp
I'm trying to create a topology with a CVX node (from the lab example), however when I try to start the topology I got the message:
"Error: "sw1" interface name "swp12" doesn't match the required pattern. It should be named as ethX, where X is >0"

The topology file I'm using is:
```
name: cvx02

topology:
  nodes:
    sw1:
      kind: cvx
      image: networkop/cx:5.3.0
      runtime: docker
      binds:
        - /lib/modules:/lib/modules:ro # for kernel modules like ebtables
        - sw1/interfaces:/etc/network/interfaces
    h1:
      kind: linux
      image: networkop/host:ifreload
      binds:
        - h1/interfaces:/etc/network/interfaces
      cmd: 2 # wait for 2 interfaces to be connected: eth0 + eth1

  links:
    - endpoints: ["sw1:swp12", "h1:eth1"]
```

steiler:
fix #1222

### DIFF
--- a/nodes/cvx/cvx.go
+++ b/nodes/cvx/cvx.go
@@ -113,9 +113,9 @@ func (c *cvx) GetImages(_ context.Context) map[string]string {
 
 // CheckInterfaceName checks if a name of the interface referenced in the topology file correct.
 func (c *cvx) CheckInterfaceName() error {
-	// allow ethX interface names
-	// https://regex101.com/r/C3Fhr0/2
-	ifRe := regexp.MustCompile(`swp[0-9]+$`)
+	// allow swpX interface names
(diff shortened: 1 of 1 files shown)


steiler:
@hellt I suggest making CheckInterfaceName() in default_node a noop. Just logging that the CheckInterfaceName() is not implemented for that kind.
If this is a rule that applies for all the vr kinds, that lets introduce either an _abstract_ vr kind or a general check function that all the vr types call.

hellt:
thanks @steiler

PR #1237, opened 2023-02-02
Update root password for vr-vqfx

kvklink:
Update default password from `admin:admin@123` to `root:Juniper`, in accordance to the documentation [here](https://github.com/Juniper/vqfx10k-vagrant/blob/master/TROUBLESHOOTING.md)


kvklink:
I stand corrected! Thank you for the reference 😄

PR #1238, opened 2023-02-02
central topopaths

steiler:


steiler:
until now all the calculation of paths is done in different places. This PR introduces a TopoPaths struct that only holds references to  three things.
- Path to the actual topology file
- path to the working dir (where the configs are stored etc.)

All the other required paths are then generated from these, using a set of constants and rules.

The path calculation is hereby streamlined and unified.

next is to probably also feed the TopoPath into the Pre- and PostDeploy methods... But lets do so in the next step.

If someone would want to change the directory layout of clab, this would be possible by introducing an interface on top of this PR, and providing a different interface implementation thereby returning different paths.

Issue #1235, opened 2023-01-31
Bump grpc requests limit for SR Linux


hellt:
default limit of 60 req/min is too low in certain cases, need to enhance the default config and bump the limit

```
  "gnmi-server": {
    "admin-state": "enable",
    "rate-limit": 65000,
    "commit-save": false,
    "trace-options": [
      "request",
      "response",
      "common"
    ],
```

PR #1239, opened 2023-02-03
update srl gnmic request limits

steiler:
implementing #1235

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -11,7 +11,7 @@ on:
   workflow_dispatch:
 
 env:
-  GOVER: 1.18.4
+  GOVER: 1.19.5
   CGO_ENABLED: 0
   MKDOCS_INS_VER: 9.0.4-insiders-4.27.1-hellt
   GORELEASER_VER: v1.11.4
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -11,7 +11,7 @@ on:
   workflow_dispatch:
 
 env:
-  GOVER: 1.19.5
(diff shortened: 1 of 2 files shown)


.github/workflows/cicd.yml
```diff
@@ -11,7 +11,7 @@ on:
   workflow_dispatch:
 
 env:
-  GOVER: 1.18.4
+  GOVER: 1.19.5
```
hellt, line 14: Hi @steiler 
can you revert this for now. We will move to 1.20 when 1.21 will come out

nodes/srl/srl.go
```diff
@@ -45,6 +45,9 @@ set / system tls server-profile clab-profile trust-anchor "{{ .TLSAnchor }}"
 set / system tls server-profile clab-profile authenticate-client false
 {{- end }}
 set / system gnmi-server admin-state enable network-instance mgmt admin-state enable tls-profile clab-profile
+set / system gnmi-server rate-limit 65000
+set / system gnmi-server commit-save false
```
hellt, line 49: let's remove this, since this is a default

.github/workflows/cicd.yml
steiler: then our tests fail ... take a look at this:
https://github.com/srl-labs/containerlab/actions/runs/4085203521/jobs/7042866028
happened to me also in the other PR

hellt: then we have to pin staticcheck to `v0.4.0` in the install command to keep 1.18 requirement

nodes/srl/srl.go
steiler: done

Issue #1244, opened 2023-02-06
Set `max-workers` to number of cores available on the host by default


hellt:
Containerlab does concurrent deployment of nodes and links and by default the amount of workers is derived from the number of nodes a topology has.

When a topology has more nodes that the host has cores it leads to contentions and goroutines starvation without bringing any benefits.

The proposal is to automatically limit the number of workers to the number of vcpus a node has, so that we don't have more workers than we have cores.

hellt:
done in #1245

PR #1245, opened 2023-02-07
default workerCount max == #vCPU

steiler:
implementing #1244


cmd/deploy.go
```diff
@@ -293,3 +289,38 @@ func setFlags(conf *clab.Config) {
...
+	// init number of Workers to the number of links and nodes
+	nodeWorkers = nodeCount
+	linkWorkers = linkCount
+
+	// default maxThreadCount to maxWorkerCount
+	maxThreadCount := maxWorkerCount
+	// if maxWorkerCount is not set (==0) use cpu count as max
+	if maxWorkerCount <= 0 {
+		// retrieve vCPU count
+		vCpus, err := cpu.Counts(true)
```
hellt, line 309: hi @steiler 
can you use this package instead https://github.com/tklauser/numcpus

easier on dependencies

steiler: Yes, done!

cmd/deploy.go
```diff
@@ -291,3 +287,41 @@ func setFlags(conf *clab.Config) {
...
+		// retrieve vCPU count
+		vCpus, err := numcpus.GetOnline()
+		if err != nil {
+			return 0, 0, err
+		}
+
+		numCPUs := uint(vCpus)
+
+		// limit node/link workers only if there is more node/links thans CPU cores available
+		if nodeCount > numCPUs {
```
hellt, line 312: Hi @steiler 
thanks for having a stab at it

What do you think about this tiny rework I did in 8b4a3bfa?

It might be subjective, but I think a `switch` makes things really easy to understand here, instead of the original maxThreadsCount var that served two purposes (both real cores count, as well as served as a default value when max workers is specified)

steiler: lgtm

Issue #1179, opened 2023-01-10
Allow to specify a path for artifacts


jk2lx:
By default containerlab stores the artifacts in the current directory.
This results in artifacts being stored in random places depending on the directory the user was at.
If an artifact path could be specified in the topology file, the artifacts would always be stored in a predictable path.

hellt:
I think we can start with having an environment variable, without touching the clab yml structure.

hellt:
Delivered in #1248
will be part of the next release

PR #1248, opened 2023-02-09
Allow adjustment of LabDir base directory via env var

steiler:
referring to #1179, this PR allows defining the `CLAB_LABS_BASE_DIR` environment variable.

Usually the lab directory is created under the current working dir (CWD). Via setting the `CLAB_LABS_BASE_DIR` env var, lab directories ("clab-TOPOLOGY_NAME") will be created under the `CLAB_LABS_BASE_DIR` defined path.


hellt:
thanks!

PR #1249, opened 2023-02-10
SR OS (vr-sros) partial config

### LINKED ISSUE #1033: make SR OS to accept startup-config in partial form
SR OS nodes must be able to accept startup config both in full form and in partial, CLI-blob-like format.
The full form is a hardcopy of a config file which is used by the node to boot from. We store this file at tftp location from which SR OS boots.

The partial config is provided in the MD-CLI format and must conform to the following structure:

1. entering into the configuration mode
2. commands to execute
3. commit command

SR OS vrnetlab code needs to be adjusted to make an analysis if the provided file contains the full config blob, or a partial command. The way to do this is under investigation.
If the full config is detected to be provided, the current startup config implementation is enforced.

If the partial config is detected, we should send the commands after the node boots, similar to how startup config is provided for XR, VMX, etc.

steiler:
Allow for partial configs for SROS by providing a file that contains ".partial." in its filename.

fix #1033

### DIFF
--- a/netconf/netconf.go
+++ b/netconf/netconf.go
@@ -7,16 +7,11 @@ package netconf
 
 import (
 	"fmt"
-	"time"
 
 	"github.com/scrapli/scrapligo/driver/netconf"
-	"github.com/scrapli/scrapligo/driver/network"
 	"github.com/scrapli/scrapligo/driver/options"
-	scraplilogging "github.com/scrapli/scrapligo/logging"
-	"github.com/scrapli/scrapligo/platform"
 	"github.com/scrapli/scrapligo/transport"
 	"github.com/scrapli/scrapligo/util"
-	log "github.com/sirupsen/logrus"
 )
 
 // SaveConfig saves the running config to the startup by means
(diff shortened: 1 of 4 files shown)


steiler:
Provide a `startup-config` with `.partial.` in the name will add the config from that file on top of the default device config, instead of doing a kind of replace-all action.

steiler:
this pr is related to issue #1033

nodes/vr_sros/vr-sros.go
```diff
@@ -84,6 +85,29 @@ func (s *vrSROS) PreDeploy(_ context.Context, _, _, _ string) error {
...
+		log.Infof("Applying partial config %s to %s", s.Cfg.StartupConfig, s.Cfg.LongName)
+		err = netconf.ApplyConfig(s.Cfg.MgmtIPv4Address, scrapliPlatformName, defaultCredentials.GetUsername(), defaultCredentials.GetPassword(), cfgSlc)
+		if err != nil {
+			return err
+		}
+		log.Infof("%s successfully configured", s.Cfg.LongName)
+	}
+	return nil
+}
+
```
hellt, line 110: I believe scrapli has a function (or option) that send configs from file already. So we can use that one, instead of reading the file ourselves.

nodes/vr_sros/vr-sros.go
```diff
@@ -129,3 +166,8 @@ func createVrSROSFiles(node nodes.Node) error {
 
 	return nil
 }
+
+// isPartialConfigFile returns true if the provided config is a partial config
+func isPartialConfigFile(c string) bool {
+	return strings.Contains(c, ".partial.")
```
hellt, line 172: I'd enforce capitalization and checking against `.PARTIAL` without a trailing dot, to make sure that `sr.cfg.partial` and `sr.partial.cfg` would be both accepted

steiler: really capital? thats files screaming at me ... lets do a strings.Lower(c) and we're fine

hellt: I mean you capitalize the string in the code to allow both lowercase and uppercase to be passed by a user.

netconf/netconf.go
```diff
@@ -48,3 +52,58 @@ func SaveConfig(addr, username, password, _ string) error {
...
+	mr, err := d.SendConfigs(config)
+	if err != nil {
+		return fmt.Errorf("failed to send command; error: %+v", err)
+	}
+
+	if mr.Failed != nil {
+		return fmt.Errorf("response object indicates failure: %+v", mr.Failed)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
steiler: Done

nodes/vr_sros/vr-sros.go
```diff
@@ -117,10 +146,13 @@ func (s *vrSROS) CheckInterfaceName() error {
...
 func createVrSROSFiles(node nodes.Node) error {
-	nodes.LoadStartupConfigFileVr(node, configDirName, startupCfgFName)
-
 	nodeCfg := node.Config()
 
+	// use default startup config load function if config in full form is provided
+	if !isPartialConfigFile(nodeCfg.StartupConfig) {
+		nodes.LoadStartupConfigFileVr(node, configDirName, startupCfgFName)
+	}
+
```
hellt, line 155: @steiler does this change look valid to you?

steiler: yes that was also my first idea until I realized that `LoadStartupConfigFileVr` is creating the node directory in the lab dir.
I did not want to have it created in multiple locations, hence I came up with the stunt as you've seen it.
Maybe we should create the directories as part of `NewDefaultNode(...)`?!?!

hellt: @steiler I don't see LoadStartupConfigFileVr creating any dirs not related to startup config?

PR #1263, opened 2023-02-21
added parameter struct to deploy methods in node interface

steiler:
Whenever we need more or different information in the pre-/post-/deploy phase all the kinds have to be changed.
I've now created 3 structs 
```
PreDeployParams 
DeployParams 
PostDeployParams
```
https://github.com/srl-labs/containerlab/pull/1263/files#diff-09882d300c601744375447cf7117fc641a310926d7701fc04deafbdbcbaff8f3R46-R54

which can easily be extended with additional values without touching any Kind method signature.

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -341,13 +341,13 @@ func (c *CLab) scheduleNodes(ctx context.Context, maxWorkers int,
 				}
 
 				// PreDeploy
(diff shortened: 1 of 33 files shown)


steiler:
DeployParams is an empty struct for now. But I thought it would be good to take this step just now instead of going through this agin in the future.

PR #1273, opened 2023-03-01
Certificate management delegated to the nodes

steiler:


hellt:
thanks @steiler 
this goes into main

PR #1274, opened 2023-03-01
Adding netconf add-license rpc to vr-vmx

steiler:


hellt (changes requested):

hellt:
I think we might wanna try reworking that to use telnet. My assumption here is that when you do 'telnet nodename 5000' you immediately open a session to the box without requiring to know username or password. With that we might patch the opening of a transport calling that command and then provisioning the license using CLI commands. 


Or, for netconf we first need to ensure that we wait till the node boots and likely need to have some vars for a user to provide a username/password

Issue #1275, opened 2023-03-02
Interface name checks are too strict, esp. for kind: ext-container but also crpd


mgreve:
The interface name checks (added around v0.35) are too strict and are causing our labs to fail deployment. For instance, for the **kind: ext-container**, the default check applies, which means interfaces must be named ethX. Such a check is unreasonably strict given that it's an ext-container kind (which could be anything).

It is also causing trouble for the **crpd** kind in our labs, where our interface names are etX. The default check also applies here since the crpd kind does not override the check.

Here's an example lab showing the problem:

```
name: tmplab

mgmt:
  network: tmp

topology:
  nodes:
    container1:
      kind: ext-container
    container2:
      kind: ext-container

  links:
    - endpoints: ["container1:swp11", "container2:et12"]
```

Deploying such a lab using v0.34 works fine, but fails using v0.35, because interfaces are not named ethX (the default interface name check). 

**Would it be possible to drop this check in a lab via a configuration parameter or something like this? **

For context, this is the default check, i.e. the failing check for ext-container and crpd container kind. I have removed the check in the below patch:
```
diff --git a/nodes/default_node.go b/nodes/default_node.go
index 2647d154..7882a3c1 100644
--- a/nodes/default_node.go
+++ b/nodes/default_node.go
@@ -175,15 +175,6 @@ func (d *DefaultNode) DeleteNetnsSymlink() error {
 // Checks the name pattern that is common for VM-based nodes. Container-based nodes override this function to implement
 // custom name checks.
 func (d *DefaultNode) CheckInterfaceName() error {
-       // allow ethX interface names
-       // https://regex101.com/r/C3Fhr0/2
-       ifRe := regexp.MustCompile(`eth[1-9][0-9]*$`)
-       for _, e := range d.Config().Endpoints {
-               if !ifRe.MatchString(e.EndpointName) {
-                       return fmt.Errorf("%q interface name %q doesn't match the required pattern. It should be named as ethX, where X is >0", d.Cfg.ShortName, e.EndpointName)
-               }
-       }
-
        return nil
 }
```

steiler:
Fully agree!
@hellt changed the implementation in the PR before merging. My view would also be, that in DefaultNode the CheckInterfaceName() should basically be a noop and every Kind that has specific requirements should be able to limit the interface name-space as specifically required.

[...]

PR #1276, opened 2023-03-02
make DefaultNode Interface Name Check a noop

steiler:
This is a quickfix for #1275 making InterfaceNameCheck on DefaultNode a NOOP

### DIFF
--- a/nodes/default_node.go
+++ b/nodes/default_node.go
@@ -171,6 +171,7 @@ func (d *DefaultNode) DeleteNetnsSymlink() error {
 }
 
 // CheckInterfaceName checks if a name of the interface referenced in the topology file is in the expected range of name values.
+// A no-op for the default node, specific nodes should implement this method.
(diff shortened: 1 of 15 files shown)


nodes/default_node.go
```diff
@@ -175,15 +174,6 @@ func (d *DefaultNode) DeleteNetnsSymlink() error {
 // Checks the name pattern that is common for VM-based nodes. Container-based nodes override this function to implement
 // custom name checks.
```
mgreve, line 175: You may want to change this comment as well as I think it could be misleading.

Btw. thanks for the link to the build in #1275.

hellt:
@steiler what if we had just a function named "vmIntfCheck" that the VR kinds would use in the override function to all use the same ethX pattern?

hellt:
Hi @steiler 
PTAL

steiler:
LGTM

PR #1282, opened 2023-03-06
Allow StartupConfig to be an http or https url

steiler:
Of your configs are e.g. stored in a git repo with http(s) UI or pastebin or something along those lines, you can now define a raw url as the startup-config.
This will result in clab creating a clab-download in your $TEMP directory (e.g. /tmp/clab-download) where the referenced files will be downloaded before the nodes are being started.
If you want to redeploy with new configs e.g. the git branch evolved, simply add the `--reconfigure` flag to the clab deploy command.

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -224,37 +224,17 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 	nodeCfg.Env = utils.MergeStringMaps(envFileContent, nodeCfg.Env)
 
 	log.Debugf("node config: %+v", nodeCfg)
-	// initialize config
-	p, err := c.Config.Topology.GetNodeStartupConfig(nodeCfg.ShortName)
+
+	// process startup-config
+	err = c.processStartupConfig(nodeCfg)
 	if err != nil {
 		return nil, err
 	}
 
-	// download http or https referenced configs
-	if strings.HasPrefix(p, "http://") || strings.HasPrefix(p, "https://") {
-		tmpLoc := c.TopoPaths.StartupConfigDownloadDir()
-		utils.CreateDirectory(tmpLoc, 0755)
-		// Try to deduce a filename from the url
-		postfix := utils.CalcFilename(p)
-		// Deduce the absolute destination filename for the downloaded content
-		absDestFile := c.TopoPaths.StartupConfigDownloadFileAbsPath(nodeCfg.ShortName, postfix)
-		log.Debugf("Fetching startup-config %q for node %q storing as %q", p, nodeCfg.ShortName, absDestFile)
-		// download the file to tmp location
-		err := utils.CopyFileContents(p, absDestFile, 0755)
-		if err != nil {
-			return nil, err
-		}
-		// adjust the nodeconfig by pointing startup-config to the local downloaded file
-		p = absDestFile
-	}
-
-	// resolve the startup config path to an abs path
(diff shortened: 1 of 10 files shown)


clab/config.go
```diff
@@ -229,6 +231,21 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 	if err != nil {
 		return nil, err
 	}
+
+	// download http or https referenced configs
+	if strings.HasPrefix(p, "http://") || strings.HasPrefix(p, "https://") {
+		// download the file to tmp location
+		tmpLoc := filepath.Join(os.TempDir(), "clab-downloads")
+		utils.CreateDirectory(tmpLoc, 0755)
+		filename := fmt.Sprintf("%s-%d.download", nodeCfg.ShortName, time.Now().Unix())
```
hellt, line 240: @steiler I think the filename should be the last element of the http-referenced file concatenated with the date in the format of `YYYY-MM-DD-HH:MM:SS`

steiler: I'm not sure if I want to agree with that procedure in any case.
To me it made the most sense to have the file downloaded under the name of the node it belongs to. We can argue if the timestamp makes much sense but here is why I chose this approach.

So if you go and download a raw config from e.g. a pastebin (see https://pastebin.com/raw/bFyBKJBU). Then the name makes no sense at all. Anyways we could still take this as the filename. However what if there are GET parameters attached because files come from some other crude inventory tooling. Then the Basefile is all the same and it is the parameters that define the filename.
We could also honor the `Content-Disposition: attachment; filename="filename.jpg"` http-header. But thats not always there and would therefore introduce multiple different naming styles and schemas. My vote is against it.

So reasonable to me is only naming the files after the nodes that they should get applied too. Again, the timestamp we can discuss about.

Does that make sense to you? Any other ideas or concept you have in mind?

hellt: let's maybe make it `<lab-name>-<node-name>-<file-name>` and ditch timestamp?
timestamp allows you to keep local copies of the remote file, but I don't think this is desired. If the lab topo refers to a remote location that you basically WANT it to follow whatever contents is available remotely, so timestamping is not really needed.

What is needed is a way distinct remote files for nodes from different labs. For that reason lab name and node name are part of the filename.

If the filename is garbage, like ID of sorts, it is not a big deal, you don't really care, since you desired to upload it under that name eventually.

I expect most users using some git service, where file names are part of the fetching URI after all. (most commonly github)

steiler:
So I realized that the download paths etc should come from topopaths as well, to source all our paths from there.
hence the calculation of also the filename is done there now and follows the pattern you've suggested. If for whatever reason an error occured when parsing the provided url or the resulting string is one of zero length, the filename / postfix component of the filename will fall back to `CalcFilename_Undefined = "undefined"`.

hellt:
Hey @steiler 
Thanks, this is a great addition. I did a bit of a refactoring, ptal if that makes sense to you.

The testing harness has been a bit simplified to allow easy local testing. If you have rf installed, then to run tests locally you do

1. `make build` to build a binary from the current tip of a branch
2. CLAB_BIN=bin/containerlab bash tests/rf-run.sh docker tests/02-basic-srl/01-two-srls.robot

And that will fire off tests locally

steiler:
I've always simply set an alias to the local build, that works equally fine, anyway.

LGTM

PR #1283, opened 2023-03-10
Respect network-mode being set to none for nodes on the docker runtime.

jlclx:
This is useful for selectively preventing docker's automatic default kernel route injection, as well as disabling connection to the management interface.

This makes it easier to test isolation scenarios, control lab egress, and prevent confusion over unexpected default kernel route injection.


jlclx:
Closing to redo PR from a different branch.

PR #1284, opened 2023-03-10
Respect network-mode being set to none for nodes on the docker runtime.

jlclx:
This is useful for selectively preventing docker's automatic default kernel route injection, as well as disabling connection to the management network.

This makes it easier to test isolation scenarios, control lab egress, and prevent confusion over unexpected default kernel route injection.


hellt:
Thank you @jlclx 
This looks good. I've added docs and merging this,

PR #1285, opened 2023-03-10
Permit per endpoint MTU settings.

jlclx:
Allow configuration of per endpoint MTU.

Useful for testing MTU related behavior, or working around differing platforms.

Example:
```yaml
topology:
  links:
    - endpoints: ["router2:eth2", "router3:eth1"]
      mtu: 1500
```


hellt:
Good stuff, thanks @jlclx

PR #1290, opened 2023-03-13
Test ubuntu 20.04 with podman since it fails with 22.04 for static ip addresses

hellt:


hellt:
this PR was just to test if podman tests will pass in ubuntu 20.04

Issue #1295, opened 2023-03-15
FR: Add flag to deploy subset of nodes from topology


jbemmel:
When deploying on GitHub CI/CD, resources are limited and deployment of a full 3-layer CLOS topology with 16 nodes - representing a digital twin of a real deployment - fails.

For testing purposes, I'd like to be able to select a representative subset ( say leaf1, leaf2, spine1, host1, host2 ) to validate the topology, without maintaining 2 separate topology files

Using Netlab I can do '--limit leaf1,leaf2,...' and it will work as expected, an optional deploy flag with a list of node names could do the same for Containerlab

jbemmel:
Some thoughts:
```deploy.go
	deployCmd.Flags().StringVarP(&nodesSubset, "subset", "", "",
		"optional comma-separated list of subset of nodes to deploy")
```

If a subset is defined, remove any other nodes and any links that refer to nodes outside of the subset from the topology

PR #1296, opened 2023-03-15
Jvb add deploy filter

jbemmel:
related to https://github.com/srl-labs/containerlab/issues/1295

Note that the '--filter' option applies to several (most) commands, and could/should be generalized (just like --topology)

Note also that the link filtering logic may need updating; corner cases like a node connected to a bridge are not handled

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -100,12 +100,12 @@ func WithKeepMgmtNet() ClabOption {
 	}
 }
 
-func WithTopoFile(file, varsFile string) ClabOption {
+func WithTopoFile(file, varsFile string, deployFilter []string) ClabOption {
 	return func(c *CLab) error {
 		if file == "" {
 			return fmt.Errorf("provide a path to the clab topology file")
 		}
-		if err := c.GetTopology(file, varsFile); err != nil {
+		if err := c.GetTopology(file, varsFile, deployFilter); err != nil {
 			return fmt.Errorf("failed to read topology file: %v", err)
 		}
 
--- a/clab/file.go
+++ b/clab/file.go
@@ -15,8 +15,10 @@ import (
 
 	"github.com/hairyhenderson/gomplate/v3"
 	"github.com/hairyhenderson/gomplate/v3/data"
+	"k8s.io/utils/strings/slices"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/srl-labs/containerlab/nodes"
 	"github.com/srl-labs/containerlab/types"
 	"github.com/srl-labs/containerlab/utils"
 	"gopkg.in/yaml.v2"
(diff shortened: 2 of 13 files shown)


clab/file.go
```diff
@@ -76,6 +77,40 @@ func (c *CLab) GetTopology(topo, varsFile string) error {
...
+				if slices.Contains(nodeFilter, ep1) && slices.Contains(nodeFilter, ep2) {
+					log.Debugf("Including link %+v", l)
+					newLinks = append(newLinks, l)
+				} else {
+					log.Debugf("Excluding link %+v between %s and %s", l, ep1, ep2)
+				}
+			}
+		}
+		c.Config.Topology.Links = newLinks
+	}
```
hellt, line 112: we have to make this a function and add a test

clab/clab.go
```diff
@@ -100,12 +100,12 @@ func WithKeepMgmtNet() ClabOption {
...
-func WithTopoFile(file, varsFile string) ClabOption {
+func WithTopoFile(file, varsFile string, deployFilter []string) ClabOption {
 	return func(c *CLab) error {
 		if file == "" {
 			return fmt.Errorf("provide a path to the clab topology file")
 		}
-		if err := c.GetTopology(file, varsFile); err != nil {
+		if err := c.GetTopology(file, varsFile, deployFilter); err != nil {
 			return fmt.Errorf("failed to read topology file: %v", err)
 		}
```
hellt, line 110: @steiler to me it seems this has to not have a dependency on WithTopoFile if possible.
WithTopoFile just loads the topo from a file (with vars support).
The nodes filtering should be disjoint

hellt:
thanks @jbemmel, this is a very good start.

@steiler and can take it from here, I suppose.

steiler:
This is a positive match, would we also want to be able to invert the filter to be able to exclude nodes?
Are we fine with exact matches or is there a need to get regex or the like in as well?

hellt:
@steiler I think both enhancements are dope, can be done later if requires some significant uptake

steiler:
Alright, will take a closer look tomorrow.

cmd/deploy.go
```diff
@@ -72,17 +75,20 @@ func init() {
...
+	deployCmd.Flags().StringSliceVarP(&deployFilter, "filter", "", []string{},
+		"comma separated list of nodes to include, optional")
 }
 
 // deployFn function runs deploy sub command.
 func deployFn(_ *cobra.Command, _ []string) error {
 	var err error
 
-	log.Infof("Containerlab v%s started", version)
+	log.Infof("Containerlab v%s started, %+v", version, deployFilter)
```
steiler, line 87: I'd leave this as is, the clab/file.go -> GetTopology() properly logs the filter.

docs/cmd/deploy.md
```diff
@@ -77,6 +77,10 @@ Following values are accepted, ordered from most verbose to least: `trace`, `deb
 
 It should be useful to enable more verbose logging when something doesn't work as expected, to better understand what's going on, and to provide more useful output logs when reporting containerlab issues, while making it more terse in production environments.
 
+#### filter
+
+The local `--filter` flag allows a user to specify a subset of nodes from the topology to deploy, instead of all (default). This can be useful e.g. in CI/CD test case scenarios, where resource constraints may prohibit the deployment of a full representative topology.
```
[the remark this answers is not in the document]
steiler: we should also call the field in the cmd/<command>.go files all the same ... nodeFilter is my suggestion, since configFilter and deployFilter are not really clear what they do and it is tedious to match names for the same functionality between files.

clab/clab.go
steiler: So it should be a WithNodeFilter, is that what you're proposing?!

hellt: This is a great idea. It is in line with customization options we have embedded with WithXXX options

hellt:
I am merging this to `node-filtering` branch so that all can create subsequent PRs to this base branch

cmd/deploy.go
```diff
@@ -72,17 +75,20 @@ func init() {
 	deployCmd.Flags().BoolVarP(&skipPostDeploy, "skip-post-deploy", "", false, "skip post deploy action")
 	deployCmd.Flags().StringVarP(&exportTemplate, "export-template", "",
 		defaultExportTemplateFPath, "template file for topology data export")
+
+	deployCmd.Flags().StringSliceVarP(&deployFilter, "filter", "", []string{},
+		"comma separated list of nodes to include, optional")
```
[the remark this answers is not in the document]
jbemmel: For consistency, ideally all command line flags should be also available as ENV variables

PR #1297, opened 2023-03-16
Throw an error if GetContainers does not find any containers

steiler:
lets hava a discussion if we should throw an error if on the GetContainers there is no match and no nodes are found.
I have the feeling that we should throw an error, that can be handled be the caller if required.

### DIFF
--- a/nodes/default_node.go
+++ b/nodes/default_node.go
@@ -133,6 +133,11 @@ func (d *DefaultNode) GetContainers(ctx context.Context) ([]runtime.GenericConta
 	if err != nil {
 		return nil, err
 	}
+	// check that we retrieved some container information
+	// otherwise throw ErrContainersNotFound error
+	if len(cnts) == 0 {
+		return nil, fmt.Errorf("Node: %s. %w", d.GetContainerName(), ErrContainersNotFound)
+	}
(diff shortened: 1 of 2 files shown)


nodes/default_node.go
```diff
@@ -133,6 +133,9 @@ func (d *DefaultNode) GetContainers(ctx context.Context) ([]runtime.GenericConta
 	if err != nil {
 		return nil, err
 	}
+	if len(cnts) == 0 {
+		return nil, fmt.Errorf("No containers found for node %q", d.GetContainerName())
+	}
```
hellt, line 138: let's make an error type for that and wrap it with `%w` so that the caller can do error.Is() and check if the error is indeed about no containers found to make informed decision about how to handle that

steiler: Totally agree, appeared a little to late to me.

hellt: what I meant was

```go
// ErrContainersNotFound is returned when no containers are found.
var ErrContainersNotFound = errors.New("containers not found")

return nil, fmt.Errorf("Node: %s. %w", d.GetContainerName(), ErrContainersNotFound)
```

PR #1298, opened 2023-03-16
Node filtering

### LINKED ISSUE #1295: FR: Add flag to deploy subset of nodes from topology

steiler:
This is the continuation of #1296

fix #1295

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -16,13 +16,13 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/srl-labs/containerlab/cert"
 	"github.com/srl-labs/containerlab/cert/cfssl"
-	"github.com/srl-labs/containerlab/internal/slices"
 	"github.com/srl-labs/containerlab/nodes"
 	"github.com/srl-labs/containerlab/runtime"
 	_ "github.com/srl-labs/containerlab/runtime/all"
 	"github.com/srl-labs/containerlab/runtime/docker"
 	"github.com/srl-labs/containerlab/runtime/ignite"
 	"github.com/srl-labs/containerlab/types"
+	"golang.org/x/exp/slices"
 )
 
 type CLab struct {
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -16,13 +16,13 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/srl-labs/containerlab/cert"
 	"github.com/srl-labs/containerlab/cert/cfssl"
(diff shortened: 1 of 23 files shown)


clab/clab.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/srl-labs/containerlab/runtime/docker"
 	"github.com/srl-labs/containerlab/runtime/ignite"
 	"github.com/srl-labs/containerlab/types"
+	"k8s.io/utils/strings/slices"
```
hellt, line 25: our `internal/slices/slices.go` contains the `Contains` function already. No need for this import

karimra:
It would be good to have the same functionality based on group and/or labels as well.

clab/clab.go
steiler: migrated to utils.StringInSlice

jbemmel:
> It would be good to have the same functionality based on group and/or labels as well.

You mean like --node-filter group:leaves

karimra:
> > It would be good to have the same functionality based on group and/or labels as well.
> 
> You mean like `--node-filter group:leaves`? or `--node-filter label:ci-nodes`

yes but I would use the same syntax as docker `group=leaves`

clab/clab.go
```diff
@@ -113,6 +114,69 @@ func WithTopoFile(file, varsFile string) ClabOption {
...
+		if containsAside && containsBside {
+			log.Debugf("Including link %+v", l)
+			newLinks = append(newLinks, l)
+		}
+	}
+	// replace the original collection of links with the links that have both endpoints in the node filter
+	c.Config.Topology.Links = newLinks
+
+	return nil
+}
```
hellt, line 178: Hi @steiler 
I changed this func a bit by removing the nested if and dealing with nodes map by deleting the elements from existing map. [ec0e1f9](https://github.com/srl-labs/containerlab/pull/1298/commits/ec0e1f99cd5f16da77e3e7d83f6ccd87963591f7)

I also carved out the function from the option func, so that we can unit test it and wrote a basic test.

PR #1307, opened 2023-03-29
Support external CA-Cert

### LINKED ISSUE #390: allow to specify path to a root-ca file
In #338 the certificate management for srl kind has been reworked as explained here https://containerlab.srlinux.dev/manual/kinds/srl/#tls

Although it is now possible to specify an existing root-ca file that will be used by containerlab to generate the node certs, it is not possible to specify a path to that file that is outside of the lab dir.

This is particularly painful, when users want to create labs where some nodes should use the same root-ca as was used to generate srl certs.
Consider the following example:

```yaml
    telegraf:
      kind: linux
      image: telegraf
      binds:
         - ~/root-ca.pem:/etc/telegraf/ca.crt:ro
```

Here, to make srl nodes to use the root-ca that was referenced in the telegraf node definition we would need to first copy that file to the Lab directory under `$labDir/ca/root-ca.pem`. That is not convenient, as it imposes an extra step that we might automate.

The question is how to specify the path to the root-ca.pem that we would like to use in the lab?

I see the following options.

### Opt 1: Labels
We can reuse docker labels to set a special label that will point to  a path of the cert file

```yaml
topology:
  defaults:
    labels:
      CLAB_ROOT_CA: ~/root-ca.pem
```

Containerlab then would take this file and copy it over to the lab directory before the nodes start

With such we don't introduce any more configuration keys, reusing labels. Also conveying information about which root-ca was used in docker labels, so clients can query this by running `inspect`

### Opt 2: New `certs` element
Another approach I have in mind is to use the new configuration blob on the main clab file level.

```
name: lab
certs:
  root-ca: ~/root-ca.pem
  # potentially more elements under certs will be possible to add along the way
``` 

The logic would still be the same: copy the file referenced under `.certs.root-ca` to the `$labDir/ca/root-ca.pem`

I wonder what others think?

/cc @karimra @steiler @jbemmel

steiler:
This PR in general enabled the usage of an external CA, which can reside (Cert- and Key-File) in a directory that's different from the ordinary clab structure ($labdir/.tls/ca.{key,pem}).
A ClabOption named `clab.WithExternalCA(certFile, keyFile)` is introduced that sets the files in the topopaths.
Further to test and make use of this option the environement variables:
`CLAB_CA_CERT_FILE` and `CLAB_CA_KEY_FILE` are introduced, to provide these filepaths.

As discussed in #390 there are multiple options on how to provide these values. Implementing the others should be a peace of cake now. If multiple should be supported, a quick discussion on their precedence might be necessary.

fix #390


cert/csr_input.go
```diff
@@ -8,6 +8,7 @@ type CACSRInput struct {
 	Organization     string
 	OrganizationUnit string
 	Expiry           string
+	KeyLength        int
```
hellt, line 11: Let's call it KeySize

cmd/deploy.go
```diff
@@ -92,6 +93,9 @@ func deployFn(_ *cobra.Command, _ []string) error {
 		),
 		clab.WithDebug(debug),
 	}
+
+	opts = PrepareExternalCAOptionEnviron(opts)
```
hellt, line 97: looks like this should be called inside `certificateAuthoritySetup(c)`

hellt: or maybe not... But then shouldn't the WithExternalCA option be invoked when settings.certificate-authority.cert and key are set?

cert/csr_input.go
steiler: Done

cmd/deploy.go
steiler: Thats a chicken-and-egg problem. The topo file is read by the clab instance. But the options are applied right after instantiation so the clab instance is already there and initialized.

protonjhow:
@hellt do you have a further opinion on this one? We would like to use this if we can

hellt:
@protonjhow you can download the binary with this feature using the following command

```
docker run --rm -v $(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 pull ghcr.io/srl-labs/clab-oci:5eba9183
```

this will download `containerlab` bin in your $pwd.
I still have to review this PR, so the next GA release with this feature is still TBD

protonjhow:
good news!

we arent trying to put undue pressue here, just tracking it. 3 mo seemed a little stale, but looks like there has been lots happening in the background.

we can wait for GA, whenever that is, just pleased it is coming i guess

clab/config.go
```diff
@@ -46,11 +46,23 @@ type Config struct {
 	Prefix   *string         `json:"prefix,omitempty"`
 	Mgmt     *types.MgmtNet  `json:"mgmt,omitempty"`
 	Topology *types.Topology `json:"topology,omitempty"`
+	Settings *Settings       `yaml:"settings"`
```
hellt, line 49: This should be part of the Topology struct, and parsed via appropriate method

I agree with the name `settings`

steiler: done

steiler:
doc is still missing

types/settings.go
```diff
@@ -0,0 +1,19 @@
...
+	KeySize          int    `yaml:"key-size"`
+	ValidityDuration string `yaml:"validity-duration"`
+}
+
+type LinkConfig struct {
+	Endpoints []string
+	Labels    map[string]string      `yaml:"labels,omitempty"`
+	Vars      map[string]interface{} `yaml:"vars,omitempty"`
+	MTU       int                    `yaml:"mtu,omitempty"`
+}
```
hellt, line 19: this seems like a wrong paste? LinkConfig iirc belongs to links package

nodes/srl/srl.go
```diff
@@ -167,10 +167,11 @@ func (s *srl) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
...
 	// force cert generation for SR Linux nodes
 	if s.Cfg.Certificate == nil {
-		s.Cfg.Certificate = &types.CertificateConfig{
-			Issue: true,
-		}
+		// if certificate config deos not exist, create it
+		s.Cfg.Certificate = &types.CertificateConfig{}
 	}
+	// et issue to true, to finally force cert creation
+	s.Cfg.Certificate.Issue = true
```
hellt, line 174: why changing this?

types/topo_paths.go
```diff
@@ -29,9 +31,11 @@ var clabTmpDir = filepath.Join(os.TempDir(), ".clab")
...
 // generally all these paths are deduced from two main paths. The topology file path and the lab dir path.
 type TopoPaths struct {
-	topoFile string
-	labDir   string
-	topoName string
+	topoFile           string
+	labDir             string
+	topoName           string
+	explicitCaCertFile string // if an external CA certificate is used the name to the Cert file is stored here
+	explicitCaKeyFile  string // if an external CA certificate is used the name to the Key file is stored here
```
hellt, line 38: based on the description these fields should rather be named `externalCa...`

hellt: and is it just a name that is stored there or a path? The comment might need to be addressed

nodes/srl/srl.go
steiler: so my thinking was, that we need to force and overrule the user if he "stupidly" decides to not generate a cert for an srl node to not create a certificat.
so if the user set 
```
certificate:
  issue: false
``` 
The s.Cfg.Certificate would not be nil, so we would ingest the possible `issue: false` value. Hence the existence of the struct `== nil` is decoupled from force setting the value to true.

types/topo_paths.go
steiler: well there was still the discussion I tried to start in https://github.com/srl-labs/containerlab/pull/1307#issuecomment-1489906045.
so if the ca-key and ca-cert are explicitly set but they do not exist, we throw an error right now.
The other option would be to create CA key / cert under the defined, explicitly set path...
That was basically the thought process behind the name.
So two decisions to make.
1. keep the actual implementation, throwing the error or generating the CA under the specified location.
2. under the consideration in 1. rename or not

types/settings.go
steiler: you are right, sorry! Fixed.

docs/manual/cert.md
```diff
@@ -1,13 +1,52 @@
...
+
+### Node certificates
+
+The decision to generate node certificates is driven by either of the following two parameters:
 
-Apart from automated pipeline for certificate provisioning, containerlab exposes the following commands that can create a CA and node's cert/key:
+1. node kind
+2. `issue` boolean parameter under `node-name.certificate` section.
+
+For SR Linux nodes the `issue` parameter is set to `true` by default. For other node kinds the `issue` parameter is set to `false` by default and can be [overridden](nodes.md#certificate) by the user.
```
steiler, line 29: Not just by default, but forcefully set to true.

docs/manual/cert.md
```diff
@@ -1,13 +1,52 @@
...
 * [`tools cert sign`](../cmd/tools/cert/sign.md) - creates certificate/key for a host and signs the certificate with CA
 
-With these two commands users can easily create CA node certificates and secure the transport channel of various protocols. [This lab](https://clabs.netdevops.me/security/gnmitls/) demonstrates how with containerlab's help one can easily create certificates and configure Nokia SR OS to use it for secured gNMI communication.
\ No newline at end of file
+With these two commands users can easily create CA node certificates and secure the transport channel of various protocols. [This lab](https://clabs.netdevops.me/security/gnmitls/) demonstrates how with containerlab's help one can easily create certificates and configure Nokia SR OS to use it for secured gNMI communication.
+
+## External CA
+
+Users who require more control over the certificate generation process can use an existing external CA. Containerlab needs to be provided with the CA certificate and key. The CA certificate and key must be provided via `.settings.certificate-authority.[key]|[cert]` configuration parameters.
```
steiler, line 42: It is actually also possible to provide the key and cert via ENV vars:
`CLAB_CA_CERT_FILE` and `CLAB_CA_KEY_FILE`

cmd/tools_cert.go
```diff
@@ -60,6 +61,7 @@ func init() {
 	signCertCmd.Flags().StringVarP(&path, "path", "p", "",
 		"path to write certificate and key to. Default is current working directory")
 	signCertCmd.Flags().StringVarP(&certNamePrefix, "name", "n", "cert", "certificate/key filename prefix")
+	signCertCmd.Flags().IntVarP(&keySize, "keysize", "", 2048, "private key size")
```
hellt, line 64: note to self: add doc in tools cmd

docs/manual/cert.md
hellt: agree, fixed

types/topo_paths.go
hellt: discussed offline, the paths to ext CA and Key should be provided and be valid

nodes/srl/srl.go
hellt: the struct is already initialized based on the Topology methods, so we just needed to set the field to *bool(true)

cmd/deploy.go
hellt: I removed the option entirely and set the env var values in the existing `certificateAuthoritySetup`

clab/config.go
hellt: I was dumb. This belongs at the root of the clab file. I moved it back, sorry.

PR #1308, opened 2023-03-29
Use golang standard lib to generate certs

steiler:


cert/stdlib/cert_auth_stdlib.go
```diff
@@ -0,0 +1,157 @@
...
+		SerialNumber: big.NewInt(1658),
+		Subject: pkix.Name{
+			Organization:       []string{input.Organization},
+			OrganizationalUnit: []string{input.OrganizationUnit},
+			CommonName:         input.CommonName,
+			Country:            []string{input.Country},
+			Locality:           []string{input.Locality},
+		},
+		DNSNames: input.Hosts,
+		// IPAddresses:  []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback},
```
hellt, line 119: This is amazing stuff @steiler 

During my quick test I noticed that IP-based SAN's are not parsed correctly?

This commented line is likely the reason I see IP addresses being encoded as DNS SANs, and not IPs:

```
            X509v3 Subject Alternative Name: 
                DNS:srl, DNS:clab-srl-srl, DNS:srl.srl.io, DNS:172.20.20.4, DNS:2001:172:20:20::4
```

Looks like we need to have a function that iterates over `input.Hosts` and checks if an IP is passed and populate IPAddresses slice with them?

hellt: Did that in fbd941c

PR #1324, opened 2023-04-11
Vrhealthypartial

steiler:
check for /healthy == "0 running" before applying partial config

### DIFF
--- a/nodes/vr_sros/vr-sros.go
+++ b/nodes/vr_sros/vr-sros.go
@@ -11,7 +11,14 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
+	"github.com/scrapli/scrapligo/driver/network"
+	"github.com/scrapli/scrapligo/driver/options"
+	scraplilogging "github.com/scrapli/scrapligo/logging"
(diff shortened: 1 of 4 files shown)


nodes/vr_sros/vr-sros.go
```diff
@@ -171,12 +172,30 @@ func isPartialConfigFile(c string) bool {
 	return strings.Contains(strings.ToUpper(c), ".PARTIAL")
 }
 
+// isHealthy checks if the "/health" file created by vrnetlab exists and contains "0 running"
+func (s *vrSROS) isHealthy(ctx context.Context) bool {
+	ex := exec.NewExecCmdFromSlice([]string{"bash", "-c", "cat /health | perl -lne '/0 running/ && $MATCH++; END{exit 1 unless $MATCH>0}'"})
+
```
hellt, line 178: maybe let's keep perl in the grave and use `grep '0 runnning' /health`?

steiler:
integrated health check into vrsros-partial-config

nodes/vr_sros/vr-sros.go
steiler: done

Issue #1316, opened 2023-04-07
Allow providing startup-config inline


hellt:
Currently, startup-config can only be a file or HTTP path.
By allowing users providing startup-config inline we can make self-contained topology files.

I propose to have a simple logic to check if the value passed in `startup-config` is an inline value. Inline configs must always be provided like this

```yaml
startup-config: |
  /configure "test"
kind: sros
```

With that syntax, the multiline string will keep all newlines (and most importantly attach a newline at the end).

<img width="1031" alt="image" src="https://user-images.githubusercontent.com/5679861/230564669-78057aad-c358-4220-9fcf-036722e17541.png">

So we can test if a value contains `\n` to figure out if this value means inline config.

PR #1326, opened 2023-04-11
allow for embedding node configs in the topology file

steiler:


steiler:
implementing #1316

clab/config.go
```diff
@@ -270,30 +270,49 @@ func (c *CLab) processStartupConfig(nodeCfg *types.NodeConfig) error {
...
+		tmpLoc := c.TopoPaths.ClabTmpDir()
 		utils.CreateDirectory(tmpLoc, 0755)
 
-		// get file name from an URL
-		fname := utils.FilenameForURL(p)
+		switch {
+		case isEmbeddedConfig:
+			// we have an embedded config with multiple newlines
+			// its an embedded config
+			absDestFile := c.TopoPaths.StartupConfigDownloadFileAbsPath(nodeCfg.ShortName, "embedded.config")
```
hellt, line 288: I think this must be named `embedded.partial.cfg` because full configs are not good for embedding, and we have to treat them as partial

steiler: they are not good ... but they work.
So your opinion is to have partial configs provided inline?

hellt: yes, the idea was to support embedding partials

Issue #1328, opened 2023-04-12
sr os kind doesn't check if a license file exists


hellt:
This is a regression.

PR #1330, opened 2023-04-13
introduce license required knob

steiler:
In default node we introduce `LicenseRequired` knob, to indicate that a kind requires a license file to function properly. The knob is taken into account in `Node.VerifyLicenseFileExists()`

### DIFF
--- a/nodes/default_node.go
+++ b/nodes/default_node.go
@@ -30,6 +30,8 @@ type DefaultNode struct {
 	Mgmt             *types.MgmtNet
 	Runtime          runtime.ContainerRuntime
 	HostRequirements *types.HostRequirements
+	// Indicates that the node should not start without no license file defined
+	LicenseRequired bool
 	// OverwriteNode stores the interface used to overwrite methods defined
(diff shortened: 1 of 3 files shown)


steiler:
implementing #1328

nodes/default_node.go
```diff
@@ -30,6 +30,8 @@ type DefaultNode struct {
 	Mgmt             *types.MgmtNet
 	Runtime          runtime.ContainerRuntime
 	HostRequirements *types.HostRequirements
+	// Indicates that the node should not start without no license file defined
+	LicenseRequired bool
```
hellt, line 34: @steiler 
maybe we should have a string field `LicensePolicy` with values

* "" - not required
* "required" - fail with an error saying license is required
* "warn" - emit a warning if not specified

"warn" - is for SR OS case, where we allow users to run it without a license, assuming they have it embedded in the image.

steiler: checkout the new commit, please.

hellt:
Thanks @steiler !

Issue #1331, opened 2023-04-14
Allow web URLs for files consistently


jbemmel:
Containerlab has a bunch of places where external files are referenced. In some cases (like startup-config) this can be a URL, but in other cases (like srl-agents) this can only be local files

It would be helpful if any external files referenced could be located anywhere, either on the web or on the local file system. Specifically in case of agent YAML files, this would allow the user to reference a centrally maintained/updated file for an agent, as opposed to having to create a local copy (which becomes stale)

steiler:
For the agent config, would we need to force a specific file ending?
we have functionality that extracts the filename from the last url part and the content-disposition header.
However if this agent-config would maybe come from a pastebin, ther will very likely be no or no proper file ending.
we could enforce so.
opinions @hellt, @jbemmel ?

jbemmel:
The agent config is (should be) YAML, so if the URL does not end with ".yml" we could add that

hellt:
done in #1336

PR #1336, opened 2023-04-17
Allow for agent configs to be downloaded from http(s) urls

steiler:
implementing #1331

### DIFF
--- a/nodes/srl/srl.go
+++ b/nodes/srl/srl.go
@@ -218,14 +218,9 @@ func (s *srl) PreDeploy(_ context.Context, params *nodes.PreDeployParams) error 
 			if utils.IsHttpUri(fullpath) {
 				basename = utils.FilenameForURL(fullpath)
 			}
-			// check for .yaml or .yml extension
-			// add if not existent
-			pos := strings.LastIndex(basename, ".")
-			if pos == -1 {
+			// enforce yml extension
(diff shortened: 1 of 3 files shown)


hellt:
Hi @steiler I am ok with that change, can you rebase it on current `main` so that srl pipeline gets to working status again?

steiler:
ok, it is rebased.
I've also added some code that will add the .yml extension if no filename extension is provided but also change it to .yml if it is not already .yaml or .yml

nodes/srl/srl.go
```diff
@@ -209,11 +209,20 @@ func (s *srl) PreDeploy(_ context.Context, params *nodes.PreDeployParams) error
...
 		for _, fullpath := range agents {
 			basename := filepath.Base(fullpath)
+			// if it is a url extract filename from url or content-disposition header
+			if utils.IsHttpUri(fullpath) {
+				basename = utils.FilenameForURL(fullpath)
+			}
+			// enforce yml extension
+			if ext := filepath.Ext(basename); !(ext == ".yml" || ext == ".yaml") {
+				basename = basename + ".yml"
+			}
```
hellt, line 224: @steiler this is likely a simpler way?

steiler: Rome and all the streets that lead there and so ... ;-)

PR #1341, opened 2023-04-20
set interface and subinterface admin-state up in tests and examples

steiler:


hellt:
@steiler hi. Thanks for looking into this. I think you only need to add the admin state for ipv4/6 levels

lab-examples/srl02/srl1.cfg
```diff
@@ -1,5 +1,7 @@
 set / interface ethernet-1/1
+set / interface ethernet-1/1 admin-state enable
 set / interface ethernet-1/1 subinterface 0
+set / interface ethernet-1/1 subinterface 0 admin-state enable
 set / interface ethernet-1/1 subinterface 0 ipv4
 set / interface ethernet-1/1 subinterface 0 ipv4 address 192.168.0.0/31
 set / interface ethernet-1/1 subinterface 0 ipv6
```
hellt, line 7: @steiler what I meant by the comment above is that here you don't need to add those admin-state statements, rather you need to add it under `subinterface 0 ipv4 admin-state enable` and `subinterface 0 ipv6 admin-state enable`

steiler: yes thanks got that.
checkout new commit

PR #1345, opened 2023-04-26
SRL Post-Deploy cert creation

steiler:
create srl certificate in post-deploy phase, thereby including the mgmt IPs as SANs

```
mava@server01:~/projects/containerlab$ openssl s_client -connect 172.20.20.2:57400 </dev/null 2>/dev/null | openssl x509 -noout -text | grep DNS:
                DNS:srl1, DNS:clab-srl02-srl1, DNS:srl1.srl02.io, IP Address:172.20.20.2, IP Address:2001:172:20:20:0:0:0:2
```

### DIFF
--- a/nodes/srl/srl.go
+++ b/nodes/srl/srl.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
+	"github.com/srl-labs/containerlab/cert"
 	"github.com/srl-labs/containerlab/clab/exec"
 	"github.com/srl-labs/containerlab/nodes"
 	"github.com/srl-labs/containerlab/types"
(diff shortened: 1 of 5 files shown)


nodes/srl/srl.go
```diff
@@ -247,11 +248,6 @@ func (s *srl) PreDeploy(_ context.Context, params *nodes.PreDeployParams) error
 func (s *srl) PostDeploy(ctx context.Context, params *nodes.PostDeployParams) error {
 	log.Infof("Running postdeploy actions for Nokia SR Linux '%s' node", s.Cfg.ShortName)
 
-	// populate config with mgmt ip addresses
-	if err := s.UpdateConfigWithRuntimeInfo(ctx); err != nil {
-		return err
-	}
-
```
hellt: Hi @steiler, I removed this, since I think it is always performed for all nodes unconditionally in https://github.com/srl-labs/containerlab/blob/181b53924719af4a16f6781bd238de1b720a2557/cmd/deploy.go#L227

steiler: ok, cool. Yeah did not specifically check to be honest :disappointed:.

hellt:
I've added the rf test to make sure we keep checking that.
Great work, Markus, kudos

PR #1347, opened 2023-04-28
Store information about NodeGeneration in registry.

steiler:
So far generate.go contained two lists indenpendent from the nodes, that would describe nodes that (1) can be generated and (b) what the interface naming convention would be for the node kind. This will introduce seperate fields in the NodeRegistry for this information, such that the information for this sort of node generation can be defined as part of the node definition

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -216,8 +216,10 @@ func generateTopologyConfig(name, network, ipv4range, ipv6range string,
 				// create a raw veth link
 				l := &links.LinkVEthRaw{
 					Endpoints: []*links.EndpointRaw{
-						links.NewEndpointRaw(node1, fmt.Sprintf(generateNodesAttributes[nodes[i].kind].GetInterfaceFormat(), k+1+interfaceOffset), ""),
-						links.NewEndpointRaw(node2, fmt.Sprintf(generateNodesAttributes[nodes[i+1].kind].GetInterfaceFormat(), j+1), ""),
+						links.NewEndpointRaw(node1, fmt.Sprintf(
+							generateNodesAttributes[nodes[i].kind].GetInterfaceFormat(), k+1+interfaceOffset), ""),
+						links.NewEndpointRaw(node2, fmt.Sprintf(
+							generateNodesAttributes[nodes[i+1].kind].GetInterfaceFormat(), j+1), ""),
 					},
 				}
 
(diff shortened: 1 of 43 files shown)


steiler:
I have, by the way, taken over the actual view on which nodes are eligable for topology generation.
But I have the feeling, that this list could be extended, Maybe take another look at it @hellt

steiler:
takes also already vr-vsrx into account.

nodes/node_registry.go
```diff
@@ -63,31 +63,106 @@ func (r *NodeRegistry) GetRegisteredNodeKindNames() []string {
...
+	if nre.attributes == nil {
+		return nil
+	}
+	return nre.attributes.generateAttributes
 }
 
 // Credentials returns entry's credentials.
-func (e *NodeRegistryEntry) Credentials() *Credentials {
+// might return nil if no default credentials present.
+func (e *NodeRegistryEntryAttributes) Credentials() *Credentials {
```
hellt, line 100: looks like a dup of `func (nrea *NodeRegistryEntryAttributes) GetCredentials() *Credentials`

cmd/generate.go
```diff
@@ -8,121 +8,20 @@ import (
...
-
-	// Add SR Linux kinds
-	for _, k := range srl.KindNames {
-		m[k] = "e1-%d"
-	}
-
-	// Add Arista cEOS kinds
-	for _, k := range ceos.KindNames {
-		m[k] = "eth%d"
-	}
```
hellt, line 48: TODO: align interfaces names given some of them are using the "native" format, which was introduced after this PR existed.

nodes/node_registry.go
```diff
@@ -63,31 +63,106 @@ func (r *NodeRegistry) GetRegisteredNodeKindNames() []string {
...
+func (nrea *NodeRegistryEntryAttributes) GetCredentials() *Credentials {
+	return nrea.credentials
+}
+
+func (nrea *NodeRegistryEntryAttributes) GetGenerateAttributes() *GenerateNodeAttributes {
+	return nrea.generateAttributes
+}
+
+type GenerateNodeAttributes struct {
+	generateable    bool
```
hellt, line 143: likely we don't need the `generatable` flag at all. All nodes should be generatable, I think some nodes were not generatable originally  because we just did not add them manually to the generage cmd

nodes/node_registry.go
hellt: 038d326

Issue #1355, opened 2023-05-01
Iptables rules in DOCKER-USER chain are not removed on a lab destroy


hellt:
Looks like regression after some refactoring.

The rules are not cleaned up in DOCKER-USER chain of the main table.

`iptables -vnL DOCKER-USER` - lists active rules

As an interim workaround remove them manually with:

```
iptables --line-number -nL DOCKER-USER | grep containerlab | awk '{print $1}' | tac | xargs -I % -n 1 sh -c 'iptables -D DOCKER-USER %; sleep 0.1'
```

hellt:
We will keep an eye on that, since it is not clear why I saw leftovers in DOCKER-USER

PR #1361, opened 2023-05-04
fixing the DOCKER-USER rule deletion

steiler:
fixing #1355


hellt:
strangely I don't understand why I had stale bridges since current 0.40 version removes the rules just fine...
Something to keep an eye on.

This change though makes sense as well, so merging it.

steiler:
This will also resolve the following Warnings.
These happen, since we're pointing to the wrong bridge.

WARN[0000] failed to enable LLDP on docker bridge: open /sys/class/net/br-ecdc59977784/bridge/group_fwd_mask: no such file or directory 
WARN[0000] failed to disable TX checksum offloading for the br-ecdc59977784 bridge interface: no such device

Will be fixed in v0.41.0 since it is already merged.

PR #1362, opened 2023-05-04
use google/nftables library to install iptables

steiler:


hellt:
Thanks @steiler!

Yes, nftables backend is not available on the older kernels (like centos 7), so in its current form this PR reduces the feature set. But I think I am fine with that.

Ultimately, we could maintain two backends for this rule insertion, legacy iptables for those systems that don't have nftables backend.

If we hear many complaints about that, we might consider adding an interface and using the proper implementation based on nftables availability.

Issue #1360, opened 2023-05-03
Implement a check for `ip_tables` and `ip6_tables` kernel module


hellt:
VM-based nodes require iptables module to be loaded to perform datapath stitching.
Newer OSes betray iptables and not have it loaded by default.

We have to start checking if iptables module is loaded and if not emit a warning in logs

steiler:
So what do you think should be preconditions to do the check?
Shall this be checked on every execution? Nodekind dependent or also dependent on runtime or a combination of these?

hellt:
Checking it every time. Even when no VM-based nodes are used, we use iptables to create rules for external lab access.
So on start we should check for ip_tables ip6_tables modules and emit a WARN if they are not loaded.

hellt:
In #1363 the auto-load of ip and ip6_tables kernel modules is implemented

PR #1363, opened 2023-05-04
implements function to check for loaded kernel modules & try auto-load

steiler:
implements #1360


steiler:
now auto loading the required modules, which is `ip_tables` and `ip6_tables` for now.

hellt:
Thanks @steiler!

Issue #1364, opened 2023-05-04
Customer Docker bridge is not correctly detected


mtache:
Hi,

I am using containerlab to spin up a lab on a pre-existing Docker network with a customer bridge. For the context, this is a bridge created by libvirt (so I can have my containerlab nodes and my KVM guests on the same Linux bridge).

I am creating my Docker network like this:
`docker network create --driver bridge --subnet 192.168.100.0/24 --ip-range 192.168.100.128/25 --gateway 192.168.100.1 -o "com.docker.network.bridge.name"="virbr0" lab-mgmt`

Result:
```
$ docker network inspect lab-mgmt
[
    {
        "Name": "lab-mgmt",
        "Id": "f1d000689d48b05cc6b9cc651f78b68bd33832fa8f96ba340a7e616ea2c26c2d",
        "Created": "2023-05-04T12:56:47.263459677+02:00",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": {},
            "Config": [
                {
                    "Subnet": "192.168.100.0/24",
                    "IPRange": "192.168.100.128/25",
                    "Gateway": "192.168.100.1"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": {},
        "Options": {
            "com.docker.network.bridge.name": "virbr0"
        },
        "Labels": {}
    }
]
```

My management configuration in the topology file:
``` yaml
mgmt:
  network: lab-mgmt
```

Let's create the lab:
```
$ sudo containerlab deploy --log-level trace
DEBU[0000] detected network name in use: lab-mgmt, backed by a bridge br-f1d000689d48
[truncated]
DEBU[0000] Docker network "lab-mgmt", bridge name "virbr0"
DEBU[0000] Disable RPF check on the docker host
DEBU[0000] Enable LLDP on the linux bridge br-f1d000689d48
WARN[0000] failed to enable LLDP on docker bridge: open /sys/class/net/br-f1d000689d48/bridge/group_fwd_mask: no such file or directory
DEBU[0000] Disabling TX checksum offloading for the br-f1d000689d48 bridge interface...
WARN[0000] failed to disable TX checksum offloading for the br-f1d000689d48 bridge interface: no such device
[...]

PR #1365, opened 2023-05-04
allow for ip-range definition on docker mgmt net

steiler:
As mentioned in #1364 this PR will enable ip-range assignments for docker networks.
```
mgmt:
  network: lab-mgmt
  bridge: testbr
  ipv4-subnet: 192.168.100.0/24
  ipv4-gw: 192.168.100.1
  ipv4-range: 192.168.100.128/25
```


steiler:
I should mention that I also adjusted the ipv{4,6}_subnet to ipv{4,6}-subnet to allign with ipv{4,6}-gw, which already uses the dash notation.

Issue #1314, opened 2023-04-04
`ansible-no-host-var` doesn't have effect when user-defined groups are used


hellt:
A bug report from a users says that `ansible-no-host-var` doesn't remove the ansible_hostvar when a user-defined group is provided for node(s) - https://containerlab.dev/manual/inventory/#user-defined-groups

hellt:
done in #1369

PR #1369, opened 2023-05-09
fixing ansible inventory no-host-var bug

steiler:


steiler:
Fixing #1314

hellt:
good stuff!

PR #1370, opened 2023-05-09
change mgmt_ipv4/6 to mgmt-ipv4/6

### LINKED ISSUE #749: check topology elements to use dashes in names, not underscores
cleanup topology element's name to use dashes in composite names, such as `ipv4-gw` and not `ipv4_gw`

steiler:
Besides changing dash to underscore for `ipv{4,6}_subnet` via #1365.
I think the `mgmt_ipv4/6` config stanza is the last remaining underscore topology setting.
I think, it would make sense to bundle this and #1365 in a single release, to have such breaking change just once.

This will in then in my view be enough to close #749

### DIFF
--- a/clab/test_data/topo1.yml
+++ b/clab/test_data/topo1.yml
@@ -9,11 +9,11 @@ topology:
       env:
         env1: val1
         env2: val2
-      mgmt_ipv4: 172.100.100.11
+      mgmt-ipv4: 172.100.100.11
     node2:
       kind: srl
       license: node1.lic
       user: custom
-      mgmt_ipv4: 172.100.100.12
+      mgmt-ipv4: 172.100.100.12
       labels:
         node-label: value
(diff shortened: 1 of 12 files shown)


hellt:
Yes, these changes definitely must go together.
Besides that, I think we need to create a special error message notifying people to check release notes when unmarshalling error happens for unrecognized fields

steiler:
That would be here then.
https://github.com/srl-labs/containerlab/blob/main/clab/file.go#L76
What do you propose as the message to show?

hellt:
PS. Currently I already see that this message is emitted:

```
 dclab des -c -t private/linux.clab.yml
Error: failed to read topology file: yaml: unmarshal errors:
  line 5: field ipv4-subnet not found in type types.MgmtNet
  line 6: field ipv4-range not found in type types.MgmtNet
```

PR #1387, opened 2023-05-16
Extend the JSON Schema to include SR OS type validation

kellerza:
Added all options from https://github.com/hellt/vrnetlab/tree/master/sros to the schema

Using `allOf` to ensure different schemas for srl & vr-sros. vr-sros includes a pattern fallback to `cp:xxxx` strings for custom types

image

image

### DIFF
--- a/schemas/clab.schema.json
+++ b/schemas/clab.schema.json
@@ -317,7 +317,7 @@
                     "if": {
                         "properties": {
                             "kind": {
-                                "const": "srl"
+                                "pattern": "(srl|nokia_srlinux)"
                             }
                         },
                         "required": [
@@ -353,7 +353,7 @@
                     "if": {
                         "properties": {
                             "kind": {
-                                "const": "vr-sros"
+                                "pattern": "(vr-sros|vr-nokia_sros)"
                             }
                         },
                         "required": [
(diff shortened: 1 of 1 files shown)


schemas/clab.schema.json
```diff
@@ -303,48 +303,96 @@
...
-                            "ixr6",
-                            "ixr6e",
-                            "ixr10",
-                            "ixr10e"
+                },
+                {
+                    "if": {
+                        "properties": {
+                            "kind": {
+                                "const": "srl"
```
hellt, line 320: Hi @kellerza 
good stuff, since you have a fresh memory of json schema maybe you remember if there is a possibility to have multiple values in the `if` clause?

Ther reason is, that we have more than a single kind constant:

1. `srl` and `nokia_srlinux`
2. `vr-sros` and `vr-nokia_sros`

The latter option is preferrable. But it would be cool if json schema allowed to use multiple consts without code duplication

kellerza: Maybe we could add an anyOf inside the if clause, will give it a go

kellerza:
"pattern" seems simple enough

image

image
image

schemas/clab.schema.json
kellerza: "pattern" seems simple enough

image

image
image

hellt: niiice

hellt:
Thank you!

PR #1388, opened 2023-05-16
extract ssh pub keys also from agent

steiler:


clab/authz_keys.go
```diff
@@ -39,13 +41,17 @@ func (c *CLab) CreateAuthzKeysFile() error {
...
+	// try extracting keys from ssh agent
+	keys, err := retrieveAgentKeys()
+	if err != nil {
+		log.Debug(err)
+	} else {
+		log.Debugf("extracted %d keys from ssh-agent", len(keys))
+		for _, k := range keys {
+			b.WriteString(k + "\n")
+		}
 	}
```
hellt, line 53: the `else` is not needed I think

PR #1390, opened 2023-05-16
Deprecated fields

steiler:
This is to enhance the MgmtNet and the NodeDefinition structs to also accept the depricated underscore fields


steiler:
this was mixed up stuff... closing it

PR #1391, opened 2023-05-16
Deprecate fields

steiler:
This is to enhance the MgmtNet and the NodeDefinition structs to also accept the deprecated underscore fields.


hellt:
This is a clever trick @steiler 

I just renamed vars

PR #1395, opened 2023-05-17
Link Creation supported by DependencyManager

steiler:
We where busy waiting for the nodes statuses to change to "created", before creating links for a set of nodes.
With this we will rely on DependencyManager to block the go func until the depending nodes are created.


clab/dependency_manager.go
```diff
@@ -34,19 +36,25 @@ type defaultDependencyManager struct {
 	// Names of the nodes that depend on a given node are listed here.
 	// On successful creation of the said node, all the depending nodes (dependers) wait groups will be decremented.
 	nodeDependers map[string][]string
+
+	nodeCreated map[string]*sync.WaitGroup
```
hellt, line 40: @steiler do we need a WaitGroup here? I have the following concerns

1. the `nodeCreated` name may be not a great one here, since it says that the node is either created or not. So we don't really say if it running, for example.  
2. For the future impl I think it makes sense to use the `nodeStatus` field with an enumerated string field such as `created`, `running`, `exited`, etc.
3. But if we just want to start small here and track running status, i'd propose to name this as `nodeStatus` and just operate with `""` and `"running"` string values.  
When initialized, the empty string means node status is unknown, and when the node has been running we set it to `running`

hellt: Scratch that, I see that in the WaitNode we rely on `wg.Wait()` which makes sense

hellt: so maybe just to clarify here, does dm finishes when it creates the node or when the node is running?

hellt:
moved to draft since more work will be done in that PR

clab/clab.go
```diff
@@ -457,30 +458,32 @@ func (c *CLab) scheduleNodes(ctx context.Context, maxWorkers int,
...
-		}(n, dm, concurrentChan, workerFuncChWG) // execute this function straight away
-	}
-
-	// Gate to make sure the channel is not closed before all the nodes made it though the channel
-	workerFuncChWG.Wait()
-	// close the channel and thereby terminate the workerFuncs
-	close(concurrentChan)
+	// schedule nodes via a go func to create links in parallel
+	go func() {
+		for _, n := range scheduledNodes {
```
hellt, line 463: @steiler what is the purpose of this goroutine? Does it make things faster in any way, because I see that we start a goroutine on ln467 anyhow, so I don't quite the optimization here

steiler: In 483 we wait for the nodes to make it through the channel before closing it.
That means, that the Link deployment cannot start, since the Call from Schedule nodes has to return.
See
https://github.com/srl-labs/containerlab/blob/4ea14a31b24d978ae3c005f2f299e4a9c5c9294e/cmd/deploy.go#L223-L230

clab/clab.go
```diff
@@ -511,60 +519,43 @@ func (c *CLab) WaitForExternalNodeDependencies(ctx context.Context, nodeName str
...
-					if err := c.CreateVirtualWiring(link); err != nil {
-						log.Error(err)
-					}
-				case <-ctx.Done():
-					return
+
+			var waitNodes []string
+			for _, n := range []*types.NodeConfig{li.A.Node, li.B.Node} {
+				// we should not wait for "host"
+				if n.Kind != "host" {
```
hellt, line 528: @steiler I think this should be also `ovs` and `ext-cont`?
I also wonder, maybe you want this to enhance so that each node signal SignalDone to the dep_manager so that you don't have to hard code the kinds?

steiler: Well I'm not 100% sure to be hones. But I think no, this should not be needed for `bridge`, `ovs` and `ext-cont`.
All of these kinds actually have their Node implementation and will therefore be "deployed" and a Done is being signaled.
In general, this also holds true for host, we have a Node for that.
But this is where the whole stuff is all mixed and messed up.
We have to differentiate between a node definition of kind `host` and a link endpoint that states `host` in the node part.
They can both exist but don't have to.

PR #1402, opened 2023-05-22
MACVLAN interfaces

### LINKED ISSUE #1393: Container interfaces connected to host interfaces via `macvlan`
To enable use cases where containers can connect to the external world using host interfaces the following options can be considered:

1. Use a linux bridge to bridge container's veth interface with host's interface.
    This mode has been supported by containerlab since the beginning.
    
    image
2. Using macvlan device type where a macvlan device is multiplexed on a MAC level with the host's interface

This issue tracks the development for 2nd option where a macvlan device should be added to container's namespace and coupled with the hosts interface.

The following material may assist in the implementation:

- http://networkstatic.net/configuring-macvlan-ipvlan-linux-networking/

steiler:
This is to support macvlan and macvtap interfaces in the endpoint section of the topology definition.
```
  links:
    - endpoints: ["l1:eth1", "macvlan:enp4s0f0"]
    - endpoints: ["l1:eth2", "macvtap:enp4s0f0"]
```
This PR fixes #1393.

from @hellt: I removed macvtap support in [b5dbaf4](https://github.com/srl-labs/containerlab/pull/1402/commits/b5dbaf4bf3ab51ac171ab325d587f9e0b369b5bb). If we need to add it later this commit can be reverted/re-added.


steiler:
I had to base this PR on top of #1395.

steiler:
I am not yet 100% sure if we should probably dissociate from the rather simple format we used for the endpoints and come up with some more formal struct.
We're right now forcing everything into this format of a slice with two elements... that already make up 4 pieces of information. 2x (Node + Interface)
We should have these seperated and probably also define an endpoint type and somehow get rid of the to the most extend procedural way of wiring the labs up.

Opinions? @hellt ?

hellt:
@steiler can you rebase this please on top of the main branch? I just merged #1395

steiler:
We're missing docs.
Also in my opinion we should only allow a new kind of link definition for the this type of interfaces, although this PR allows defining them via a fake node "macvlan" or "macvtap" in the links section.

clab/dependency_manager/dependency_manager.go
```diff
@@ -104,9 +104,21 @@ type defaultDependencyManager struct {
...
+	// TODO, we need to rework this stuff
+	for _, n := range []string{"macvlan", "macvtap"} {
+		dm.AddNode(n)
+		// artificially indicate that all the helper nodes have reached all the defined states
+		for _, s := range RegularNodeStates {
+			dm.SignalDone(n, s)
+		}
+	}
+
+	return dm
```
hellt, line 121: @steiler why is this needed to support macvlan, but was not needed to support `host` kind?

If there is a difference, then maybe we can use `host` kind for macvlan/tap as well and not touch depmgr?

hellt:
@steiler with #1450 merged we can move this one I believe

clab/dependency_manager/dependency_manager.go
steiler: you are right, there is always multiple ways towards rome.
For `host` and `mgmt-net` we prevented this dependency from getting inserted.
See https://github.com/srl-labs/containerlab/blob/main/clab/clab.go#L505
I'll extend that filter to also exclude the macvtap and macvlan (fake) nodes.

hellt:
With a recent forced push I have rebased this PR on top of the current main, to include the work done in #1450

clab/dependency_manager/dependency_manager.go
hellt: either I removed this with a wrong force push, or you never added this condition?

steiler: yepp you trashed 3 commits

steiler: cherry-picked and pushed them now

PR #1407, opened 2023-05-27
fix(schema): add cvx as known kind

DerInti:
minor change to add cvx as a known kind in schema


hellt:
Hi @DerInti 

Can you please also add cumulus_cvx there, as it is an alias for cvx

hellt:
thanks @DerInti

PR #1412, opened 2023-05-30
resolve homedir of sudo user

steiler:


clab/authz_keys.go
```diff
@@ -18,24 +19,35 @@ import (
...
+		userId = "0" // default to root
+	}
+
+	// lookup user to figure out Home Directory
+	u, err := user.LookupId(userId)
+	if err != nil {
+		return err
+	}
+
+	p := utils.ResolvePath(filepath.Join(u.HomeDir, pubKeysGlobPart), "")
```
hellt, line 43: Hi @steiler 
this is a good idea, but I'd rather do this centrally in `utils.ResolvePath`
There you can use the code branch that resolves the `~` char to the user's home directory. You can just use the code block that looks up the SUDO_UID there to identify which user's home dir should it be resolved to

steiler: Alright, yes, that definetely makes sense. Thanks!

clab/authz_keys.go
```diff
@@ -18,24 +18,24 @@ import (
 )
 
 const (
-	pubKeysGlob = "~/.ssh/*.pub"
+	pubKeysGlobPart = "~/.ssh/*.pub"
 	// authorized keys file path on a clab host that is used to create the clabAuthzKeys file.
-	authzKeysFPath = "~/.ssh/authorized_keys"
+	authzKeysFPathPart = "~/.ssh/authorized_keys"
```
hellt, line 23: These are not `...Part` anymore I presume

steiler: sure yes

utils/file.go
```diff
@@ -169,10 +170,24 @@ func ReadFileContent(file string) ([]byte, error) {
...
+		return curUserHomeDir
+	}
+
+	// lookup user to figure out Home Directory
+	u, err := user.LookupId(userId)
+	if err != nil {
+		return curUserHomeDir
+	}
+
+	p = strings.Replace(p, "~", u.HomeDir, 1)
```
hellt, line 190: In 5c6fe6b I used the current user home dir as a backup dir for cases when sudo user is failed to be retrieved.
If you OK this change we can merge

steiler: Yes sure, totally fine with that!

PR #1414, opened 2023-05-31
Download License + scp and ftp (also e.g. for startup-config)

### LINKED ISSUE #1377: get license file from a remote URL
### Discussed in https://github.com/srl-labs/containerlab/discussions/1376

<div type='discussions-op-text'>

<sup>Originally posted by **jcpvdm** May 12, 2023</sup>
For example:

```
topology:
  kinds:
    vr-sros:
      image: registry.srlinux.dev/pub/vr-sros:23.3R2
      type: sr-1
      license: ftp://user:pass@1.2.3.4/licenses.txt
```</div>

### PR DESCRIPTION
**Pull request by steiler**
This PR makes license files downloadable and embeddable.
It further also implements scp and ftp as download protocols for also als the existing downloadable content (e.g. startup-config).

### CODE REVIEW

File: clab/config_test.go [hunk 1]
```diff
@@ -11,7 +11,7 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/containers/podman/v4/pkg/util"
+	"github.com/containers/libpod/pkg/util"
```
hellt, line 14: this looks like an error import

steiler:
implementing #1377

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.20
 require (
 	github.com/a8m/envsubst v1.4.2
 	github.com/awalterschulze/gographviz v2.0.3+incompatible
+	github.com/bramvdbogaerde/go-scp v1.2.1
```
hellt, line 8: I propose using sftp rather than scp
example: https://github.com/openconfig/gnmic/blob/ee2c9de75216ce19198ca5bb8f61dcc91e5da88e/utils/file.go#L128

mabra94:
Is there any timeline when we could expect this PR to be merged? Is there anything needed to be tested/checked to get this? For Nokia internal usage this would further simplify labbing!

steiler:
Let me base it on the actual master branch to resolve merge conflicts then @hellt can probably get to it. ;-)

steiler:
Do we want to run the tests for the downloads as seperate suite or also make it part of smoke tests?
I feel like it is wuite some heavily lifting for whats the checked in the end.

go.mod
FloSch62: I tried the current PR from @steiler rebased with main. So far scp works, but with scp I get the warnings: 

WARN[0000] error performing host key validation based on "/root/.ssh/known_hosts" for hostname "flosch1:22" (knownhosts: key is unknown). continuing anyways 

Also not sure if we should have this in the config file and not only as env var: CLAB_SSH_KEY

I also tried sftp, but I do not get it to work. I remains in:
DEBU[0000] Fetching "ftp://user:pass@localhost:2222/license" for node "leaf1" storing at "/tmp/.clab/clos02-leaf1-license"

I can debug a bit more...

utils/file.go
```diff
@@ -176,6 +208,159 @@ func CopyFileContents(src, dst string, mode os.FileMode) (err error) {
...
+	if u.User == nil {
+		return nil, fmt.Errorf("no username provided for scp connection")
+	}
+
+	knownHostsPath := ResolvePath("~/.ssh/known_hosts", "")
+
+	clientConfig := ssh.ClientConfig{
+		User:            u.User.Username(),
+		Auth:            []ssh.AuthMethod{},
+		HostKeyCallback: getCustomHostKeyCallback(knownHostsPath),
```
asadarafat, line 228: suggest to use ssh.InsecureIgnoreHostKey() to allow any host.

go.mod
asadarafat: > I tried the current PR from @steiler rebased with main. So far scp works, but with scp I get the warnings:
> 
> WARN[0000] error performing host key validation based on "/root/.ssh/known_hosts" for hostname "flosch1:22" (knownhosts: key is unknown). continuing anyways
> 
> Also not sure if we should have this in the config file and not only as env var: CLAB_SSH_KEY
> 
> I also tried sftp, but I do not get it to work. I remains in: DEBU[0000] Fetching "ftp://user:pass@localhost:2222/license" for node "leaf1" storing at "/tmp/.clab/clos02-leaf1-license"
> 
> I can debug a bit more...

could be `HostKeyCallback` setting this code 

```
clientConfig := ssh.ClientConfig{
		User:            u.User.Username(),
		Auth:            []ssh.AuthMethod{},
		HostKeyCallback: getCustomHostKeyCallback(knownHostsPath),
```

 `HostKeyCallback: getCustomHostKeyCallback(knownHostsPath)`, suggest to use `ssh.InsecureIgnoreHostKey()` to allow any host.

asadarafat: also, would suggest to keep SCP as a fallback in SFTP service does not available in the router.

steiler:
I've adjusted the hostkey checking... thats why it throws the warning plus the continuing anyways.
In my opinion it should use the users (user who called sudo) known hosts not the root keys.
Did you execute via sudo or did you run as root?

go.mod
steiler: this is a custom Hostkey callback that is being used.
What happens is, that it informs you that it acted like `InsecureIgnoreHostKey()` in case no known key was found.
Thats the reason for the warning plus the "continuing anyways".

FloSch62:
@steiler @hellt 
I made this PR up2date with main and implemented the missing SFTP support. Up to you guys what to do with it.

hellt:
thanks gents, lets have it in

Issue #1389, opened 2023-05-16
Implement a check for kernel version if `srl` nodes are present


hellt:
SR Linux 23.3+ container exists if the kernel version is <4.10 (centos7 under the cull).

We have to check the kernel version and stop deployment if the kernel is older than 4.10

steiler:
Taking a look at this I encounter the issue of figuring out which version is about to be deployed.
You cannot rely on tags. if `latest` is in use, that will not provide any clue. Executing the node does not make much sense either.
One thing I could think of is adding a label to the image itself at build time, that we could then rely on.

hellt:
For every Srl node (regardless of it's version) you can emit a warning instead of stopping a lab saying that "Starting from release 23.3.1 Nokia SR Linux requires kernel newer than 4.10 and kernel x.xx has been detected"

PR #1415, opened 2023-05-31
Host kernel version check for SRL

steiler:
Check the host kernel version and compare it to the SRL required minimum version.
If the minimum requirement is not met, we log a warning.


steiler:
implementing #1389

Issue #1422, opened 2023-06-06
vr-sros sendpartialconfig


steiler:
having partial startup-config which is empty leads to a panic.
We should therefor catch that and avoid the call.

Or maybe allow that to happen in scrapli fix that in scrapli. 
```
panic: runtime error: slice bounds out of range [:-1]

goroutine 109 [running]:
github.com/scrapli/scrapligo/driver/generic.(*Driver).SendCommands(0xc0006a6ff0, {0x0?, 0x0, 0x0}, {0x0, 0x0, 0x0})
        github.com/scrapli/scrapligo@v1.1.8/driver/generic/sendcommands.go:22 +0x388
github.com/scrapli/scrapligo/driver/network.(*Driver).SendConfigs(0xc000887b60, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0})
        github.com/scrapli/scrapligo@v1.1.8/driver/network/sendconfigs.go:37 +0xb8
github.com/scrapli/scrapligo/driver/network.(*Driver).SendConfigsFromFile(0xc000887b60?, {0xc0007e85f0?, 0xc000e1ede0?}, {0x0, 0x0, 0x0})
        github.com/scrapli/scrapligo@v1.1.8/driver/network/sendconfigs.go:51 +0x71
github.com/srl-labs/containerlab/nodes/vr_sros.(*vrSROS).applyPartialConfig(0xc000134050, {0x335c018, 0xc0011f73e0}, {0xc000e1ede0, 0xb}, {0x2d7fef5, 0xa}, {0x2d74a8a, 0x5}, {0x2d74a8a, ...}, ...)
        github.com/srl-labs/containerlab/nodes/vr_sros/vr-sros.go:243 +0x725
github.com/srl-labs/containerlab/nodes/vr_sros.(*vrSROS).PostDeploy(0xc000134050, {0x335bfa8, 0xc0000c1220}, 0x0?)
        github.com/srl-labs/containerlab/nodes/vr_sros/vr-sros.go:108 +0x1db
github.com/srl-labs/containerlab/cmd.deployFn.func2({0x3373fb0, 0xc000134050}, 0xc000738000?)
        github.com/srl-labs/containerlab/cmd/deploy.go:254 +0xdc
created by github.com/srl-labs/containerlab/cmd.deployFn
        github.com/srl-labs/containerlab/cmd/deploy.go:251 +0x189c
```

steiler:
[...]

PR #1423, opened 2023-06-07
allow for zero length partial config

steiler:
If in the course of building a lab, you want to already point to a partial config, which is being filled later on, probably with the use of the lab itself (configuring single node via cli and adapting to other nodes ...) This will prevent the call to scrapligo and return a nil error instead of right now resulting in a panic.
Which in both cases makes the deployment fail.


steiler:
handling #1422

Issue #1421, opened 2023-06-05
Implement healthcheck conditions used in compose


hellt:
We have more and more sophisticated labs which introduce more and more external systems.
Quite often the external systems must depend one on another, for example, a database client needs the database to be healthy first.

Docker compose has a notion of health checks which might be defined in a compose file:

```yaml
  postgres:
    image: postgres:15.3
    healthcheck:
      test: [ "CMD-SHELL", "pg_isready -d irrd" ]
      interval: 15s
      timeout: 15s
      retries: 5
```

These kind of health-checks would be super useful to have in containerlab, and I suspect they might be integrated without substantial effort if we use the same syntax and call the relevant APIs.

@steiler What do you think?

steiler:
Healthchecks might also be already be baked into the image, for general purpose healt checks.
https://docs.docker.com/engine/reference/builder/#healthcheck
It might be incarnation specific checks that we could support I assume.

hellt:
Implemented in #1426, with the goal to add them to dep manager as per #1700

PR #1426, opened 2023-06-07
support healthchecks configuration for nodes

steiler:
implementing #1421 for podman and docker. containerd does not support healthchecks

- Docs are missing (@hellt can you cover that?!)
- schema should be double checked.


hellt:
thanks @steiler 
docs and schema fixed/added.

PR #1433, opened 2023-06-18
Add Mermaid diagram support for "graph" command

YutaroHayakawa:
The description on the second commit describes the motivation and use cases for this. Putting it here for convenience.

```
Add a new --mermaid flag to the graph command and generate a Mermaid
graph when specified. The main usecase is embedding the generated text
to the Markdown. The Markdown renderer that supports Mermaid diagram can
take care of rendering the graph. That's why we only generate the text
and don't generate images like --dot flag.

I intentionally reduce the configuration knob for specifying the style.
The only knob we provide is the one that specifies the direction. This
is because there are an infinite amount of style requirements and
containerlab is not a tool for designing diagrams. Direction is helpful
for making a "baseline" structure. For example, when we write a DC network
topology, I know some people write server on top, but some people write
server at the bottom. This kind of style preference is common for network
domains so I think it's worth exposing direction as a flag to adjust those.

Why do we need to support Mermaid while we have support for Dot format?
That's because there are some use cases that we can't achieve with Dot,
but can achieve with Mermaid. For example, embedding a topology diagram
inside GitHub Markdown. Now GitHub natively supports rendering The
mermaid diagram on Markdown doesn't support Dot.

Another usecase is importing generated diagrams to draw.io. Sometimes a
bare Dot or Mermaid renderer doesn't draw the diagram that we want. In
such a case, we'd like to tweak the diagram a bit on the rich diagram
editor like draw.io, but without writing the entire diagram from
scratch. For such a case, importing a diagram into draw.io and editing
it helps. However, they only support the Mermaid format. No Dot support
at this moment.
```


docs/cmd/graph.md
```diff
@@ -106,6 +111,10 @@ With this flag, it is possible to link to local files (JS, CSS, fonts, etc.) fro
 
 With `--dot` flag provided containerlab will generate the `dot` file instead of serving the topology with embedded HTTP server.
 
+#### mermaid
+
+With `--mermaid` flag provided containerlab will generate the `mermaid` file instead of serving the topology with embedded HTTP server.
+
```
hellt, line 117: Hi @YutaroHayakawa 
Thank you for this sweet contribution, I will review it more thoroughly in the coming days.
One thing I noticed is the new flag `--mermaid-direction`, would it be possible to add a doc entry for this flag explaining which values can it receive?

YutaroHayakawa: Yep, thanks. Added in the latest push.

hellt:
Thanks @YutaroHayakawa 
Good addition, merging

Issue #1445, opened 2023-06-27
Override binds at the node level


perimore:
I understand it is possible to merge binds but can we also override matching binds? E.g. at the node level I have:

```
topology:
  kinds:
    ceos:
      binds:
        - templates/EosIntfMapping.json:/mnt/flash/EosIntfMapping.json:ro
```

But I wish to override on a specific node using:

```
    DC1-LEAF1A:
      kind: ceos
      mgmt-ipv4: 172.100.100.111
      binds:
        - templates/DC1-LEAF1A:/mnt/flash/ceos-config:ro
        - templates/48Q8.json:/mnt/flash/EosIntfMapping.json:ro
```

Therefore the mount point of `/mnt/flash/EosIntfMapping.json:ro` conflicts and we get:

`ERRO[0000] failed deploy phase for node "DC1-LEAF1A": Error response from daemon: Duplicate mount point: /mnt/flash/EosIntfMapping.json `

Is there any way to allow this or do we need to specify the same mount point on every node?

steiler:
looking into it already...

hellt:
@perimore With the enhancements provided by @steiler I think we achieved what has been requested.
You can download the beta build by running

```
docker run --rm -v $(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 pull ghcr.io/srl-labs/clab-oci:c8c5b843
```

this will download a containerlab binary in your pwd, so you have to use it as `./containerlab` to override the installed binary.
Let me know if you have a chance to test it, otherwise it will be part of the next official release.

PR #1446, opened 2023-06-27
Allow more specific bind mounts overwrite 

steiler:
Allow bind mounts to be overwritten destination wise on more specific levels

Default -> Kind -> Node

addressing #1445


hellt:
Cool stuff, @steiler 
I made a few fixes to make it work, but the implementation idea is good. Thanks for working on that.
Let me know if you lgtm the changes I made

Issue #1444, opened 2023-06-27
Cannot partially start or destroy a topology if one of the nodes is not running


perimore:
It's very frustrating that if say a node fails to start, any subsequent deploy of the topology will complain that other containers are already running:

```
INFO[0000] Containerlab v0.42.0 started                 
INFO[0000] Parsing & checking topology file: topology.yaml 
Error: containers [<list of containers>] already exist. Add '--reconfigure' flag to the deploy command to first remove the containers and then deploy the lab
```

The instruction is to use reconfigure. But this fails as the containers already exist:

```
INFO[0000] Removing <dir> directory... 
Error: containers [<list of containers>] already exist. Add '--reconfigure' flag to the deploy command to first remove the containers and then deploy the lab
```

Then simply trying to destroy the remaining nodes gives an error as some of the nodes in the topology are not running:

```
$ sudo containerlab destroy -a
INFO[0000] Parsing & checking topology file: topology.yaml 
ERRO[0000] Error occurred during the demo lab deletion: could not get container for node <failed node>: Node: <failed node>. containers not found 
Error: error(s) occurred during the deletion. Check log messages
```

The only way around this seems to be then to destroy each node one by one which is a PITA if there are lots of nodes.

hellt:
Unfortunately, we failed to reproduce this issue. It seems it appears only in specific scenarios with Docker API misbehaving and not returning all stopped containers.
Because it never appears in our setups (which we have many) it is impossible for us to troubleshoot this without users' being able to interrogate their setup.

To remove all containers which belong to a certain lab in one go you can do the following:

```bash
docker ps --filter "name=clab-srl" -q | xargs -n1 docker rm --force
```

This command removes every container whose name starts with `clab-srl`, which means any container that belongs to a lab named `srl`

/cc @steiler

steiler:
Found the issue.
further details in #1413

perimore:
Yes, I was sure this worked in an earlier version

PR #1447, opened 2023-06-27
Fix reconfigure on removed nodes

steiler:


hellt:
@steiler when you remove the container it is not longer present at all. The #1444 has a different scenario, a container is present, but in an existed state.
Can this be reproduced?

steiler:
no in that way I still can't reproduce.

kellerza:
What I do see that is that when you build a new lab and add a new node to the topo file, `--reconfigure` does not work

```bash
❯ clab deploy --reconfigure                               
INFO[0000] Containerlab v0.41.1 started
INFO[0000] Parsing & checking topology file: epe.clab.yaml
INFO[0000] Removing /home/kellerza/ip-lab/clab/sr-os/bgp-epe/clab-epe directory...
Error: containers ["clab-epe-ni1" "clab-epe-ni2" "clab-epe-oti1" "clab-epe-oti2" "clab-epe-otibgp"] already exist. Add '--reconfigure' flag to the deploy command to first remove the containers and then deploy the lab
```

steiler:
This became part of #1475

Issue #1443, opened 2023-06-26
Error: failed reading the file /root: read /root: is a directory


angely-dev:
Hello,

So I'm using containerlab for the first time on Ubuntu (22.04.2 LTS) and I'm encountering this issue:

```yml
name: simple
topology:
  nodes:
    srl:
      kind: nokia_srlinux
      image: ghcr.io/nokia/srlinux
```

*I first tried with a config file and it failed, so I tried with no config file at all and got the same error.*

```bash
$ docker images | grep srlinux
ghcr.io/nokia/srlinux      latest    07c39b3d625e   3 weeks ago    2.45GB
```

Trace:

```bash
$ sudo containerlab --log-level trace deploy
DEBU[0000] trying to find topology files automatically  
DEBU[0000] topology file found: simple.clab.yml         
INFO[0000] Containerlab v0.42.0 started                 
DEBU[0000] template variables: <nil>                    
DEBU[0000] topology:
name: simple
topology:
  nodes:
    srl:
      kind: nokia_srlinux
      image: ghcr.io/nokia/srlinux
 
DEBU[0000] method initMgmtNetwork was called mgmt params &{Network: Bridge: IPv4Subnet: IPv4Gw: IPv4Range: IPv6Subnet: IPv6Gw: IPv6Range: MTU: ExternalAccess:<nil>} 
DEBU[0000] New mgmt params are &{Network:clab Bridge: IPv4Subnet:172.20.20.0/24 IPv4Gw: IPv4Range: IPv6Subnet:2001:172:20:20::/64 IPv6Gw: IPv6Range: MTU: ExternalAccess:0xc000caf30a} 
DEBU[0000] env runtime var value is                     
DEBU[0000] Running runtime.Init with params &{Timeout:2m0s GracefulShutdown:false Debug:false KeepMgmtNet:false} and &{Network:clab Bridge: IPv4Subnet:172.20.20.0/24 IPv4Gw: IPv4Range: IPv6Subnet:2001:172:20:20::/64 IPv6Gw: IPv6Range: MTU: ExternalAccess:0xc000caf30a} 
DEBU[0000] Runtime: Docker                              
DEBU[0000] detected docker network mtu value - 1500     
DEBU[0000] initialized a runtime with params &{config:{Timeout:120000000000 GracefulShutdown:false Debug:false KeepMgmtNet:false} Client:0xc000cb2b00 mgmt:0xc000746000} 
[...]

PR #1449, opened 2023-06-27
Added homedir retrieval via getent

hellt:
For cases where os/user.LookupId fails
see #1443

### DIFF
--- a/utils/file.go
+++ b/utils/file.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"os/user"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


steiler:
Does this make sense?
He said he was making use of an ActiveDirectory account. 
The os.lookup() implementation does specifically lookup stuff in `/etc/passwd`
https://cs.opensource.google/go/go/+/master:src/os/user/lookup_unix.go;l=227-234
https://cs.opensource.google/go/go/+/master:src/os/user/lookup.go;drc=185766de0ff2810ee018501addb1f58be2226856;l=10

You're explicitly instructing `getent` now to use `passwd` to resolve the useraccount.
I don't think this will resolve the issue.

steiler:
This seems to be a much more elaborate implementation.
At least the description makes me think it would be the right fit
https://pkg.go.dev/github.com/containers/storage/pkg/idtools#LookupUser

PR #1450, opened 2023-06-29
Implement new Links format

steiler:


types/topology.go
```diff
@@ -10,14 +10,16 @@ type Topology struct {
 	Defaults *NodeDefinition            `yaml:"defaults,omitempty"`
 	Kinds    map[string]*NodeDefinition `yaml:"kinds,omitempty"`
 	Nodes    map[string]*NodeDefinition `yaml:"nodes,omitempty"`
-	Links    []*LinkConfig              `yaml:"links,omitempty"`
+	RawLinks []*RawLinkType             `yaml:"links,omitempty"`
+	Links    []*LinkConfig              `yaml:"-"`
```
hellt, line 14: it seems to me that `Links` in its current form shouldn't be part of the Topology struct

What if you make it like that:

```go
    Links []*LinkDefinition             `yaml:"links,omitempty"`
```
where `LinkDefinition` is what is your RawLinkType is currently?

steiler: yes, fine.

steiler: Take a look at [974872c](https://github.com/srl-labs/containerlab/pull/1450/commits/974872c488388b6757a39e3d30acfc88a3e27384) where this is addressed.

types/link.go
```diff
@@ -0,0 +1,218 @@
...
+		case LinkTypeMgmtNet:
+			continue
+		case LinkTypeHost:
+			continue
+		case LinkTypeMacVLan, LinkTypeMacVTap:
+			return nil, fmt.Errorf("Link type %q needs to be defined in new link format", string(lt))
+		}
+	}
+	return lc, nil
+}
```
hellt, line 218: Hi @steiler 
I am close to finialize the review, and I am concerned about this piece.
I think we should allow using the brief format for macvlan links, there is nothing wrong with it. I am quite positive that users will in 90% cases use the brief link format, as the new link format is too verbose to type in.

With that in mind my view is that internally we should use the links foundation you created here, but we need to strive to keep the brief format supported. Otherwise people will complain that it is too hard to define topologies with clab.

So I wonder if you are ok with removing this function entirely and allowing us to keep brief link definition type.

types/link.go
```diff
@@ -0,0 +1,222 @@
...
+	x := struct {
+		RawVEthLink `yaml:",inline"`
+		Type        string `yaml:"type"`
+	}{
+		RawVEthLink: *rawVEth,
+		Type:        string(LinkTypeVEth),
+	}
+
+	return x, nil
+}
```
hellt, line 81: @steiler I wonder, why do we currently need a custom marshaller, since `(*CLab).Links` contains LinkDefinitions that are not different from the old/current implementation?

types/link_veth.go
```diff
@@ -0,0 +1,54 @@
+package types
+
+import "fmt"
+
+type RawVEthLink struct {
+	LinkCommonParams `yaml:",inline"`
+	Endpoints        []*EndpointRaw `yaml:"endpoints"`
+}
```
hellt, line 8: maybe use the consistent naming pattern here and do `LinkRawVEth`?
btw, why `Raw`?

steiler: Because this is just the input format, like everything is raw strings. The one of the next steps should be to implement the non-Raw links. That would be the stuff, where e.g. the simple node name string in the endpoint is resolved and the pointer to the node is store in the actual Link object.
Sticking with the raw format is just fuel for this procedural mess we have infront of us.

types/link.go
steiler: If you ask we, I do not want to get rid of LinkDefinition in the internal code. It should be the Raw input format (see comment below) and be converted to the specific type with resolved string references (e.g. string id -> node reference).

Yes I did think of allowing the old / brief format only for the veth links which I assume is at least 90% of the cases. And force the new format onto people for all the other link types.
If you think this is the wrong path... fine, we can broaden the accepted brief definition types.

types/link.go
steiler: Yes, possible.
It was now a means to restrict the brief format to existing links (switch statement) but is also meant to, in one of the the next step perform the conversion from the brief format, which is now represented by the LinkConfig to the type specific implementation. Where LinkConfig is just the raw / input version of the brief format.

hellt: removed conversion in [f2e3fd3](https://github.com/srl-labs/containerlab/pull/1450/commits/f2e3fd3416a1502f39070d2106e0fc4161bf0b63)

we will add it when it will convert to concrete link types

types/link.go
hellt: in [b8e4ecd](https://github.com/srl-labs/containerlab/pull/1450/commits/b8e4ecda23049259e91b9b435baeafe2045bd541) I removed the custom unmarshaller for LinkDefinition, and made `generate` command to use brief format by default.

This removed a few funcs as a consequence.

hellt:
@steiler I have finilized my review, thank you for doing this work, it lays a good foundation I hope.
Let me know if you want to give it another pass and then I can merge

types/link.go
steiler: it seems you removed the marshaller not the unmarshaller

hellt: Yeah, I meant to say marshaller 😅

steiler:
Should be fine

Issue #442, opened 2021-06-10
add examples on how to use `pumba netem` for link impairments


hellt:
**ADD:** the decision is to document the ways https://github.com/alexei-led/pumba can be used in conjunction with clab, instead of adding a bespoke tc integration

a quick test by one of the colleagues showed that tc is working for vrnetlab based containers, which already warrants a `tools tc` command to be considered.

```bash
root@node1:/# tc qdisc add dev eth1 root netem delay 100ms

# ping from node 2 to node 1 via node 1 eth 1:
A:admin@node2# ping cafe:affe::101
PING cafe:affe::101 56 data bytes
64 bytes from cafe:affe::101 icmp_seq=1 hlim=64 time=206ms.
64 bytes from cafe:affe::101 icmp_seq=2 hlim=64 time=102ms.
64 bytes from cafe:affe::101 icmp_seq=3 hlim=64 time=102ms.
64 bytes from cafe:affe::101 icmp_seq=4 hlim=64 time=102ms.
64 bytes from cafe:affe::101 icmp_seq=5 hlim=64 time=101ms.
---- cafe:affe::101 PING Statistics ----
5 packets transmitted, 5 packets received, 0.00% packet loss
round-trip min = 101ms, avg = 122ms, max = 206ms, stddev = 41.7ms

# deleted delay again from inside ubuntu
root@node1:/# tc qdisc delete dev eth1 root netem delay 100ms

# same ping as before:
A:admin@node2# ping cafe:affe::101
PING cafe:affe::101 56 data bytes
64 bytes from cafe:affe::101 icmp_seq=1 hlim=64 time=1.72ms.
64 bytes from cafe:affe::101 icmp_seq=2 hlim=64 time=1.74ms.
64 bytes from cafe:affe::101 icmp_seq=3 hlim=64 time=1.64ms.
ping aborted by user
---- cafe:affe::101 PING Statistics ----
3 packets transmitted, 3 packets received, 0.00% packet loss
round-trip min = 1.64ms, avg = 1.70ms, max = 1.74ms, stddev = 0.000ms

```

We need to check if srlinux/ceos containers can leverage the same `tc` impairments

sk2:
This should also work outside the container too, which would be a more generalised approach. 
I have been using Pumba https://github.com/alexei-led/pumba which sets tc directly. 

That may possibly be limited to certain link types, but would be container agnostic.

hellt:
[...]

PR #1453, opened 2023-06-30
`tools netem` command set

steiler:
* tools netem set - to set link impairments
* tools netem show - to show link impairments

delay, jitter, loss and rate are the supported impairments


hellt:
@steiler I was testing it and couldn't make it work...
When I used your latest commit [cee2894](https://github.com/srl-labs/containerlab/pull/1453/commits/cee2894e9e3a555265ba545ed9f0e965fcd8ab2c) I never saw qdisc to be applied to the interface.

So I did a few changes to TcMsg as shown here https://github.com/florianl/go-tc/blob/main/example_netem_test.go#L50 now I see netem qdisc applied to the interface, but all traffic is stopped =|

I commented out everything but delay to make it simpler. Maybe you have some ideas?

hellt:
addresses #442

PR #1460, opened 2023-07-06
fix kernel version check logic

LimeHat:
Current comparison function for kernel ver check returns wrong results for cases where major version is > than required but minor is < than required (e.g. 6.0.1 will return a false result against 4.10.2).
This is the simplest fix


hellt (approved):

hellt:
Good catch, thanks @LimeHat

PR #1475, opened 2023-07-20
New link structs from raw

steiler:
the new link struct was just the first step.
Here we're continuing the refactoring.


steiler:
@hellt you've removed the type error, which is simply a mistake. Try adding a link in the full version with all it's attributes (mtu, Mac etc) and you will encounter the error.

types/link_test.go
```diff
@@ -101,10 +101,82 @@ func TestUnmarshalRawLinksYaml(t *testing.T) {
...
+				yaml: []byte(`
+                    type:              veth
+                    endpoints:
+                      - node:          srl1
+                        interface:     e1-1
+                        mtu:           1400
+                      - node:          srl2
+                        interface:     e1-2
+                `),
+			},
```
hellt, line 172: @steiler I've added `mtu` here to check what you said in https://github.com/srl-labs/containerlab/pull/1475#issuecomment-1666545946
locally I didn't encounter an error.

steiler:
Can we call the linkbrief method any different then resolve? Because resolve on the other link types is something totally different... This is something that might leed to a total confusion.

types/link_test.go
hellt: ha, but the test is incomplete, I have to add MTU to the expected link and it is not set there. Will check what's up

hellt: should be all good
[76fb628](https://github.com/srl-labs/containerlab/pull/1475/commits/76fb628d54ee91f0c42591dad0f4cf0a80c4d46b)

initially put MTU on the wrong object.

cmd/deploy.go
```diff
@@ -111,22 +111,32 @@ func deployFn(_ *cobra.Command, _ []string) error {
...
 			&runtime.RuntimeConfig{
 				Debug:            debug,
 				Timeout:          timeout,
 				GracefulShutdown: graceful,
 			},
 		),
+		clab.WithTopoFile(topo, varsFile),
+		clab.WithNodeFilter(nodeFilter),
 		clab.WithDebug(debug),
 	}
```
hellt, line 124: @steiler do you recall why you had to move WithTopoFile and WithNodeFilter after WithRuntime?

steiler:
I would like to propose a different name for what you now called linknoderesolver.
A link node resolver to me would ba a structure that would hold references to multiple nodes, that would resolve nodes somehow.
But it simply is a node. We could call it node and put it into the links pkg or so.
But the actual name is misleading.
Thanks

cmd/deploy.go
steiler: Yes, it has to do with attaching interfaces to the `mgmt-net`. The mgmt-net -> bridge is resolved via the runtime.
The FakeMgmtNetNode is created as a singleton and thereby initialized only on the first call. So the bridge infos need to be present before any resolution takes place.
I can take a look at if it structurally makes sense to delay the population of the fakemgmtnetbridge with the bridges name.

steiler: The bigger issue even was the CheckBridgeExists validation check that ran quite early... anyways I've prepared a change for this.
Take a look at https://github.com/steiler/containerlab/pull/2 that PR is based on this branch.

clab/clab_test.go
```diff
@@ -258,184 +260,184 @@ func Test_filterClabNodes(t *testing.T) {
...
+		// 					"node1": {
+		// 						Kind: "linux",
+		// 					},
+		// 					"node2": {
+		// 						Kind: "linux",
+		// 					},
+		// 				},
+		// 			},
+		// 		},
+		// 	},
```
hellt, line 293: @steiler this test is commented because node filtering doesn't work yet?

clab/clab.go
```diff
@@ -642,12 +596,52 @@ func (c *CLab) GetNodeRuntime(contName string) (runtime.ContainerRuntime, error)
...
-func (c *CLab) VethCleanup(_ context.Context) error {
+func (c *CLab) VethCleanup(ctx context.Context) error {
+	// remove all the links
 	for _, link := range c.Links {
-		err := c.RemoveHostOrBridgeVeth(link)
+		err := link.Remove(ctx)
 		if err != nil {
-			log.Infof("Error during veth cleanup: %v", err)
+			// if error log and continue
+			log.Error(err)
```
hellt, line 605: TODO: link removal seems to be unimplemented

steiler: I was unable to find a case in which the explicit removal of a link was required.
On removal of the netns, the interfaces are also cleaned up. I did not remove the Remove() function so far.
Can you point out a case in whcih the remove would be needed explicitely?

clab/clab.go
```diff
@@ -626,6 +580,21 @@ func (c *CLab) ListNodesContainers(ctx context.Context) ([]runtime.GenericContai
...
+		cts, err := n.GetContainers(ctx)
+		if err != nil {
+			continue
+		}
+		containers = append(containers, cts...)
+	}
+
+	return containers, nil
+}
+
```
hellt, line 597: @steiler  can the error be checked here to be exactly "not found" and not just any error?

nodes/host/host.go
```diff
@@ -44,11 +45,15 @@ func (n *host) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
...
 }
-func (*host) Deploy(_ context.Context, _ *nodes.DeployParams) error { return nil }
-func (*host) GetImages(_ context.Context) map[string]string         { return map[string]string{} }
-func (*host) PullImage(_ context.Context) error                     { return nil }
-func (*host) Delete(_ context.Context) error                        { return nil }
-func (*host) WithMgmtNet(*types.MgmtNet)                            {}
+func (n *host) Deploy(_ context.Context, _ *nodes.DeployParams) error {
+	n.SetState(state.Deployed)
+	return nil
+}
```
hellt, line 51: @steiler I wonder if setting state here is needed, given that GenericLinkNode sets this state to deployed by default?

nodes/ext_container/ext_container.go
```diff
@@ -52,6 +53,8 @@ func (e *extcont) Deploy(ctx context.Context, _ *nodes.DeployParams) error {
 	// set nspath in node config
 	e.Cfg.NSPath = nspath
 
+	e.SetState(state.Deployed)
+
```
hellt, line 57: to verify if this is required for external nodes, or is it being set somewhere else already

clab/clab.go
steiler: yes, but I think we need to then implement a "not found" error and make the runtimes convert their respective not found to  a clab generic "not found" and handle that.
Can you maybe take that action point?

nodes/host/host.go
steiler: It is a little tricky, but the fakehostnode is a different struct, host and fakehostnode might exist at the same time in the nodes.Node. So from that perspective it is needed.

steiler:
@hellt alright, so to me this looks good

clab/config_test.go
```diff
@@ -471,22 +482,6 @@ func TestLabelsInit(t *testing.T) {
...
-		t.Fatal(err)
-	}
-
-	err = c.verifyRootNetnsInterfaceUniqueness()
-	if err == nil {
-		t.Fatalf("expected duplicate rootns links error")
-	}
-	t.Logf("error: %v", err)
-}
-
```
hellt, line 489: @steiler this test was removed, as the underpinning function `verifyRootNetnsInterfaceUniqueness` was removed. But the check disappeared, you can try deploying the lab test_data/topo7-dup-rootnetns.yml and it fails.

We have to fix this

steiler: Hhmmm shit... that is two bridges that live in the root net ns. That was not on my radar.
The check disappeared, because my assumption now was, that we have the maps of links per node and that we would thereby figure out is there is a dup. but two different bridges also live in the host netns, thats correct.
I'll come up with something

PR #1482, opened 2023-07-24
feat: support falling back to docker inspect data for gleaning mgmt bridge gw address

carlmontanari:
title!

similar issue reported in discord [here](https://discord.com/channels/860500297297821756/860500297297821759/1067963224223387658) a while back. 

tl;dr -- that commit we snag the bridge address so ceos (or presumably others) can set gateway, buuuuuut if you are docker in docker (like me and my little weird launcher tool so i can run this on my mac) then the bridge is not mounted in the "launcher" container so trying to get that bridges address fails. 

we already have the output of docker inspect network from earlier on in this func, so this just falls back to attempting to iterate over the ipam section to set v4/v6 addresses. if we still dont get any we just return the original error.

tested this in my dind setup (basically mount docker sock to debian container and launch clab from there but all in docker desktop on Mac) and it worked nicely whereas before it was bombing out due to the error returned from FirstLinkIPs.

lemme know if this makes sense or any changes are needed!

### DIFF
--- a/runtime/docker/docker.go
+++ b/runtime/docker/docker.go
@@ -256,7 +256,33 @@ func (d *DockerRuntime) CreateNet(ctx context.Context) (err error) {
 	// this was added to allow mgmt network gw ip to be available in a startup config templation step (ceos)
 	var v4, v6 string
 	if v4, v6, err = utils.FirstLinkIPs(bridgeName); err != nil {
-		return err
+		log.Warn(
+			"failed gleaning v4 and/or v6 addresses from bridge via netlink," +
+				" falling back to docker network inspect data",
+		)
+
+		for _, ipamEntry := range netResource.IPAM.Config {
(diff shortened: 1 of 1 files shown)


runtime/docker/docker.go
```diff
@@ -256,7 +256,33 @@ func (d *DockerRuntime) CreateNet(ctx context.Context) (err error) {
 	// this was added to allow mgmt network gw ip to be available in a startup config templation step (ceos)
 	var v4, v6 string
 	if v4, v6, err = utils.FirstLinkIPs(bridgeName); err != nil {
```
hellt, line 258: @carlmontanari was it painful to look at that err handling line? :P

carlmontanari: lol yes! im too dumb to see it all on the one line!!!!

PR #1495, opened 2023-08-04
Log `stdout` when exec errors

hellt:
When docker exec fails to find a command to run it seems it puts it in the stdout, rather than stderr.

This new log print prints stderr and stdout when exec errors or contains non 0 rc.

### DIFF
--- a/nodes/srl/srl.go
+++ b/nodes/srl/srl.go
@@ -105,8 +105,8 @@ var (
 	//go:embed topology/*
 	topologies embed.FS
 
-	saveCmd          = `sr_cli -d "tools system configuration save"`
-	mgmtServerRdyCmd = `sr_cli -d "info from state system app-management application mgmt_server state | grep running"`
+	saveCmd          = `/opt/srlinux/bin/sr_cli -d "tools system configuration save"`
+	mgmtServerRdyCmd = `/opt/srlinux/bin/sr_cli -d "info from state system app-management application mgmt_server state | grep running"`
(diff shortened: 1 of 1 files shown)


clab/exec/exec.go
```diff
@@ -101,7 +101,18 @@ func (e *ExecCmd) GetCmdString() string {
...
+	s.WriteString(fmt.Sprintf("Cmd: %s\nReturnCode: %d", e.GetCmdString(), e.ReturnCode))
+
+	if e.Stdout != "" {
+		s.WriteString(fmt.Sprintf("\nStdout: %s", e.Stdout))
+	}
+	if e.Stderr != "" {
+		s.WriteString(fmt.Sprintf("\nStderr: %s", e.Stderr))
+	}
+
+	return s.String()
```
hellt, line 115: @steiler I think I am going mad lol
For some reason the logger eats up the first character when printing it the `Stdout:` here.
I am curious if you can crack this mystery:

```
DEBU[0002] mgmt_server status check failed, output: 
Cmd: sr_cli -d info from state system app-management application mgmt_server state | grep running
ReturnCode: 126
 tdout: OCI runtime exec failed: exec failed: unable to start container process: exec: "sr_cli": executable file not found in $PATH: unknown
```

Note, the `tdout` string.

hellt: for this log message to appear, change the command here from sr_cli to some garbage
https://github.com/srl-labs/containerlab/blob/ada8a8d26a07bd12ec1542610c8e37b9aa9ae149/nodes/srl/srl.go#L109

PR #1506, opened 2023-08-07
Adjust the robotframework tests to all use CLAB_BIN

steiler:
There was only one library that utilized the `CLAB_BIN` variable via the `tests/rf-run.sh`. This PR makes all the other tests also use the `CLAB_BIN` variable.


tests/rf-run.sh
```diff
@@ -9,10 +9,10 @@
...
-  export CLAB_BIN="containerlab"
+if [ -z "${CLAB_BIN}" ]; then
+  CLAB_BIN=containerlab  
 fi
 
 echo "Running tests with containerlab binary at $(which ${CLAB_BIN}) path and selected runtime: $1"
 
-robot --consolecolors on -r none --variable runtime:$1 -l ./tests/out/$(basename $2)-$1-log --output ./tests/out/$(basename $2)-$1-out.xml $2
\ No newline at end of file
+robot --consolecolors on -r none --variable CLAB_BIN:${CLAB_BIN} --variable runtime:$1 -l ./tests/out/$(basename $2)-$1-log --output ./tests/out/$(basename $2)-$1-out.xml $2
```
hellt: @steiler why did you decide to use this as a variable and define it via common.robot file and not from the ENV VAR as it used to be?

PR #1507, opened 2023-08-07
Do not use endpoints to deploy links

hellt:
@steiler When reviewing #1475 I noticed the indirection you used to deploy links
you called (endpoint).Deploy, which called associated (link).Deploy.

I find this indirection unnecessary, and propose to deploy links directly. For that I store []Link under the node implementation so that we can call (link).Deploy without touching endpoints.

Check this PR, if you're ok with it, I will push it to the existing PR.

### DIFF
--- a/links/link_veth.go
+++ b/links/link_veth.go
@@ -92,6 +92,8 @@ type LinkVEth struct {
 	m sync.Mutex `yaml:"-"`
 	LinkCommonParams
 	Endpoints []Endpoint
+
+	deploymentState LinkDeploymentState
 }
 
 func (*LinkVEth) GetType() LinkType {
@@ -106,10 +108,10 @@ func (l *LinkVEth) Deploy(ctx context.Context) error {
 	l.m.Lock()
 	defer l.m.Unlock()
 
-	for _, ep := range l.Endpoints {
-		if ep.GetState() < EndpointDeployStateReady {
-			return nil
-		}
+	// since each node calls deploy on its links, we need to make sure that we only deploy
+	// the link once, even if multiple nodes call deploy on the same link.
+	if l.deploymentState == LinkDeploymentStateReady {
+		return nil
 	}
 
 	// build the netlink.Veth struct for the link provisioning
@@ -159,6 +161,9 @@ func (l *LinkVEth) Deploy(ctx context.Context) error {
 			return err
 		}
 	}
+
+	l.deploymentState = LinkDeploymentStateReady
+
 	return nil
 }
 
--- a/nodes/default_node.go
+++ b/nodes/default_node.go
@@ -38,6 +38,8 @@ type DefaultNode struct {
 	// OverwriteNode stores the interface used to overwrite methods defined
 	// for DefaultNode, so that particular nodes can provide custom implementations.
 	OverwriteNode NodeOverwrites
+	// List of links that reference the node.
+	Links []links.Link
 	// List of link endpoints that are connected to the node.
 	Endpoints []links.Endpoint
 }
@@ -439,9 +441,12 @@ func (d *DefaultNode) ExecFunction(f func(ns.NetNS) error) error {
 	return netns.Do(f)
 }
 
-func (d *DefaultNode) AddEndpoint(e links.Endpoint) error {
+func (d *DefaultNode) AddLink(l links.Link) {
+	d.Links = append(d.Links, l)
+}
+
+func (d *DefaultNode) AddEndpoint(e links.Endpoint) {
 	d.Endpoints = append(d.Endpoints, e)
-	return nil
 }
 
 func (d *DefaultNode) GetEndpoints() []links.Endpoint {
@@ -458,13 +463,14 @@ func (d *DefaultNode) GetShortName() string {
 	return d.Cfg.ShortName
 }
 
-// DeployLinks deploys links for a node by calling Deploy on each endpoint of a node.
+// DeployLinks deploys links associated with the node.
 func (d *DefaultNode) DeployLinks(ctx context.Context) error {
-	for _, ep := range d.Endpoints {
-		err := ep.Deploy(ctx)
+	for _, l := range d.Links {
+		err := l.Deploy(ctx)
 		if err != nil {
 			return err
 		}
 	}
+
 	return nil
 }
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -641,6 +641,12 @@ func (c *CLab) ResolveLinks() error {
 
 		c.Endpoints = append(c.Endpoints, l.GetEndpoints()...)
 		c.Links[i] = l
+
+		// add the link to the nodes connected with it
+		for _, ep := range l.GetEndpoints() {
+			n := c.Nodes[ep.GetNode().GetShortName()]
+			n.AddLink(l)
+		}
 	}
 
 	return nil
--- a/links/endpoint.go
+++ b/links/endpoint.go
@@ -1,7 +1,6 @@
 package links
 
 import (
-	"context"
 	"fmt"
 	"net"
 
@@ -16,7 +15,6 @@ type Endpoint interface {
 	GetIfaceName() string
 	GetRandIfaceName() string
 	GetMac() net.HardwareAddr
-	Deploy(ctx context.Context) error
 	String() string
 	// GetLink retrieves the link that the endpoint is assigned to
 	GetLink() Link
@@ -25,7 +23,7 @@ type Endpoint interface {
 	// HasSameNodeAndInterface returns true if an endpoint that implements this interface
 	// has the same node and interface name as the given endpoint.
 	HasSameNodeAndInterface(ept Endpoint) bool
-	GetState() EndpointDeployState
+	// GetState() EndpointDeployState
 }
 
 // EndpointGeneric is the generic endpoint struct that is used by all endpoint types.
@@ -36,7 +34,7 @@ type EndpointGeneric struct {
 	Link     Link
 	MAC      net.HardwareAddr
 	randName string
-	state    EndpointDeployState
+	// state    EndpointDeployState
 }
 
 func (e *EndpointGeneric) GetRandIfaceName() string {
@@ -51,9 +49,9 @@ func (e *EndpointGeneric) GetIfaceName() string {
 	return e.IfaceName
 }
 
-func (e *EndpointGeneric) GetState() EndpointDeployState {
-	return e.state
-}
+// func (e *EndpointGeneric) GetState() EndpointDeployState {
+// 	return e.state
+// }
 
 func (e *EndpointGeneric) GetMac() net.HardwareAddr {
 	return e.MAC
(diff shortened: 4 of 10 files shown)


hellt:
this stuff started to break =(
Couldn't figure why, but there seems to be a data race as well

```
==================
WARNING: DATA RACE
Read at 0x00c000ca0cf8 by goroutine 51:
  github.com/srl-labs/containerlab/nodes.(*DefaultNode).AddNetlinkLinkToContainer()
      /root/srl-labs/containerlab/nodes/default_node.go:407 +0xa4
  github.com/srl-labs/containerlab/nodes/linux.(*linux).AddNetlinkLinkToContainer()
      <autogenerated>:1 +0xe8
  github.com/srl-labs/containerlab/links.(*LinkVEth).Deploy()
      /root/srl-labs/containerlab/links/link_veth.go:159 +0xd1e
  github.com/srl-labs/containerlab/nodes.(*DefaultNode).DeployLinks()
      /root/srl-labs/containerlab/nodes/default_node.go:469 +0x13d
  github.com/srl-labs/containerlab/nodes/linux.(*linux).DeployLinks()
      <autogenerated>:1 +0x76
  github.com/srl-labs/containerlab/clab.(*CLab).scheduleNodes.func1()
      /root/srl-labs/containerlab/clab/clab.go:411 +0x10b1
  github.com/srl-labs/containerlab/clab.(*CLab).scheduleNodes.func3()
      /root/srl-labs/containerlab/clab/clab.go:437 +0xd8

Previous write at 0x00c000ca0cf8 by goroutine 50:
  github.com/srl-labs/containerlab/runtime/docker.(*DockerRuntime).postStartActions()
      /root/srl-labs/containerlab/runtime/docker/docker.go:574 +0x184
  github.com/srl-labs/containerlab/runtime/docker.(*DockerRuntime).StartContainer()
      /root/srl-labs/containerlab/runtime/docker/docker.go:567 +0x684
  github.com/srl-labs/containerlab/nodes/linux.(*linux).Deploy()
      /root/srl-labs/containerlab/nodes/linux/linux.go:56 +0x2a6
  github.com/srl-labs/containerlab/clab.(*CLab).scheduleNodes.func1()
      /root/srl-labs/containerlab/clab/clab.go:405 +0xdf0
  github.com/srl-labs/containerlab/clab.(*CLab).scheduleNodes.func3()
      /root/srl-labs/containerlab/clab/clab.go:437 +0xd8

Goroutine 51 (running) created at:
  github.com/srl-labs/containerlab/clab.(*CLab).scheduleNodes()
      /root/srl-labs/containerlab/clab/clab.go:437 +0x509
  github.com/srl-labs/containerlab/clab.(*CLab).CreateNodes()
      /root/srl-labs/containerlab/clab/clab.go:271 +0x5d0
  github.com/srl-labs/containerlab/cmd.deployFn()
      /root/srl-labs/containerlab/cmd/deploy.go:234 +0x2c24
  github.com/spf13/cobra.(*Command).execute()
      /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:940 +0x152f
  github.com/spf13/cobra.(*Command).ExecuteC()
      /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0xe24
  github.com/spf13/cobra.(*Command).Execute()
      /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992 +0x4f
  github.com/srl-labs/containerlab/cmd.Execute()
      /root/srl-labs/containerlab/cmd/root.go:46 +0x44
  main.main()
      /root/srl-labs/containerlab/main.go:10 +0x29

Goroutine 50 (running) created at:
  github.com/srl-labs/containerlab/clab.(*CLab).scheduleNodes()
      /root/srl-labs/containerlab/clab/clab.go:437 +0x509
  github.com/srl-labs/containerlab/clab.(*CLab).CreateNodes()
      /root/srl-labs/containerlab/clab/clab.go:271 +0x5d0
  github.com/srl-labs/containerlab/cmd.deployFn()
      /root/srl-labs/containerlab/cmd/deploy.go:234 +0x2c24
  github.com/spf13/cobra.(*Command).execute()
      /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:940 +0x152f
  github.com/spf13/cobra.(*Command).ExecuteC()
      /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0xe24
  github.com/spf13/cobra.(*Command).Execute()
      /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992 +0x4f
  github.com/srl-labs/containerlab/cmd.Execute()
      /root/srl-labs/containerlab/cmd/root.go:46 +0x44
  main.main()
      /root/srl-labs/containerlab/main.go:10 +0x29
==================
```

Even with the data race the simple linux lab with one veth link sometimes deploys, but sometimes fails with 


```
❯ dclab dep -c -t private/linux.clab.yml 
INFO[0000] Parsing & checking topology file: linux.clab.yml 
INFO[0000] Destroying lab: linux                        
INFO[0000] Removed container: clab-linux-l1             
INFO[0000] Removed container: clab-linux-l2             
INFO[0000] Removing containerlab host entries from /etc/hosts file 
INFO[0001] Removing /root/srl-labs/containerlab/clab-linux directory... 
INFO[0001] Creating docker network: Name="clab", IPv4Subnet="172.20.20.0/24", IPv6Subnet="2001:172:20:20::/64", MTU="1450" 
INFO[0001] Creating lab directory: /root/srl-labs/containerlab/clab-linux 
INFO[0002] Creating container: "l2"                     
INFO[0002] Creating container: "l1"                     
ERRO[0003] failed deploy links for node "l2": failed to Statfs "": no such file or directory 
ERRO[0003] failed deploy links for node "l1": file exists 
WARN[0003] Cannot parse export template /etc/containerlab/templates/export/auto.tmpl: template: auto.tmpl:42:22: executing "auto.tmpl" at <$l.A.Node.ShortName>: can't evaluate field A in type links.Link 
INFO[0003] Adding containerlab host entries to /etc/hosts file 
INFO[0003] 🎉 New containerlab version 0.43.0 is available! Release notes: https://containerlab.dev/rn/0.43/
+---+---------------+--------------+----------+-------+---------+----------------+----------------------+
| # |     Name      | Container ID |  Image   | Kind  |  State  |  IPv4 Address  |     IPv6 Address     |
+---+---------------+--------------+----------+-------+---------+----------------+----------------------+
| 1 | clab-linux-l1 | 9449dd4ecb53 | alpine:3 | linux | running | 172.20.20.3/24 | 2001:172:20:20::3/64 |
| 2 | clab-linux-l2 | 41de39914ffb | alpine:3 | linux | running | 172.20.20.2/24 | 2001:172:20:20::2/64 |
+---+---------------+--------------+----------+-------+---------+----------------+----------------------+
```

The `file exists` for l1 likely is the culprit...

steiler:
So in my view this is the wrong approach.
Looking at it from a physical perspective, a node has interfaces (Endpoints) and the wire / connection between these interface (Link) connect these nodes.
Endpoints themselfes contain properties, e.g. their state but also the mac address etc. These are all very clear attributes of the a nodes specific interface. By assigning the Link to the node, you again end up in a situation, where you have to first figure out which side of a link in meant to be associated with a certain node, to figure out the relevant attributes.
This specific association will reapper in the case of graphing, where you will also need to know which endpoint is associated to a node.
By implementing it this way, the good thing is, that on e.g. veth links, we will not need any dependency manager any more to deploy the links as soon as both ends of the link are readily deployed.
Overall my vote is clearly against this proposal.

steiler:
Further I think the part you missed is here:
https://github.com/srl-labs/containerlab/pull/1507/files#diff-d435b691711a25fa1a5f1a75919a459ef9cf34304ac6ccf0e84e3ec008fef543R111-R115

In the original PR, each node calls Deploy in the endpoint. Which first sets the EndpointDeploymentState or so to Ready and then calls deploy in the link. The link then checks if both sides of the link are already in Ready state, because it might be that the other end of the link the endpoints node is not even deployed and running yet.
Only if both endpoints are in ready state the link deployment is good to go.
In your implementation you straight away try to deploy the links on the link.deploy call and thats probably also where one of the races comes from, you'r deploying links twice, like both nodes trigger the link deployment.
Lets stick to the original approach, I strongly beleave it is the right abstraction and provides for the most flexibility when it comes to further development and extensibility.

hellt:
@steiler do we have a single case where in the original approach endpoint.Deploy doesn't directly cause its associated link.Deploy and does something valuable?

And in my world from a physical pov you don't "deploy" endpoints, you deploy links, which in veth cases have endpoints

steiler:
No, not yet.

links/link_veth.go
```diff
@@ -106,8 +108,14 @@ func (l *LinkVEth) Deploy(ctx context.Context) error {
...
+	// the link once, even if multiple nodes call deploy on the same link.
+	if l.deploymentState == LinkDeploymentStateDeployed {
+		return nil
+	}
+
+	for _, ep := range l.GetEndpoints() {
+		if ep.GetNode().GetState() != "deployed" {
 			return nil
 		}
 	}
```
hellt, line 121: @steiler this seems to match what we have discussed.

1. The link is the entity that gets deployed.
2. The link has a deployment state. If it is already deployed the link deployment function returns. This is when several nodes concurrently attempt to deploy the same link
3. The link is deployed when both nodes it needs are deployed. This is done via `GetState()` method of a Node interface. The state is encoded as string, because we have to find a way where we can store an aliased type without causing a circular dep.

Since several nodes may be assi

steiler: Can we store that enum under `types`?!
I would want to avoid using plane strings from the beginning.

hellt: I tried this at first, but types import links (because of link.LinkDefinition).
Some structs replacement would be needed

nodes/default_node.go
```diff
@@ -38,8 +38,12 @@ type DefaultNode struct {
 	// OverwriteNode stores the interface used to overwrite methods defined
 	// for DefaultNode, so that particular nodes can provide custom implementations.
 	OverwriteNode NodeOverwrites
+	// List of links that reference the node.
+	Links []links.Link
 	// List of link endpoints that are connected to the node.
 	Endpoints []links.Endpoint
+	// State of the node
+	State string
```
hellt, line 46: I think an easy and dirty way to create a state enum and not deal with circular deps much is to create a package like nodes/state and put the enum there

hellt: done in [25d99e0](https://github.com/srl-labs/containerlab/pull/1507/commits/25d99e0795ab87507baf4d7a68fb6cf8166f88f7)

links/link_veth.go
hellt: did in [25d99e0](https://github.com/srl-labs/containerlab/pull/1507/commits/25d99e0795ab87507baf4d7a68fb6cf8166f88f7)

hellt:
merged into #1475

PR #1512, opened 2023-08-10
Update lab-example clos02

ubaumann:
To be able to configure the lab, I needed to add an address family in BGP and to bring the IP layer up, I had to add the "admin-state".

I also needed to install `gnmic` and change the password in the script.

This is my first time using Nokia SR Linux, so I hope I did everything correctly. I am happy to discuss the changes and update it.

### DIFF
--- a/lab-examples/clos02/setup.sh
+++ b/lab-examples/clos02/setup.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 
 CFG_DIR=./configs
+SRL_PASSWORD=NokiaSrl1!
 
 configure_SRL() {
-  OUT=$(gnmic -a clab-clos02-$1 --timeout 30s -u admin -p NokiaSrl1! -e json_ietf --skip-verify set --update-path / --update-file $CFG_DIR/$1.yaml 2>&1)
+  OUT=$(gnmic -a clab-clos02-$1 --timeout 30s -u admin -p $SRL_PASSWORD -e json_ietf --skip-verify set --update-path / --update-file $CFG_DIR/$1.yaml 2>&1)
   echo $OUT | grep -q -e '\"operation\": \"UPDATE\"'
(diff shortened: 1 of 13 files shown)


lab-examples/clos02/setup.sh
```diff
@@ -3,7 +3,7 @@
 CFG_DIR=./configs
 
 configure_SRL() {
-  OUT=$(gnmic -a clab-clos02-$1 --timeout 30s -u admin -p admin -e json_ietf --skip-verify set --update-path / --update-file $CFG_DIR/$1.yaml 2>&1)
+  OUT=$(gnmic -a clab-clos02-$1 --timeout 30s -u admin -p $PASSWORD -e json_ietf --skip-verify set --update-path / --update-file $CFG_DIR/$1.yaml 2>&1)
```
hellt, line 6: it is fine to use the documented current password -- `NokiaSrl1!`

hellt:
Thanks @ubaumann, this is a great contribution.
The reason you had to make those changes is because we didn't pin the SR Linux container to a particular version these initial config snippets were tested against.

Would you be able to ping the version in the clab file to `23.3.3` (current latest) so that others won't have the same issues should they appear with later releases?
Thanks!

lab-examples/clos02/setup.sh
ubaumann: okay I changed it to the documented password. Was not sure if you want to have it in git or not

steiler: Maybe define the PASSWORD variable at the top.?!

ubaumann: Good idea. I hope it looks good now.

ubaumann:
@hellt @steiler Hope it looks good now. If I can change anything else, please let me know.

Issue #1513, opened 2023-08-11
ext-container complains about pre-existence


steiler:
Deploying a lab with ext-containers will fail with the error, that the containers already exist if the topology prefix is set to the empty string ("").
The issue is in the VerifyContainerUniqueness() check. It checks if a container with the nodes LongName (prefix+node.shortname) already exists. For ext-containers in general only the shortname matters when interacting with the runtime, because that is how users reference the external containers.
If now the prefix is set to the empty string, the nodes longname equal the short name and querying the runtime for this name will result in a match.

My idea now is, to introduce a kind based flag that indicates that the UniquenessCheck is to be skipped.

PR #1514, opened 2023-08-11
Ext-container UniquenessCheck fix

steiler:
fixing #1513

### DIFF
--- a/nodes/ext_container/ext_container.go
+++ b/nodes/ext_container/ext_container.go
@@ -34,6 +34,7 @@ func (s *extcont) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
 	for _, o := range opts {
 		o(s)
 	}
+	s.Cfg.SkipUniquenessCheck = true
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


types/types.go
```diff
@@ -184,6 +184,7 @@ type NodeConfig struct {
 	WaitFor              []string   `json:"wait-for,omitempty"`
 	DNS                  *DNSConfig `json:"dns,omitempty"`
 	IsRootNamespaceBased bool
+	SkipUniquenessCheck  bool
```
hellt, line 187: can we add a comment here when this might be set up to true with a ref to ext container

nodes/ext_container/ext_container.go
```diff
@@ -34,6 +34,7 @@ func (s *extcont) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
 	for _, o := range opts {
 		o(s)
 	}
+	s.Cfg.SkipUniquenessCheck = true
```
hellt, line 37: I'd add a comment here explaining why this is set to true

types/types.go
steiler: done. please check

nodes/ext_container/ext_container.go
steiler: done. please check

hellt (approved):

PR #1523, opened 2023-08-14
Migrate `tools veth` to new link structs

steiler:


hellt (commented):
Hi @steiler 
thank your for looking into this depthness of repeated links creation.

When I was reviewing this I couldn't shake off the feeling that we should leverage the existing clab.NewNode function to create nodes for us to later resolve them.

So my approach here was to
1. create fake nodes that represent the nodes between which the veth pair is to be laid out. The fake nodes are created with c.NewNode existing function, so we don't have to create anything new.

2. the fake "linux" node is created for a regular endpoint, since it is a good enough to attach a veth. For host and bridge the relevant node kinds are created.

3. Once the fake nodes are created, we can create a linkBrief link type as if it was passed from the topology in brief format. That way we rely on the same link resolving process, reusing the code paths we use for regular deployment.

links/endpoint.go
```diff
@@ -41,13 +41,14 @@ type EndpointGeneric struct {
...
 
-func NewEndpointGeneric(node Node, iface string) *EndpointGeneric {
+func NewEndpointGeneric(node Node, iface string, link Link) *EndpointGeneric {
 	return &EndpointGeneric{
 		Node:      node,
 		IfaceName: iface,
 		// random name is generated for the endpoint to avoid name collisions
 		// when it is first deployed in the root namespace
 		randName: genRandomIfName(),
+		Link:     link,
```
hellt, line 51: now that I don't quite use link directly in the endpoint, I think this can even be removed...

PR #1526, opened 2023-08-15
Include special nodes in root ns duplicate interface detection

steiler:


steiler:
This is to extend #1514 which could not cope with interfaces defined in the special LinkNodes "mgmt-net" and "host".
Their endpoints weren't considerred in the duplication check implemented in #1514.

hellt:
thanks @steiler

PR #1532, opened 2023-08-18
New link struct - Vxlan

steiler:
Implement VxLAN as a struct that implements the links.Link interface.
This is then meant to be utilized via the tools command, but will also allow for the definition of VxLAN links inbetween containerlab nodes via the topology file.


steiler:
@hellt this one is fine to get reviewed
It also contains tests and docs already

steiler:
The tools command still needs to be migrated.

hellt:
note to self:
* check that we use 14789 port by default instead of 4789 as the latter is often blocked
* check that we expose an option to change the port (both in link definition and tools cmd)

nodes/srl/srl.go
```diff
@@ -585,18 +584,7 @@ func (n *srl) addDefaultConfig(ctx context.Context) error {
...
-		// via the endpoint we acquire the link, and check if the link is of type LinkMacVlan
-		// if so cast it and get the parent Interface MTU and finally set that for the interface
-		if link, ok := e.GetLink().(*links.LinkMacVlan); ok {
-			mtu, err := link.GetParentInterfaceMtu()
-			if err != nil {
-				return err
-			}
-			iface.Mtu = mtu
-		}
+		iface.Mtu = e.GetLink().GetMtu()
```
hellt, line 587: this is what causes config to break now in the srl tests in CI

two issues here

1. when we use `srl` node kind the errors from srl postdeploy are not visible (need to understand why)
2. when changed to `nokia_srlinux` kind, the following becomes evident:
```
ERRO[0017] failed to run postdeploy task for node srl1: command execution error:At line 10: Error: Error in path: .interface{.name=="ethernet-1/1"}.mtu
    [InvalidArgument] mtu 9500 out of range [1500-9412]
 
ERRO[0017] failed to run postdeploy task for node srl2: command execution error:At line 13: Error: Error in path: .interface{.name=="ethernet-1/1"}.mtu
    [InvalidArgument] mtu 9500 out of range [1500-9412]
```

the default 9500 MTU shouldn't be used in the srl config, as the 9500 link mtu is a veth mtu only, not for config in NOSes.

hellt: in [b51def3](https://github.com/srl-labs/containerlab/pull/1532/commits/b51def3673f0a8b73e8ab2ff87eced9ecd5b5689) I changed setting the mtu value in the template endpoints struct only if it is not the default veth mtu value.

tests/08-vxlan/01-vxlan.clab.yml
```diff
@@ -0,0 +1,17 @@
...
+      startup-config: 01-vxlan-s1.config
+  links:
+    - type: vxlan
+      endpoint:
+        node: s1
+        interface: e1-1
+        mac: 02:00:00:00:00:04
+      remote: 192.168.66.1
+      vni: 100
+      udp-port: 5555
```
hellt, line 17: I think instead of this lab we can use the following:

```yaml
name: vxlan-embed

topology:
  nodes:
    l1:
      kind: linux
      image: alpine:3
      exec:
        - ip addr add dev eth1 192.168.0.1/30
      mgmt-ipv4: 172.20.20.91
    l2:
      kind: linux
      image: alpine:3
      exec:
        - ip addr add dev eth1 192.168.0.2/30
      mgmt-ipv4: 172.20.20.92

  links:
    - type: vxlan
      endpoint:
        node: l1
        interface: l1eth1
      remote: 172.20.20.92
      vni: 100
    - type: vxlan
      endpoint:
        node: l2
        interface: l2eth1
      remote: 172.20.20.91
      vni: 100
```

That way you don't have to touch the host at all (and clean it up, which is not happening currently)

hellt: looking at the code it seems this lab is not possible to deploy. Since we have to find the outgoing interface (I wonder if this is a mandatory step?) we rely on the host's routing stack. It makes sense, since in real life applications, the tunnels are established between host nodes, not from within the containers, as in the lab above.

links/link_vxlan.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+type LinkVxlan struct {
+	LinkCommonParams
+	localEndpoint  Endpoint
+	remoteEndpoint *EndpointVxlan
+	noLearning     bool
+	noL2Miss       bool
+	noL3Miss       bool
+}
```
hellt, line 281: I don't think it is good to have localEndpoint and remoteEndpoint
Vxlan link seems to be best to be described with a single endpoint which is the link itself.
We create a single endpoint and put it into the node's netns

links/link_vxlan.go
```diff
@@ -0,0 +1,368 @@
...
+	return LinkTypeVxlan
+}
+
+type LinkVxlan struct {
+	LinkCommonParams
+	localEndpoint  Endpoint
+	remoteEndpoint *EndpointVxlan
+	noLearning     bool
+	noL2Miss       bool
+	noL3Miss       bool
```
hellt, line 280: this has to be reworked to pointers or default values without negating statement

tests/08-vxlan/01-vxlan.clab.yml
hellt: It is not possible to deploy this one, since vxlan links created in a single netns can't have the same VNI.
I will rework this lab

hellt: This lab works, but with prereq.
When the vlxan interface is first created in the host netns and then moved to the container netns, it needs to initiate connection towards the remote first.
If the remote end (`l2`) starts to ping `l1` first, it won't succeed. l1 needs to start pinging first.

```yaml
name: vxlan-embed

topology:
  nodes:
    l1:
      kind: linux
      image: alpine:3
      exec:
        - ip addr add dev eth1 192.168.0.1/30
      mgmt-ipv4: 172.20.20.91
    l2:
      kind: linux
      image: alpine:3
      exec:
        - >
          ash -c '
          apk add iproute2 &&
          ip link add name vxlan0 type vxlan id 100 remote 172.20.20.91 dstport 14789 &&
          ip l set dev vxlan0 up &&
          ip addr add dev vxlan0 192.168.0.2/30'
      mgmt-ipv4: 172.20.20.92

  links:
    - type: vxlan
      endpoint:
        node: l1
        interface: eth1
      remote: 172.20.20.92
      vni: 100
```

links/link_vxlan.go
hellt: can actually be removed entirely, I don't think these fields are something users would every need to touch.

nodes/srl/srl.go
hellt: So here is the issue with SR Linux and low MTU interfaces.
SR Linux emulates BCOM datapath, and port mtu on recent hw platforms can't be less that 1500B.
When using vxlan interfaces we ultimately rely on the MTU of the outgoing (parent) interfaces, which is often configured with 1500B itself.
This leaves vxlan interface with 1450B MTU which is too low, and can't be configured on the SR Linux platform.

Because of that, using vlxan interfaces inside the SRL netns can only be done if the parent interface of the vlxan tunnel has 1550B+ MTU.

tests/08-vxlan/01-vxlan.robot
```diff
@@ -0,0 +1,67 @@
...
+Suite Teardown      Cleanup
+
+
+*** Variables ***
+${lab-name}         vxlan
+${lab-file}         01-vxlan.clab.yml
+${runtime}          docker
+${vxlan-br}         clab-vxlan-br
+${vxlan-br-ip}      172.20.25.1/24
+
```
hellt, line 16: I don't yet know why this test if failing in CI, where it runs just fine locally.
Could be the resource issue (increasing wait timer?) or maybe vxlan issue where encapsulated packets are dropped for whatever reason.

For now I don't know how to fix it, some troubleshooting is required to identify the state of the interface once the lab is booted.

links/link.go
```diff
@@ -301,26 +366,38 @@ const (
...
+			}
+		} else {
+			// else we set the desired long name as alias
+			// in future we need to set it as an alternative name,
+			// pending https://twitter.com/ntdvps/status/1709580216648024296
+			err := netlink.LinkSetAlias(l, endpt.GetIfaceName())
+			if err != nil {
+				return fmt.Errorf(
+					"failed to add alias: %v", err)
+			}
```
hellt, line 384: Here is a ~big change I made to the original proposal.
Instead of creating some hashed name for the interface I propose the following:

1. if an interface name is 14+ chars we use the random name that has already been generated by clab and use it
2. the long name we set as an alias for the interface. The alias is like a description of the interface that can container 128 chars. The alias though doesn't let you use `ip link show <alias>`.
3. I am waiting on this to merge https://github.com/vishvananda/netlink/pull/862 to use alternative name instead of the alias. The altname allows users to use it in `ip` utility.

clab/clab.go
```diff
@@ -663,38 +671,9 @@ func (c *CLab) GetNodeRuntime(contName string) (runtime.ContainerRuntime, error)
...
-		// finally remove all the collected endpoints
-		log.Debugf("removing endpoint %s", ep.String())
-		err := ep.Remove()
-		if err != nil {
-			joinedErr = errors.Join(joinedErr, err)
-		}
-	}
-
-	return joinedErr
-}
```
hellt, line 694: we need to find a solution for this removed func. While working on the vlxan-stich I see that when I remove the lab, the host link remains.
This was the reason to have this func.

steiler: The Remove() function is implemented on the vxlan link and forwarded to the local endpoint. The local endpoint is actually a generic endpoint. My thinking as of now is, that the removal is triggered but executed in the wrong namespace, that of the container not in the host namespace.
Let me verify.

tests/08-vxlan/01-vxlan.robot
hellt: I made changes to only check datapath when running locally.
In CI we only check that the interfaces with right configuration are present in the relevant namespaces.

links/link_vxlan.go
```diff
@@ -0,0 +1,358 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("can't establish netlink connection: %s", err)
+		}
+		defer conn.Close()
+		r, err := conn.RouteGet(ip)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find a route to VxLAN remote address %s", ip.String())
+		}
+		parentIf = r.Interface.Name
+	}
```
hellt, line 116: @steiler I think this code piece is now shared between 
`func (lr *LinkVxlanRaw) resolveStitchedVxlanComponent(params *ResolveParams) (*LinkVxlan, error)`

and 

`func (lr *LinkVxlanRaw) resolveVxlan(params *ResolveParams) (Link, error)`

It is a good candidate to create two common funcs in utils.

links/link_vxlan_stitched.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+
+	// the veth creation might be skipped if it already exists
+	if !skipVethCreation {
+		err = l.vethLink.Deploy(ctx)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
steiler: Yes in my opinion we do.
If you take a look at from where the deploy is called, it is

```
// DeployWithExistingVeth provisons the stitched vxlan link whilst the
// veth interface does already exist, hence it is not created as part of this
// deployment
func (l *VxlanStitched) DeployWithExistingVeth(ctx context.Context) error {
	return l.deploy(ctx, true)
}

// Deploy provisions the stitched vxlan link with all its underlaying sub-links
func (l *VxlanStitched) Deploy(ctx context.Context) error {
	return l.deploy(ctx, false)
}
```
The withExistingVeth is basically to support the tools command that we have.

links/link_vxlan.go
steiler: No not in utils but still in LinkVxlanRaw.
However there is a slight difference in the vxlan endpoint part.
Anyways, let me see how we can carve the common part out.

steiler: @hellt, take a look at 8458f0b

links/link_vxlan.go
hellt: removed them [f7965ff](https://github.com/srl-labs/containerlab/pull/1532/commits/f7965ffa21b298273d07028a3e4ded00a0d02144)

tests/08-vxlan/01-vxlan.clab.yml
michelredondo: There is something strange happening with vxlan tunnels moved to container netns.

I have built a topology that it's deployed in two different hosts like this:

```yaml title="Host1"

#host1
name: vxlan-embed
prefix: ""

mgmt:
  network: fixedips
  ipv4_subnet: 100.103.1.0/24

topology:
  kinds:
    linux:
      image: ghcr.io/hellt/network-multitool

  nodes:
    node1:
      kind: linux
      mgmt_ipv4: 100.103.1.11
      exec:
        - ip addr add dev eth1 192.168.0.1/30
  
  links:
    - type: vxlan
      endpoint:
        node: node1
        interface: eth1
        mac: 02:00:00:00:00:02
      remote: 100.103.2.22
      vni: 100
      udp-port: 4789
```

```yaml title="Host1"
#host2
name: vxlan-embed
prefix: ""

mgmt:
  network: fixedips
  ipv4_subnet: 100.103.2.0/24

topology:
  kinds:
    linux:
      image: ghcr.io/hellt/network-multitool

  nodes:
    node2:
      kind: linux
      mgmt_ipv4: 100.103.2.22
      exec:
        - ip addr add dev eth1 192.168.0.2/30
  
  links:
    - type: vxlan
      endpoint:
        node: node2
        interface: eth1
        mac: 02:00:00:00:00:01
      remote: 100.103.1.11
      vni: 100
      udp-port: 4789
```

ping from 192.168.0.1 to 192.168.0.2 and vice versa does not work.

The first thing I see is that vxlan tunnels are created with the “nolearning” flag. This disables the snooping of incoming packets that builds MAC to VTEP address fdb table.

```bash
2829: eth1@if2829: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether 02:00:00:00:00:01 brd ff:ff:ff:ff:ff:ff link-netnsid 0 promiscuity 0 minmtu 68 maxmtu 65535
    vxlan id 100 remote 100.103.6.11 dev if30 srcport 0 0 dstport 4789 nolearning ttl auto ageing 300 udpcsum noudp6zerocsumtx noudp6zerocsumrx numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535
    inet 192.168.0.1/30 scope global eth1
```

If we look at fdbs in node1 and node2:

```bash
host1:
00:00:00:00:00:00 dev eth1 dst 100.103.2.22 via ifindex 9 link-netnsid 0 self permanent

host2:
00:00:00:00:00:00 dev eth1 dst 100.103.1.11 via ifindex 30 link-netnsid 0 self permanent
```

BUM traffic will be delivered to those dst IP addresses.
These ifindexes (Nexthops) are root namespace interfaces. They are the interfaces that are used to reach destination IP tunnel. It can be the mgmt. bridge or other interface in the host.

If we do a tcpdump to check vxlan traffic, we see that it’s not coming from “mgmt_ipv4” of the node but from the IP of the bridge in the host. It looks that the udp socket of the vxlan tunnel is created at the root namespace. I understand that’s the reason you also see “udp port 4789 unreachable” in the tcpdumps of container.

If I add fdb entries, using IPs in the root namepspace:

```bash

##host1
ip netns exec node1  bridge fdb append 00:00:00:00:00:00 dev eth1 self dst 100.103.2.1 vni 100 port 4789

##host2
ip netns exec node1 bridge fdb append 00:00:00:00:00:00 dev eth1 self dst 100.103.1.1 vni 100 port 4789
```

.. traffic starts to flow immediately.

Hope it helps.

hellt: Hi @michelredondo 
We will bring back the learning flag back.
As for the interface IP - it is indeed expected that the interface belongs to the root netns, as this is where the tunnel interface was created.

michelredondo: Even with the learning flag, topologies like the one I described will not work. Both containers have the  “00:00:00:00:00:00” fdb entry pointing to the incorrect IP, so it will never be possible to learn MACs from the other tunnel end. 

The https://github.com/srl-labs/containerlab/blob/main/tests/08-vxlan/01-vxlan.clab.yml test case works because srl1 fdb entry points to the correct IP (172.20.25.22). When you originate traffic from that side, srl1 sends BUM traffic to l2 node and l2 learns how to send traffic back.
If you try to originate traffic from l2 node, BUM traffic is sent to srl1 mgmt IP (172.20.25.21), but the udp vxlan socket is not listening there. It’s listening in root namespace (bridge IP: 172.20.25.1). Learning in this case never happens.

The only way I can think to make this work is to add an additional fdb entry pointing to the IP address in the root ns.  Maybe we could add this as an optional parameter in the link definition.

PR #1534, opened 2023-08-18
fix openvswitch port assignment

### LINKED ISSUE #1533: ovs support seems to have dissapeared in 0.44.0
```
mgmt:
  bridge: mgmt
  ipv4-subnet: 10.20.30.0/24
  network: mgmt1
name: lab1
topology:
  links:
  - endpoints:
    - n1:e1-1
    - ovsbr0:ovsp0
  nodes:
    n1:
      image: alpine:latest
      kind: linux
      mgmt-ipv4: 10.20.30.105
    ovsbr0:
      kind: ovs-bridge

```

```
root@lab1:/opt# clab deploy -t test.yaml
INFO[0000] Containerlab v0.44.0 started                 
INFO[0000] Parsing & checking topology file: test.yaml  
INFO[0000] Creating docker network: Name="mgmt1", IPv4Subnet="10.20.30.0/24", IPv6Subnet="", MTU="1500" 
Error: interface ovsbr0 found. expected type "bridge", actual is "openvswitch"
```

steiler:
fix #1533

### DIFF
--- a/links/endpoint_bridge.go
+++ b/links/endpoint_bridge.go
@@ -13,7 +13,7 @@ type EndpointBridge struct {
 }
 
 func (e *EndpointBridge) Verify(p *VerifyLinkParams) error {
-	errs := []error{}
+	var errs []error
 	err := CheckEndpointUniqueness(e)
 	if err != nil {
 		errs = append(errs, err)
--- a/links/endpoint_bridge.go
+++ b/links/endpoint_bridge.go
@@ -45,8 +45,8 @@ func CheckBridgeExists(n Node) error {
 			return fmt.Errorf("bridge %q referenced in topology but does not exist", n.GetShortName())
 		case err != nil:
 			return err
-		case br.Type() != "bridge":
-			return fmt.Errorf("interface %s found. expected type \"bridge\", actual is %q", n.GetShortName(), br.Type())
+		case br.Type() != "bridge" && br.Type() != "openvswitch":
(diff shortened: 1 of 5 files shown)


steiler:
waiting for patrick to double check the build

pdumais:
when creating a lab, the ovsp port is created and added in the selected ovs bridge. That's good

if creating a lab that references a bridge that has not manually been created with ovs-vsctl, clab throws an error. But this is not a change in behaviour. ok

when destroying a lab, ovsp ports get deleted but their interface is not removed from the ovs bridge. Not a change in behaviour but it would be nice to fix that some day.
                                                                                         
it is impossible to set the bridge to protocols=OpenFlow13 before launching clab. We can set the protocol version after clab
is launched. But this complicates automation scripts. It would be nice to get that fixed. This is not a change in behaviour but it is a bug. Want me to raise a separate ticket for that?
                                                                                         
setting a default drop flow and then adding flows to patch likns together works as expected

pdumais:
using `client := ovs.New(goOvs.Sudo(), ovs.FlowFormat("OXM-OpenFlow14"), ovs.Protocols("OpenFlow14"))` 
instead of `c := goOvs.New(goOvs.Sudo(),)` 
may help.

This can be tested by setting the bridge mode: `ovs-vsctl set bridge ovsbr1 protocols=OpenFlow13` before creating the lab

Issue #1535, opened 2023-08-19
0.44.0 broke srlinux tcp stack


hellt:
see https://github.com/srl-labs/learn-srlinux/pull/87#discussion_r1298976274
can reproduce with this fixed up srlfrr example (copy paste into /etc/containerlab/lab-examples/srlfrr01/srlfrr01.clab.yml):

```yaml
name: srlfrr01

topology:
  nodes:
    srl:
      kind: srl
      image: ghcr.io/nokia/srlinux:23.3.3
      startup-config: |
        # configure loopback and data interfaces
        set / interface ethernet-1/1 admin-state enable
        set / interface ethernet-1/1 subinterface 0 admin-state enable
        set / interface ethernet-1/1 subinterface 0 ipv4 admin-state enable address 192.168.1.1/24

        set / interface lo0 subinterface 0 admin-state enable
        set / interface lo0 subinterface 0 ipv4 address 10.10.10.1/32
        set / network-instance default interface ethernet-1/1.0
        set / network-instance default interface lo0.0

        # configure BGP
        set / network-instance default protocols bgp admin-state enable
        set / network-instance default protocols bgp router-id 10.10.10.1
        set / network-instance default protocols bgp autonomous-system 65001
        set / network-instance default protocols bgp afi-safi ipv4-unicast admin-state enable
        set / network-instance default protocols bgp group ibgp export-policy export-lo
        set / network-instance default protocols bgp neighbor 192.168.1.2 admin-state enable
        set / network-instance default protocols bgp neighbor 192.168.1.2 peer-group ibgp
        set / network-instance default protocols bgp neighbor 192.168.1.2 peer-as 65001

        # create export policy
        set / routing-policy policy export-lo statement 10 match protocol local
        set / routing-policy policy export-lo statement 10 action policy-result accept
    frr:
      kind: linux
      image: frrouting/frr:v7.5.0
      binds:
        - daemons:/etc/frr/daemons
      exec:
        - |
          vtysh -c 'configure terminal
          interface eth1
          ip address 192.168.1.2/24
          !
          interface lo
            ip address 10.10.10.2/32
          !
          router bgp 65001
            bgp router-id 10.10.10.2
            neighbor 192.168.1.1 remote-as 65001
            !
            address-family ipv4 unicast
             network 10.10.10.2/32
            exit-address-family'

  links:
    - endpoints: ["srl:e1-1", "frr:eth1"]
```

[...]

PR #1536, opened 2023-08-19
Test PR with reverted #1475 to address srlinux issue

hellt:
Revert "New link structs from raw (#1475)"
This reverts commit 0b023dc882c8a562798b95d291bdecfefa8bbcda.

tests are failing, but you can build it to check that reverting this PR addresses #1535


hellt:
it was a missing tx offload
#1537

PR #1539, opened 2023-08-20
List OVS bridges to allow for OF version >1.0

pdumais:
There is a problem with the OVS library being used when retrieving a bridge. But since we just need to know if the bridge exists or not, and we dont actually need any details about it, then I changed it to list the bridges. 

This now works with bridges that are created with `ovs-vsctl set bridge s1 protocols=openflow13`

### DIFF
--- a/nodes/ovs/ovs.go
+++ b/nodes/ovs/ovs.go
@@ -56,11 +56,11 @@ func (n *ovs) CheckDeploymentConditions(_ context.Context) error {
 	)
 
 	// We were previously doing c.VSwitch.Get.Bridge() but it doesn't work
-	// when the bridge has a propotocol version higher than 1.0
+	// when the bridge has a protocol version higher than 1.0
 	// So listing the bridges is safer
 	bridges, err := c.VSwitch.ListBridges()
 	if err != nil {
-		return fmt.Errorf("error while looking for ovs bridge %q: %v", n.Cfg.ShortName, err)
+		return fmt.Errorf("error while listing ovs bridges: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


nodes/ovs/ovs.go
```diff
@@ -54,7 +55,14 @@ func (n *ovs) CheckDeploymentConditions(_ context.Context) error {
 		goOvs.Sudo(),
 	)
 
-	if _, err := c.VSwitch.Get.Bridge(n.Cfg.ShortName); err != nil {
+	// We were previously doing c.VSwitch.Get.Bridge() but it doesn't work
+	// when the bridge has a propotocol version higher than 1.0
```
hellt, line 59: @pdumais do you have a link to the upstream issue about it? I wonder why this is happening because they use ovsctl in the backend

pdumais: I did not see any issue about it. I tried specifying the version in New() but it still wasn't working. It has never worked ( I tried with a few older versions).

I did not troubleshoot their code that much. All I know is that are parsing json output but it yields a parsing error. When trying the OVS command manually it works. So I'm not sure where their library is not handling it corrently. 

It has never worked. I tried with clab 0.42 and this issue was present then. 

The nice thing about this fix is that it makes it version-agnostic. There is no need to know the bridge version when listing the bridges.

pdumais: It's actually an issue with `ovs-vsctl`.
https://www.mail-archive.com/search?l=discuss@openvswitch.org&q=subject:%22%5C%5Bovs%5C-discuss%5C%5D+%5C%5Bovs%5C-vsctl%5C%5D%22&o=newest&f=1

PR #1551, opened 2023-08-23
add suppress-startup-config

### LINKED ISSUE #1548: Add config option to suppress startup-config generation
It would be useful in cases where testing things such as ZTP (zero touch provisioning) to be able to generate a lab where one or more devices does not have a startup configuration present on flash.

Some platforms (EOS) will skip ZTP if a startup file is present at all, even if blank.

Happy to work on a PR for this later this week or weekend.

bewing:
fix #1548


types/topology.go
```diff
@@ -217,6 +217,20 @@ func (t *Topology) GetNodeEnforceStartupConfig(name string) bool {
...
 
+func (t *Topology) GetNodeSuppressStartupConfig(name string) bool {
+	if ndef, ok := t.Nodes[name]; ok {
+		if ndef.GetSuppressStartupConfig() {
+			return true
+		}
+		if t.GetKind(t.GetNodeKind(name)).GetSuppressStartupConfig() {
+			return true
+		}
+		return t.GetDefaults().GetSuppressStartupConfig()
```
steiler, line 228: The issue with this is, that you can not set SuppressStartupConfig on the default or kind level but overwrite it on the more specific level (kind or node) because if you explicitly set it to false, it will evaluate the higher level.
The field must be defined as a pointer  with yaml omitempty hence default nil and not false and the evaluation here needs to get adjusted as well.

bewing: Thanks for the feedback.  Refactored as a boolean and pushed a new changeset.  Let me know if a third test case is warranted with an explicit false config setting

steiler: Id make  GetNodeSuppressStartupConfig() return a bool not the pointer because that func is able to resolve the finit value from the hierarchy. But itself it need the info if the setting is defined on a per level basis.

bewing: That sounds like halfway between the approach of `topology.GetNodeEnforceStartupConfig` and `topology.GetNodeAutoRemove`, without really matching either of them.  Is it worth adding a third method?

Would also imply needing to test `GetNodeSuppressStartupConfig` directly in `config_test.go`, I imagine?

bewing: pushed new changeset with recommended changes and additional tests

steiler: You are right. 
In my viewe should also adjust the two functions mentioned.
- `topology.GetNodeEnforceStartupConfig`
- `topology.GetNodeAutoRemove`

to also resolve bool pointers and return actual bools, to figure out if a certain value was set at a given level or not (allow for overwrites). @hellt, whats your opinion?

nodes/default_node.go
```diff
@@ -229,10 +229,13 @@ func (d *DefaultNode) VerifyStartupConfig(topoDir string) error {
 // GenerateConfig generates configuration for the nodes
 // out of the template based on the node configuration and saves the result to dst.
 func (d *DefaultNode) GenerateConfig(dst, templ string) error {
-	// If the config file is already present in the node dir
-	// we do not regenerate the config unless EnforceStartupConfig is explicitly set to true and startup-config points to a file
-	// this will persist the changes that users make to a running config when booted from some startup config
-	if utils.FileExists(dst) && (d.Cfg.StartupConfig == "" || !d.Cfg.EnforceStartupConfig) {
+	if d.Cfg.SuppressStartupConfig {
+		log.Infof("Startup config for '%s' node suppressed", d.Cfg.ShortName)
```
hellt, line 233:
```suggestion
		log.Infof("Startup config generation for '%s' node suppressed", d.Cfg.ShortName)
```

types/topology.go
hellt: @steiler  yes, I agree, that it would be better to unify the logic for boolean node properties to use pointers as NodeDefinition fields, but return bool when *topology.GetXXX is used

steiler: alignment via #1567

hellt:
Thanks for the contribution @bewing @steiler

PR #1560, opened 2023-08-28
Simplify if else block in GenerateConfig()

steiler:


hellt:
@steiler would you mind rebasing this on top of main to get rid of the merge conflict that happened after suppress-config merge? Thanks!

steiler:
Done. Please take another look if the logic still makes sense.
I'm positive, but please double-check.

nodes/srl/srl.go
```diff
@@ -486,13 +485,12 @@ func (s *srl) createSRLFiles() error {
...
 
-		cfgTemplate := string(c)
-
-		err = s.GenerateConfig(dst, cfgTemplate)
-		if err != nil {
-			log.Errorf("node=%s, failed to generate config: %v", s.Cfg.ShortName, err)
-		}
+	err = s.GenerateConfig(cfgPath, cfgTemplate)
+	if err != nil {
+		log.Errorf("node=%s, failed to generate config: %v", s.Cfg.ShortName, err)
```
hellt, line 493: @steiler this piece was needed I think.
without it. gen config func was never called when startup config was empty, hence no log message was printed

steiler: Alright, then I think it is fine.

Issue #1562, opened 2023-08-29
Automatically look for `*.clab.yml` file when `--topo /some/dir` is provided.


hellt:
Currently we allow users to omit `-t` flag entirely in which case we look for a file with the `*clab.yml` pattern.
We have to use the same pattern when a user provides a topo path that points to a dir.
For example, if a user provides the following:

```
clab dep -t /some/dir
```

then we should search for a clab topo files in `/some/dir`

PR #1565, opened 2023-08-31
Reference topology via its dir

steiler:
implements #1562

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -124,7 +124,7 @@ func WithKeepMgmtNet() ClabOption {
 	}
 }
 
-func WithTopoRef(topoRef, varsFile string) ClabOption {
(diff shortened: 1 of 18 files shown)


clab/clab.go
```diff
@@ -123,11 +124,17 @@ func WithKeepMgmtNet() ClabOption {
 	}
 }
 
-func WithTopoFile(file, varsFile string) ClabOption {
+func WithTopoRef(topoRef, varsFile string) ClabOption {
```
hellt, line 127: How about `WithTopoPath`? A path can be all the way to the file or to the dir

steiler: done

hellt:
thanks @steiler 
great feature delivery as always

Issue #1569, opened 2023-08-31
Add `ansible_httpapi_use_proxy: false` to the generated ansible inventory


hellt:
Since clab generated inventory assumes that the HTTP client that uses it resides on the same host, we have to add `ansible_httpapi_use_proxy: false` variable to the inventory to make sure we instruct the module to not attempt using global proxy if it is configured.

A comment should be added next to this var explaining this.

steiler:
is this what you would expect? 
Like vars under all or should it be different?
```
all:
  vars:
    ansible_httpapi_use_proxy: false
  children:
    linux:
      hosts:
        clab-host-link-l1:
          ansible_host: 172.20.20.2
```

hellt:
yes, thanks

PR #1571, opened 2023-09-01
Adding ansible_httpapi_use_proxy: false to ansible inventory

steiler:
implementing #1569


hellt:
@steiler yes, looks good, only the test needs to be adjusted

PR #1603, opened 2023-09-20
Update c8000.md

svautour:
Added installation command


hellt:
Hi @svautour 
can you move this text under the `Getting Cisco 8000 ContainerLab docker images` section?
Can omit "installation" heading and just write something like:

the obtained image archive can be loaded to local docker image store with:
```
command here
```

svautour (commented):
Added docker load command
Fixed docker logs command
Added example

hellt:
Thanks!

PR #1612, opened 2023-09-27
Fix a typo in topo-def-file.md

Lachstec:
There was a typo in the word "point-to-point" (previously "point-po-point").


hellt:
thanks @Lachstec

PR #1622, opened 2023-10-07
Fetch remote topo

friday963:
- Adds functionality to pull in topology file from remote github repository.  
  - User can pass in topology file using generic https://github.com/path/to/yml/file.yml || https://raw.githubusercontent.com/path/to/yml/file.yaml.
  - file can be either .yml or .yaml.
 - Errors on any other url at this time.


friday963:
Integration testing I conducted
- containerlab deploy -t https://raw.githubusercontent.com/srl-labs/containerlab/main/lab-examples/srl01/srl01.clab.yml (PASSING)
- containerlab deploy -t https://github.com/srl-labs/containerlab/blob/main/lab-examples/srl01/srl01.clab.yml (PASSING)
- containerlab deploy -t srl01.clab.yml (PASSING)
- containerlab deploy -t https://gitlab.com/test.yml (FAILS bc gitlab and not github)

utils/http.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func DownloadFile(url string, ouputFileName string) error {
+	// Get the data
+	resp, err := http.Get(url)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return errors.New("URL does not exist")
```
steiler, line 47: There are may other http status codes, that have different meanings, other then "URL does not exist".
The returned error should account for that.

utils/http.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return url
+}
+
+func IsGitHubURL(url string) bool {
+	return strings.Contains(url, "github")
+
+}
+
+func GetFileName(url string) string {
```
steiler, line 25: This should probably use https://pkg.go.dev/path#Base.

utils/http.go
```diff
@@ -0,0 +1,64 @@
...
+func GetRawURL(url string) string {
+	if strings.Contains(url, "github.com") {
+		raw:=strings.Replace(url, "github.com", "raw.githubusercontent.com", 1)
+		return strings.Replace(raw, "/blob", "", 1)
+	}
+	return url
+}
+
+func IsGitHubURL(url string) bool {
+	return strings.Contains(url, "github")
```
steiler, line 21: This contains in my view is too broad. since also a "raw.githubusercontent.com" url would match.
In cmd/root.go you use this in the switch and consecutively apply the GetRawUrl. Which then does the conversion.
although that would check for "github.com" it should probably be more kind of "consistent".

utils/http.go
steiler: Generally I'm actually seeing that we already have the http download functions available.
Check out `CopyFileContents `https://github.com/srl-labs/containerlab/blob/main/utils/file.go#L86

utils/http.go
steiler: actually this is also available under utils/file.go -> `FilenameForURL`
https://github.com/srl-labs/containerlab/blob/main/utils/file.go#L259

steiler:
@hellt we should probably, when this is in shape merge this and #1414 and continue from there. Opinion?

hellt:
@steiler my idea was that this should solve all artifacts retrieval, both topo file as well as all linked artifacts like license, binds, startup-configs, etc.

steiler:
yes, and e.g. the general license retrieval is already implemented in #1414 so they sould be merged and finalized altogether

utils/http.go
friday963: Valid point on the status code.  As for the CopyFileConents, I tried to implement that originally but was not able to get it to work for the remote file contents.  I'll re-explore and see if I can get to work.

friday963: Updated to use copy method.

utils/http.go
friday963: updated to use FilenameForURL.

utils/http.go
friday963: I'm not sure I'm grasping what you'd like to see from this method.
You mentioned that it would match on raw.githubusercontent.com and you're right.  I had planned on it matching on either a generic github.com or the raw.github path to offer flexibility to the end user.  It then proceeds to that GetRawURL method where it only converts it if its a generic github.com url else it returns the url without converting anything.
How would you envision this changing?

PR #1624, opened 2023-10-08
fix: use "vx-<iface>" name for host interface name of vxlan tunnels

carlmontanari:
per chat w/ @hellt  -- this "fixes" a c9s thing but as im not sure about all the changes in vxlan bits lately im not very confident that this doesnt break anything else :)


carlmontanari:
closing, see #1625

PR #1625, opened 2023-10-08
Fix `tools vxlan create` cmd

hellt:
I screwed up the tools cmd with not respecting the user-provided link name.


carlmontanari:
LGTM and tested w/ c9s and looks good!

Issue #1634, opened 2023-10-09
set learning for vxlan links


hellt:
https://github.com/srl-labs/containerlab/pull/1532#discussion_r1350740640

PR #1636, opened 2023-10-10
Enable vxlan learning flag

steiler:


steiler:
Fixing #1634

hellt:
@steiler can you tune one of the vxlan tests where we do 'ip -d link show' and check in the output that learning is set?

steiler:
will do

Issue #1640, opened 2023-10-12
Deploying lab with SRL nodes fails when attaching mgmt0


pdumais:
```
# cat test1.yaml 
name: test1 
topology:
  kinds:
    srl:
      image: ghcr.io/nokia/srlinux:22.11.2
  nodes:
    POD1-LEAF01:
      kind: srl
        #type: ixrd3l
      network-mode: none
  links:
    - endpoints: ["POD1-LEAF01:mgmt0", "macvlan:eth2"]

```
```
# clab deploy -t test1.yaml 
INFO[0000] Containerlab v0.46.2 started
INFO[0000] Parsing & checking topology file: test1.yaml 
INFO[0000] Creating docker network: Name="clab", IPv4Subnet="172.20.20.0/24", IPv6Subnet="2001:172:20:20::/64", MTU="1500"
INFO[0000] Creating lab directory: /opt/dtaas/crap/clab-test1 
INFO[0000] Creating container: "POD1-LEAF01"
INFO[0000] Creating MACVLAN link: host:eth2 <--> POD1-LEAF01:mgmt0 
INFO[0000] Running postdeploy actions for Nokia SR Linux 'POD1-LEAF01' node 
panic: runtime error: index out of range [1] with length 1

goroutine 16 [running]:
github.com/srl-labs/containerlab/nodes/srl.(*srl).addDefaultConfig(0xc0000d6000, {0x3506fc8, 0xc000b9cb90})
        github.com/srl-labs/containerlab/nodes/srl/srl.go:595 +0xa47
github.com/srl-labs/containerlab/nodes/srl.(*srl).PostDeploy(0xc0000d6000, {0x3506fc8, 0xc000b9cb90}, 0xc0000bcf78)
        github.com/srl-labs/containerlab/nodes/srl/srl.go:322 +0x4e5
github.com/srl-labs/containerlab/cmd.deployFn.func1({0x3528e18, 0xc0000d6000}, 0xc000607fb8?)
        github.com/srl-labs/containerlab/cmd/deploy.go:251 +0xdf
created by github.com/srl-labs/containerlab/cmd.deployFn
        github.com/srl-labs/containerlab/cmd/deploy.go:248 +0x1865
```

srl.go:595 seems to indicate that only interfaces with names e-X-Y are supported

hellt:
yes, this is a legit issue. We haven't accounted for non datapath interfaces to be used in the link definition. 
You can comment out the offending string and continue your testing.

[...]

PR #1641, opened 2023-10-13
Process srl templating differently for mgmt0

steiler:
Fixing #1640

### DIFF
--- a/nodes/srl/srl.go
+++ b/nodes/srl/srl.go
@@ -66,6 +66,10 @@ set / system snmp network-instance mgmt
 set / system snmp network-instance mgmt admin-state enable
 set / system lldp admin-state enable
 set / system aaa authentication idle-timeout 7200
+{{- /* if e.g. node is run with none mgmt networking but a macvlan interface is attached as mgmt0, we need to adjust the mtu */}}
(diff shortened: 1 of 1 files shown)


nodes/srl/srl.go
```diff
@@ -586,6 +592,16 @@ func (n *srl) addDefaultConfig(ctx context.Context) error {
...
 	for _, e := range n.Endpoints {
 		ifName := e.GetIfaceName()
+		if ifName == "mgmt0" {
+			// skip the mgmt0 interface. This is just for traffic carrying interface
+
+			// if the endpoint has a custom MTU set, use it in the template logic
+			// otherwise we don't set the mtu as srlinux will use the default max value 9232
+			if m := e.GetLink().GetMTU(); m != links.DefaultLinkMTU {
+				tplData.MgmtMTU = m
+			}
```
hellt, line 602: I think also another condition can be copied from the generic mtu handling where we do not allow setting mtu less than 1500

hellt: not needed, the interface won't come up if the mtu is lower than 1500 anyways, so the check is not quite useful

Issue #1638, opened 2023-10-12
Auto-provisioning of DNS servers for SR Linux


hellt:
SR Linux management is a separate network namespace which doesn't share the DNS of the root netns setup by docker.

Because of that, users who want to get internet/external access from the management (or any other) net-inst have to configure DNS manually.

We can improve this a bit by

1. figuring out which DNS servers are configured on the host
2. adding these dns entries in the SRL config template

This will remove one of the tedious workflows of adding DNS servers manually, because often you can't just add 1.1.1.1 or 8.8.4.4 and you need to use the DNS servers of the cloud environment your VM is on.

## finding DNS addresses

It all starts with DNS addresses. They can be found in `/etc/resolv.conf` for older systems and in `sudo resolvectl status <interface>` on newer distros.
We have to check both and add addresses from both if both exist.

To find which interface DNS addresses to use we should use rtnetlink and finding a route to 8.8.8.8 and the interface that is returned for this addr.

## Configuring DNS entries
<img width="404" alt="image" src="https://github.com/srl-labs/containerlab/assets/5679861/a42dfbc5-4bd4-4e82-b41c-4e5f6806a885">

hellt:
DNS in srl should be configured via srlinux config

hellt:
Do you mean to use this file to read which DNS servers are configured in the system to populate it in the srl config?

steiler:
yes. We do not need to run any route lookup or anything, just take the two files, extract the ip from the `nameserver <ip>` lines and we're all good. Take a look at above PR.

PR #1650, opened 2023-10-16
Query hosts dns servers and populate nodes config with it

steiler:
Fixing #1638

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -184,6 +184,10 @@ func deployFn(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
+	if err := c.PrepareDNSServers(); err != nil {
+		return err
+	}
+
 	// determine the number of node and link worker
 	nodeWorkers, _, err := countWorkers(uint(len(c.Nodes)), uint(len(c.Links)), maxWorkers)
 	if err != nil {
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -184,10 +184,6 @@ func deployFn(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	if err := c.PrepareDNSServers(); err != nil {
-		return err
-	}
-
 	// determine the number of node and link worker
 	nodeWorkers, _, err := countWorkers(uint(len(c.Nodes)), uint(len(c.Links)), maxWorkers)
 	if err != nil {
--- a/utils/resolve.go
+++ b/utils/resolve.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"io/fs"
 	"regexp"
-	"slices"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
@@ -53,8 +52,5 @@ func ExtractDNSServerFromResolvConf(filesys fs.FS, filenames []string) ([]string
 	for k := range DNSServersMap {
 		DNSServers = append(DNSServers, k)
 	}
-
-	// sort the slice to allow for stable testing
-	slices.Sort(DNSServers)
 	return DNSServers, nil
 }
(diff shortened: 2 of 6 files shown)


steiler:
I'm not too happy about calling `PrepareDNSServers()` in `deployFn()`. @hellt opinion?

steiler:
An option might be to do that stuff as part of https://github.com/srl-labs/containerlab/blob/main/types/topology.go#L508

We maybe add the `ExtractDNSServerFromResolvConf()` to the topology, cache the result and in the GetDNS function we evaluate if DNS servers are set, if not, we populate it with the determined system DNS IPs

utils/resolve.go
```diff
@@ -0,0 +1,56 @@
...
+	"regexp"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// ExtractDNSServerFromResolvConf takes a list of filenames
+// the files are supposed to be formated in the resolv.conf format.
+// it will extract ipv4 DNS server addresses and return these
+func ExtractDNSServerFromResolvConf(filenames []string) ([]string, error) {
```
hellt, line 15: I would suggest the func to deal with an io.ReadCloser (file) instead of a list of file paths.
That way we can write a unit test easily for it and then in the caller we will range over the provided paths and catenate the extracted results.

cmd/deploy.go
```diff
@@ -184,6 +184,10 @@ func deployFn(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
+	if err := c.PrepareDNSServers(); err != nil {
```
hellt, line 187: let's rename to `ExtractDNSServers`

As for the calling this in deploy I think it can be moved to NewContainerlab? After the config is initialized.

utils/resolve.go
steiler: Went down the road of io.fs which allows for a MapFS to be handed to the tests.

cmd/deploy.go
steiler: Done

jbemmel:
I don't like this feature, as it causes labs to behave differently when deployed in different environments.

As a minimum it should be disabled by default, with an option/flag to enable it

hellt:
@jbemmel what do you mean? Every environment will have its own DNS servers set up. If you need to rewrite it the default servers you do that manually. 

As with regular docker we just let you leverage DNS servers you have in your env for srl nodes

jbemmel:
> @jbemmel what do you mean? Every environment will have its own DNS servers set up. If you need to rewrite it the default servers you do that manually.
> 
> As with regular docker we just let you leverage DNS servers you have in your env for srl nodes

My concern is that this feature adds non-reproducibility - the config you end up with in the nodes is different in a behind-the-scene context dependent way. That is a problem for certain usages of Containerlab (you'll have cases where things work in one lab, and then someone deploys the exact same lab in a different context and it doesn't work)

hellt:
This particular feature is perfectly in line with containerlab promises of delivering docker UX for networking labs. 
With docker you have dns resolution that just works. Currently it doesn't for srl because of a different netns management lives in.

jbemmel:
> This particular feature is perfectly in line with containerlab promises of delivering docker UX for networking labs. With docker you have dns resolution that just works. Currently it doesn't for srl because of a different netns management lives in.

I agree it's a nice and useful feature, I'm merely saying that imho it should be explicit (or at least have the ability to disable it if as a user I want my lab to be self-contained, free from dependencies on the environment)

hellt:
If you'd want to get rid of this you'd clear the DNS stanza with the startup config. 

I don't see a reason why would you do that though, buy in 1% cases that would be the way to turn this feature off or rewrite it with a list of DNS servers you want to have. 

I wouldn't want to introduce a flag or env var to turn this feature off until there is a legitimate use case behind it. 

We will document the DNS provisioning in the srl doc

PR #1654, opened 2023-10-17
Fetch remote topo

friday963:
Updated code to fetch a remote topology on github.  Users can point to a repo with a topology file in it or without, the hyperlink could either be a generic github link or a raw github link. 

Here is what I've tested:

https://github.com/srl-labs/srl-evpn-mh-lab.git
https://github.com/srl-labs/srl-evpn-mh-lab
https://github.com/srl-labs/srl-evpn-mh-lab/tree/truncated-configs
https://github.com/srl-labs/srl-evpn-mh-lab/blob/main/evpn-mh.clab.yml
https://raw.githubusercontent.com/srl-labs/srl-evpn-mh-lab/main/evpn-mh.clab.yml


friday963:
Feel free to hit me up on the discord channel or here when you've reviewed and determined if these changes meet the AC.

utils/http.go
```diff
@@ -0,0 +1,101 @@
...
+	// ckecks if the url has a valid suffix, if not it returns an error
+	supported_suffix := []string{".yml", ".yaml", ".git"}
+	for _, suffix := range supported_suffix {
+		if strings.HasSuffix(url, suffix) {
+			return suffix, nil
+		}
+	}
+
+	return "", ErrInvalidSuffix
+}
```
hellt, line 101: this demands some thinking
I removed in [97dd280](https://github.com/srl-labs/containerlab/pull/1654/commits/97dd28062ae735b84566f8bb022b3de9f9237764) the original test/function because the func had an empty string as a supported suffix and every string has it, so it was not correct.

Probably this has to be refactored in its entirety

hellt: removed the return of the error. The function will either return yml/yaml or git extension, without erroring out.

hellt:
I did a quick refactoring
but
* `clab dep -c -t https://github.com/srl-labs/srl-evpn-mh-lab`
* `clab dep -c -t https://github.com/srl-labs/srl-evpn-mh-lab/`

still fails

utils/http.go
friday963: So I had that empty string in there to support a github link like https://github.com/srl-labs/srl-evpn-mh-lab/tree/truncated-configs or https://github.com/srl-labs/srl-evpn-mh-lab.  Without that empty string the HasSuffix function will not recognize them as acceptable links.  So it makes sense to me that its not working below.  I did not however think about a url with a forward slash without a path after it.

friday963:
@hellt My pleasure, glad I could contribute back :)

Issue #1659, opened 2023-10-19
SSH config


steiler:
When redeploying a lab we always run into host key issues and we also need to provide the correct username.
I want to, besides populating the `/etc/hosts` file, create a file `/etc/ssh/ssh_config.d/clab-<labname>.conf`, which would contain the following content, repeated per node:

```
Host <longName>
	Hostname <IP>
	{{  if is set ->  User <credentials->User> }}
	StrictHostKeyChecking=no 
	UserKnownHostsFile=/dev/null
```

This would generally avoid the know_host key issues and in the general case not reqire us to provide the user name on ssh connections (altough can always be owerwriten via commandline user@ or -l <user>).

On topology deletion, the file `/etc/ssh/ssh_config.d/clab-<labname>.conf` would simply need to be deleted.

PR #1660, opened 2023-10-19
Generate SSH config

steiler:
Generate SSH config that:
- Set the Username for the nodes
- Set StrictHostKeyChecking=no
- Set UserKnownHostsFile=/dev/null

Been thinking of using wildcard match, but in case the prefix of nodes is set to ""
```
name: MyTopoName 
prefix: ""
```
The Names would not have a prefix name of "clab-<labname>" anymore. So going with individual entries is the best option.

Implementing #1659

### DIFF
--- a/clab/sshconfig.go
+++ b/clab/sshconfig.go
@@ -17,7 +17,7 @@ type tmplSshConfigData struct {
 // thats handed to the template engine
 type tmplSshConfigDataNode struct {
 	Name     string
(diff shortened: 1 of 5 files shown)


hellt:
@steiler thanks for this, good quality of life improvement IMO
if you're happy with my polishing stuff I am good with merging this

steiler:
lgtm

PR #1664, opened 2023-10-23
Subdirectories in cloned topologies and gitlab support

steiler:
This PR aims at evolving the "fetching remote repositorys" functionality.

- It is now possible to reference also files that do not reside in the root of a repository.
- There is now a notion of different git webservices
- Gitlab is kind of not that straight forward, since it can also be a selfhosted instance. Right now I have a strings.Contains(url,"gitlab) in place. This might lead to issues, on the other hand if there is no gitlab in the hostname you can simply add a query parameter <url>?gitlab  or if parameters already exist in the URL <url-with-params>&gitlab to trigger using the gitlab implementation.
- The actual git action, is seperated from the parsing of the git webservice URLs
- Two implementations of the git clone exist. ExecGit and GoGit. Whilst ExecGit is the os.exec() variant that uses the git binary and GoGit uses the go-git golang library.
- The GoGit implementation on consecutive execution inspects the already downloaded repository, checks that the repository URL for the remote "origin" equals the provided URL and, if the same branch is referenced as on the previous execution, pulls the remote branch, to make sure the repo is up to date. In case a different branch is referenced, it would fetch the branch, create a local branch for the remote one and checkout the branch. If the branch already exists locally it will jsut be checked out and the remote is pulled to also forward it to the remote repo state. _[This enhanced processing is not implemented in ExecGo. I would also assume that we abandon ExecGo and just go with GoGit. Kept it for now]_

### Open topics:
- How to identify self hosted git deplyments (gitlab/gitea/etc.) github is quite easy, we use the domainname for it. For the others we need a feasible means to determine the git repo kind.
-  No authentication support in place right now.


hellt:
@steiler can you make unit tests pass?

steiler:
@hellt unittests are fine now.
take a look at functinality, if thats all fine with you and then I'll also add some smoke tests.

git/github.go
```diff
@@ -0,0 +1,92 @@
...
+
+		if len(splitPath) > 4 {
+			r.Path = splitPath[4:]
+		}
+
+		r.FileName = "" // no filename, a dir is referenced
+	}
+
+	return r, nil
+}
```
hellt, line 81: @steiler I got rid of a registry (sorry), didn't see a need for it just yet.
Instead a factory function is used that creates a GitRepo out of a user-provided URL which is backed either by GitHubRepo or GitLabRepo.

I've created `NewGitHubRepoFromURL` and the tests for it, but I didn't do anything for GitLab one.
Can you check it out and take GitLab in the same shape?

hellt:
I have created https://gitlab.com/rdodin/clab-test-repo that matches the contents of the github repo.
To be used in integration tests for this feature.

PR #1667, opened 2023-10-24
Fix name filtering in Ignite

robertvolkmann:
Between versions 0.33 and 0.34, a change was implemented where containers, in some cases, are identified by their name instead of the presence of the containerlab label. Unfortunately, only the label filter was implemented for the Ignite runtime.

As a result, duplicate lines are reported for Ignite:

```
+---+--------+--------------+-----------------------------------------+-------+---------+---------------+--------------+
| # |  Name  | Container ID |                  Image                  | Kind  |  State  | IPv4 Address  | IPv6 Address |
+---+--------+--------------+-----------------------------------------+-------+---------+---------------+--------------+
| 1 | leaf01 | 104df67b8e91 | docker.io/networkop/cx:3.7.0            | cvx   | running | 172.17.0.3/24 | N/A          |
| 2 |        |              | docker.io/networkop/cx:3.7.0            | cvx   | running | 172.17.0.3/24 | N/A          |
| 3 | leaf02 | 74ff82043e6e | docker.io/networkop/cx:3.7.0            | cvx   | running | 172.17.0.2/24 | N/A          |
| 4 |        |              | docker.io/networkop/cx:3.7.0            | cvx   | running | 172.17.0.2/24 | N/A          |
+---+--------+--------------+-----------------------------------------+-------+---------+---------------+--------------+
```
Additionally, the same IP address is used in `ansible-inventory.yaml` for all nodes running on the Ignite runtime:
```
all:
  children:
    cvx:
      hosts:
        leaf01:
          ansible_host: 172.17.0.2
        leaf02:
          ansible_host: 172.17.0.2
```

@networkop

### DIFF
--- a/runtime/ignite/ignite.go
+++ b/runtime/ignite/ignite.go
@@ -198,7 +198,7 @@ func (c *IgniteRuntime) StartContainer(ctx context.Context, _ string, node runti
 	vm.Labels = nodecfg.Labels
 	metadata.SetNameAndUID(vm, providers.Client)
 
-	copyFiles := []api.FileMapping{}
+	var copyFiles []api.FileMapping
 	for _, bind := range nodecfg.Binds {
 		parts := strings.Split(bind, ":")
 		if len(parts) < 2 {
@@ -328,7 +328,7 @@ func (c *IgniteRuntime) ListContainers(_ context.Context, gfilters []*types.Gene
 		return result, fmt.Errorf("failed to generate filters: %s", err)
 	}
 
-	filteredVMs := []*api.VM{}
+	var filteredVMs []*api.VM
 	for _, vm := range allVMs {
 		isExpected, err := filters.AreExpected(vm)
 		if err != nil {


robertvolkmann:
@hellt Addressing issues flagged by DeepSource that I didn't introduce can be quite frustrating.

hellt:
Hi @robertvolkmann 
thank you for the PR! Indeed, deepsource sometimes looks at the same file twice and raise more warning over time. We can take care of that.

/cc @steiler

steiler:
> @hellt Addressing issues flagged by DeepSource that I didn't introduce can be quite frustrating.

I also to go through this pain over any over again. It seems like these check have not been there when the referenced code parts where initially introduced. You now touching it triggers a new evaluation. Anyways, Thanks for leaving clab in an even better shape!

hellt:
@robertvolkmann this is now part of 0.47.1

PR #1675, opened 2023-10-26
chown labdir

steiler:
As a final step of the deployment ownership of the lab dir are adjusted to <SUDO_UID>:<SUDO_GID>.
Which will help when cleaning up labs. These days a `sudo rm <labdir> ...` is needed. With this PR no sudo is required anymore.


steiler:
@hwllt This code is depricated. It would break stuff. We need to utilize ACLs.
Have a solution that relies on cgo ready. But will come up with something that solely relies on golang standard libs in a bit.

hellt:
I'll close this then, thanks.

PR #1681, opened 2023-10-29
Add support for apk and Alpine Linux

dBitech:
Add support for installing via apk on Alpine Linux,   note that because there is no signer we have to --allow-untrusted


hellt:
Hi @dBitech!
Thanks for your contribution! Looks good, merging.

PR #1686, opened 2023-10-30
Support for SSH Public Key Provisioning for vr-sros

mabra94:
Added support to automatically provision SSH public keys into vr-sros nodes. 
Used the embed function to store my config command template in a dedicated external file. I liked this approach better than to specify the commands directly in the vr-sros.go file. 

I created a dedicated function for the health check/readiness probe. 

I would like to improve sessions handling a bit though. I would prefer to re-use the same session for the readiness probe, the applyPartialConfig and the applyDefaultConfig functions. Although it works as is, I would be also happy about a pointer on how to improve this.


mabra94:
Thanks for the pointer. Did that

hellt:
superseded by #1706

PR #1689, opened 2023-11-01
Update install.md and improve dev container setup and fix some spelling mistakes.

neilschark:
Hi,

as I use the dev container setup daily for my work, I made some improvements to make using this setup easier. Some of these changes are specific to my use case, but everything which is a general improvement is added in this PR.

And i fixed some spelling mistakes I found on the way.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -207,6 +207,11 @@ Another option to run containerlab on ARM-based Macs is to use Docker in Docker 
         # containelab version will be set in devcontainer.json
         ARG _CLAB_VERSION
 
+        # Set permissions for mounts in devcontainer.json
+        RUN mkdir /home/vscode/.vscode-server/
+        RUN mkdir /home/vscode/.vscode-server/bin
+        RUN chown -R vscode:vscode /home/vscode/.vscode-server
+
         # install some basic tools inside the container
         # adjust this list based on your demands
         RUN apt-get update \
--- a/docs/install.md
+++ b/docs/install.md
@@ -214,8 +214,8 @@ Another option to run containerlab on ARM-based Macs is to use Docker in Docker 
 
         # install some basic tools inside the container
         # adjust this list based on your demands
-        RUN apt-get update \
-            && apt-get upgrade -y
+        RUN apt-get update && \
+            apt-get upgrade -y && \
             apt-get install -y --no-install-recommends \
             sshpass \
             curl \
--- a/docs/install.md
+++ b/docs/install.md
@@ -150,7 +150,7 @@ Once installed, issue `sudo service docker start` to start the docker service in
 
 ### Apple macOS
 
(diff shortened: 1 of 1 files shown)


hellt:
Hi @neilschark 
thanks for your contribution.
Since the contributor of the dev container part is @ankudinov I will let him comment/eval the improvements first.

docs/install.md
```diff
@@ -200,68 +200,89 @@ Another option to run containerlab on ARM-based Macs is to use Docker in Docker
...
+        RUN mkdir /home/vscode/.vscode-server/
+        RUN mkdir /home/vscode/.vscode-server/bin
+        RUN chown -R vscode:vscode /home/vscode/.vscode-server
+
         # install some basic tools inside the container
         # adjust this list based on your demands
         RUN apt-get update \
-            && apt-get install -y --no-install-recommends \
+            && apt-get upgrade -y
+            apt-get install -y --no-install-recommends \
```
hellt, line 219: @neilschark 
I wouldn't do the upgrade here, it will take quite some time and will leave you with a different base OS more frequently then package updates.
Not sure it is worth it, but I may be wrong.

What seems to be clearly missing here though is `&&` after upgrade

docs/install.md
```diff
@@ -200,68 +200,89 @@ Another option to run containerlab on ARM-based Macs is to use Docker in Docker
...
         # The base container already has entrypoint, vscode user account, etc. out of the box
-        FROM mcr.microsoft.com/devcontainers/python:0-3.9-bullseye
+        FROM mcr.microsoft.com/vscode/devcontainers/base:bullseye
 
         # containelab version will be set in devcontainer.json
         ARG _CLAB_VERSION
 
+        # Set permissions for mounts in devcontainer.json
+        RUN mkdir /home/vscode/.vscode-server/
+        RUN mkdir /home/vscode/.vscode-server/bin
```
hellt, line 212: can be simplified
```suggestion
        RUN mkdir -p /home/vscode/.vscode-server/bin
```

docs/install.md
neilschark: I don't think time is a problem here, as the creation of the dev container from scratch depending on the added feature already takes several minutes if you have a slow Internet connection. But after the first building everything is cached, so it doesn't matter anymore, as it is not run again until you pull a clean base image.

I get you second point though, it makes the image possibly different. I don't think it matters much, as for example the docker in docker feature is added with the 'latest' tag, so you will have this problem anyway.

In my experience the base image of debian is updated less enough, that you get some updates in this step regularly. I would guess that the image provided by Microsoft which is the base here is updated even less, so updates with security fixes would likely occur.

But at the and it is not a mayor difference either way, so I would leave it up to the project (and therefore you) to decide.

Just to have it complete I just add the missing '&&' you mentioned. Simply either accept the PR like it is now or tell me to remove the update line, I am fine with it both ways.

hellt:
thnx, lets ship it and later down the road we will see if that works for everybody

PR #1695, opened 2023-11-04
Vsrxlab1

sonicepk:
This is a minimal Juniper vSRX test lab. It includes two Alpine Linux hosts and tests connectivity from one linux host to the other one via the vSRX. This lab also enables the VSRX web gui.


hellt:
Thanks @sonicepk 
I have made slight editorial changes, but the rest looks ok.
Mergin

PR #1708, opened 2023-11-10
Do not attempt creating ssh config files if `ssh_config.d` doesn't exist

hellt:
reported by a user:

```
INFO[0025] Adding containerlab host entries to /etc/hosts file 
INFO[0025] Adding ssh config for containerlab nodes     
ERRO[0025] failed to create ssh config file: open /etc/ssh/ssh_config.d/clab-srl01.conf: no such file or directory 
```

this means the path did not exist and this PR ensures the path exists.
It might be that the OS user runs or OpenSSH there won't honor the newly created path either, but it seems still safe to create the path

### DIFF
--- a/clab/sshconfig.go
+++ b/clab/sshconfig.go
@@ -2,8 +2,10 @@ package clab
 
 import (
 	"os"
+	"path"
 	"text/template"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/srl-labs/containerlab/types"
 	"github.com/srl-labs/containerlab/utils"
 )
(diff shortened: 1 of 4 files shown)


steiler:
Hmm my feeling is more like, if that path does not exist, we do not generate the config rather then generating something that is not used anyways.

hellt:
I will ask the user to create this file and see if his ssh client picks it up

hellt:
@steiler re-applied the patch to follow the suggested logic, as old ssh clients do not respect the created path.

PR #1709, opened 2023-11-10
Fixing Test related to SRL 23.10

steiler:
fix tests


nodes/srl/srl.go
```diff
@@ -598,12 +598,15 @@ func (n *srl) addDefaultConfig(ctx context.Context) error {
...
 	// in srlinux >= v23.10+ linuxadmin and admin user ssh keys can only be configured via the cli
 	// so we add the keys to the template data for rendering.
 	if len(n.sshPubKeys) > 0 && (semver.Compare(n.swVersion.String(), "v23.10") >= 0 || n.swVersion.major == "0") {
 		tplData.SSHPubKeys = catenateKeys(n.sshPubKeys)
+	} else {
+		// prior to 23.10.1 only rsa keys are supported
+		// this function filters out non-rsa keys from the
+		// list of ssh public keys found on the system/agent
+		n.filterSSHPubKeys()
 	}
```
hellt, line 610: @steiler I brought back the filtering function since it is required for SLR < 23.10.1

hellt (approved):

nodes/srl/srl.go
hellt: I am dumb. Filtering is not needed as now we support all types of keys in the config in 23.10, and older releases use mounted files, where, again, all keys can be put.

PR #1721, opened 2023-11-13
Remove unbound ssh option for older clients

### LINKED ISSUE #1719: Use `unbound` ssh option only for ssh clients supporting it.
When we started to use unbound option to mitigate SR OS SSH server not understanding modern ssh options we started to do this for all systems.

But apparently picky ssh clients bail out when they see unbound value being used which they don't recognize:

```
ssh -l admin clab-nc-sr1
/etc/ssh/ssh_config.d/clab-nc.conf line 6: unsupported option "unbound".

ssh -V
OpenSSH_8.0p1, OpenSSL 1.1.1k  FIPS 25 Mar 2021
```

I think we need to
1. identify which ssh client version started to support this option by running `ssh -V`
2. grep the version from it
3. and only set it for clients newer than the release that first started to support this option.

## OpenSSH version

OpenSSH version 8.9 is the first release where unbound option is supported:

```
OpenSSH_8.8p1, OpenSSL 1.1.1v  1 Aug 2023 - does not work
```

```
OpenSSH_8.9p1 Ubuntu-3ubuntu0.3, OpenSSL 3.0.2 15 Mar 2022 - works
```

```
OpenSSH_9.0p1, OpenSSL 1.1.1v  1 Aug 2023 - works
```

hellt:
fix #1719

### DIFF
--- a/clab/sshconfig.go
+++ b/clab/sshconfig.go
@@ -1,6 +1,7 @@
 package clab
 
 import (
+	_ "embed"
 	"os"
 	"path"
 	"text/template"
@@ -8,6 +9,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/srl-labs/containerlab/types"
 	"github.com/srl-labs/containerlab/utils"
+	"golang.org/x/mod/semver"
 )
 
 // SSHConfigTmpl is the top-level data structure for the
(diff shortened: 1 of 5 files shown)


clab/sshconfig.go
```diff
@@ -73,10 +71,19 @@ func (c *CLab) AddSSHConfig() error {
...
 		}
+
+		// if we couldn't parse the ssh version we assume we can't use unbound option
+		// or if the version is lower than 8.9
+		// and the node has the PubkeyAuthentication set to unbound
+		// we set it to empty string since it is not supported by the SSH client
+		if (sshVersion == "" || semver.Compare(sshVersion, "8.9") < 0) && nodeData.SSHConfig.PubkeyAuthentication == types.PubkeyAuthValueUnbound {
+			nodeData.SSHConfig.PubkeyAuthentication = ""
+		}
+
```
hellt, line 82: @steiler I am going to merge this one so that people who upgrade won't have to wait long for this crit fix, but if you find a better way to handle this (like grepping ssh version once and storing it centrally) then we can revisit this in a subsequent PR

PR #1723, opened 2023-11-14
Fixing nil pointer deref on Scrapli send config errors

steiler:


nodes/vr_sros/vr-sros.go
```diff
@@ -280,19 +280,16 @@ func (s *vrSROS) applyPartialConfig(ctx context.Context, addr, platformName, use
...
-	}
+	cfgs = append(cfgs, "commit", "/admin save")
 
-	r, err = d.SendCommand("/admin save")
-	if err != nil || r.Failed != nil {
-		return fmt.Errorf("failed to persist config; error: %+v %+v", err, mr.Failed)
+	mr, err := d.SendConfigs(cfgs)
+	if err != nil || (mr != nil && mr.Failed != nil) {
+		if mr != nil {
+			return fmt.Errorf("failed to apply config; error: %+v %+v", err, mr.Failed)
```
hellt, line 289: this one is not needed, as we check for `mr != nil` above

steiler: thats not correct.
its an OR up there ... we could do it in to seperate branches but ... same same

hellt: then here it shoudl also be `if mr != nil && err != nil`
otherwise the statement above might have err == nil, but mr.Failed != nil

hellt:
fix verified, thanks @steiler !

PR #1724, opened 2023-11-14
Adjusting ssh key indexes to start from max (32) to go backwards

steiler:


hellt (approved):
@steiler check the two commits I added. They are just test refactoring, so I believe you won't have objections to that.

Thanks for picking this up, looks good.

PR #1727, opened 2023-11-16
Removing the binary integration of git

steiler:
We settled for the go-git implementation, so this is no longer needed


hellt:
Thnx!

PR #1738, opened 2023-11-22
Destroy all: retrieve labdir from container labels

steiler:
We where always expecting the labdir to be a subdir of the actual pwd. With this PR we will query the labels and deduce the labdir from the "containerlab" label.


hellt:
Thanks for checking this out! Mark it ready when you are.
May be a need for 0.48.4 :D

Issue #1700, opened 2023-11-06
Integration of Health Checks with wait-for Not Honoring Container 'Healthy' Status


FloSch-Nokia:
### Background
The recent addition of health checks to Containerlab, implemented in [Pull Request #1426](https://github.com/srl-labs/containerlab/pull/1426), has been a significant advancement aligning Containerlab closer to Docker Compose's feature set. However, it has been observed that the `wait-for: <containername>` functionality is not fully integrating with these health checks, specifically, it does not wait for the container it is dependent on to reach a 'healthy' state before initiating.

### Relation to Past Issues and Discussions
This behavior seems related to previous discussions which can be seen in:
- [Issue #1042](https://github.com/srl-labs/containerlab/issues/1042) and [Discussion #1048](https://github.com/srl-labs/containerlab/discussions/1048) regarding `wait-for` functionality for vrnetlab containers.
- [Issue #1027](https://github.com/srl-labs/containerlab/issues/1027) on DependencyManager enhancement, where the difference between a container being 'readily configured' versus 'healthy' was brought up, and a redesign to allow the post-deploy phase to run per node was considered.

### Current Problem
The `wait-for` mechanism proceeds with subsequent container startups without ensuring that the prior container has a 'healthy' status. This is inconsistent with the orchestration logic expected by users who are familiar with Docker Compose's health check dependencies.

### Expected Behavior
It is anticipated that the `wait-for` flag would delay the startup of dependent containers until the preceding container is marked 'healthy' by the new health check system.

### Steps to Reproduce
1. Implement health checks for container A.
2. Set up container B to include `wait-for: A`.
3. Start the Containerlab environment.
4. Notice that the initiation of container B is not delayed until container A reaches a 'healthy' state.

### Impact
This issue undermines the reliability and predictability of container startup sequences in orchestrated environments, particularly where containers have explicit dependencies on the operational readiness of others.

### Suggestion for Resolution
Revising the `wait-for` functionality to integrate a polling mechanism for health check status could provide a solution. This enhancement should consider the health check implementation details in [PR #1426](https://github.com/srl-labs/containerlab/pull/1426) and the insights from the related issues mentioned above.

Appreciating the team's efforts and looking forward to a collaborative resolution.

hellt:
done in #1759, thanks @steiler !

PR #1759, opened 2023-11-30
Node Stages and Dependency Manager

steiler:
Implementing #1700 

This change will first of all execute the Post-Deploy phase as another step in the scheduleNode func.
This will allow for a wider variety of dependencies. e.g. just deploying a node when another is already healthy.
Further the new struct for WaitFor allows to wait for specific states of other nodes, before certain deployment stages are started.
Multiple dependencies for different or the same phase are also possible.

So the structure of the `WaitFor `field changes as follows:

```
    NodeC:
      ...
      wait-for:
        create:                    # For NodeB to enter the _create_ stage, the following preconditions must be met
          - node: NodeA            # NodeA
            phase: healthy         # Must have entered the healthy
        configure:                 # For NodeB to enter the _configure_ stage, the following preconditions must be met
          - node: NodeB            # NodeB
            phase: create          # Must have finished the _create_ stage
```

These phases are used in two means.
```
  wait-for:
    create:                      # Means 1
      - node: xyz
        phase: healthy           # Means 2
```
**Means 1** defines what needs needs to happen before the node enters the **Means 1** phase / stage.
**Means 2** on the other hand defined which phase / stage node xyz must have finished for the enclosing node to start the _Means 1_ phase / stage.

The available phases are:
- **create** - Container create and start
- **create-links** - Deploy container links
- **configure** - Deploy container configuration
- **healthy** - Container turns healthy (only viable as a _Means 2_ statement)
- **exit** - Container exists (usefull for init containers, also only viable as _Means 2_ statement

Taking a look at this example:
```
name: linux

topology:
  kinds:
    linux:
      image: alpine
      cmd: sleep infinity
  nodes:
    l1:
      kind: linux
    l2:
      kind: linux
      wait-for:
        configure:
          - node: l1
            phase: create
    l3:
      kind: linux
      image: mavatest
      wait-for:
        create:
          - node: l2
            phase: configure
    l4:
      kind: linux
      wait-for:
        create:
          - node: l3
            phase: healthy
        configure:
          - node: l1
            phase: configure
```
Node `l1` starts without any dependency. For `l2` to enter the configure/post-deploy stage, `l1` is required to have finished the create phase.
Further `l3` will start its creation as soon as node `l2` finished its configure/post-deploy stage. `l4` will get created as soon as `l3` turned healthy (make sure healthchecks are enabled*) and node `l1` is done with its configure phase.

*We have the healthcheck PR still pending, that enables you to define healthchecks as part of the topology or you rely on healthcheck defined already via the [Dockerfile -> Healthcheck](https://docs.docker.com/engine/reference/builder/#healthcheck)


types/types.go
```diff
@@ -181,9 +181,9 @@ type NodeConfig struct {
 	Memory string  `json:"memory,omitempty"`
 
 	// Extra node parameters
-	Extras  *Extras    `json:"extras,omitempty"`
-	WaitFor []string   `json:"wait-for,omitempty"`
-	DNS     *DNSConfig `json:"dns,omitempty"`
+	Extras  *Extras                     `json:"extras,omitempty"`
+	WaitFor map[WaitForPhase][]*WaitFor `json:"wait-for,omitempty"`
```
hellt, line 185: @steiler  this is cool

I wonder how do you feel about this layout:


```yaml
name: linux

topology:
  kinds:
    linux:
      image: alpine
  nodes:
    l1:
      kind: linux
    l2:
      kind: linux
      stage:
        configure:
          wait-for:
            - node: l1
              status: created
    l3:
      kind: linux
      image: mavatest
      stage:
        configure:
          wait-for:
            - node: l2
              status: configured
    l4:
      kind: linux
      stage:
        create:
          wait-for:
            - node: l3
              status: healthy
        configure:
          wait-for:
            - node: l1
                status: configured
```

The idea here is that nodes have a `stage` that denotes a certain stage a node goes through the deployment process.

Each stage has a `wait-for` that works like you cleverly did here.
But it also allows us to parameterized stages by introducing other nested structs.
Like

```yaml
    l2:
      kind: linux
      stage:
        configure:
          skip: true
```

steiler: Uhhh I see nice idea...

steiler:
CAREFULL, the structure of the topo file has changed to what is documented in the top.
It now follows the structure @hellt suggested further up.

types/node_definition.go
```diff
@@ -68,8 +68,8 @@ type NodeDefinition struct {
 	Sysctls map[string]string `yaml:"sysctls,omitempty"`
 	// Extra options, may be kind specific
 	Extras *Extras `yaml:"extras,omitempty"`
-	// List of node names to wait for before satarting this particular node
-	WaitFor []string `yaml:"wait-for,omitempty"`
+	// stage based dependency definition
```
hellt, line 71: @steiler I want us to get rid of the

```go
Stage map[string]*WaitForEnvelop `yaml:",inline"`
```

and instead do a struct-based stage config

```go
// Stages represents a configuration of stages for a given node.
type Stages struct {
	Create      StageConfig `yaml:"create,omitempty"`
	CreateLinks StageConfig `yaml:"create-links,omitempty"`
	Configure   StageConfig `yaml:"configure,omitempty"`
	Healthy     StageConfig `yaml:"healthy,omitempty"`
	Exit        StageConfig `yaml:"exit,omitempty"`
}

// StageConfig represents a configuration of a given stage.
type StageConfig struct {
	WaitFor []*WaitFor `yaml:"wait-for,omitempty"`
}
```

We have a finite amount of stages, and doing a struct with fields feels more appropriate.
Unfortunately this means that changes to GetWaitFor are required, and I think it may be rather `GetStages()` as this is now the top level container that dependency manager should work on

hellt: to identify one stage from the other we might create different types, like

```go
type CreateStageConfig struct {
  StageConfig
}
```
with an embedded common StageConfig. Then a String() method on those structs could return the string stage if required

steiler: If we go down that path, we might even have seperate StageConfig structs.
Because There are certain options which will only be viable in certain stages.
So we could tage the generic StageConfig and implement that basic case. But include it in the specific structs, whcih would allow us to overwrite certain things per stage... kind of like the DefaultNode and Specific nodes implementations.
opinion?

hellt: This seems to match what I added here https://github.com/srl-labs/containerlab/pull/1759#discussion_r1452051061 ?

steiler: *Facepalm* ... yes! Sorry!

types/types.go
```diff
@@ -181,10 +181,9 @@ type NodeConfig struct {
 	Memory string  `json:"memory,omitempty"`
 
 	// Extra node parameters
-	Extras  *Extras    `json:"extras,omitempty"`
-	WaitFor []string   `json:"wait-for,omitempty"`
-	DNS     *DNSConfig `json:"dns,omitempty"`
-
+	Extras  *Extras                     `json:"extras,omitempty"`
+	WaitFor map[WaitForPhase][]*WaitFor `json:"wait-for,omitempty"`
```
hellt, line 185: and similar to https://github.com/srl-labs/containerlab/pull/1759/files#r1464920202 I think this should rather be `Stages`, as WaitFor is now part of the stage, and not directly a property of a Node

runtime/docker/docker.go
```diff
@@ -1003,3 +1003,16 @@ func (d *DockerRuntime) containerPid(ctx context.Context, cID string) (int, erro
 	}
 	return inspect.State.Pid, nil
 }
+
+// GetContainerHealth returns true is the container is reported as being healthy, false otherwise.
+func (d *DockerRuntime) GetContainerHealth(ctx context.Context, cID string) (bool, error) {
```
hellt, line 1008: I think `IsHealthy` is a better func name here.

runtime/docker/docker.go
```diff
@@ -1003,3 +1003,16 @@ func (d *DockerRuntime) containerPid(ctx context.Context, cID string) (int, erro
...
+	inspect, err := d.Client.ContainerInspect(ctx, cID)
+	if err != nil {
+		return false, err
+	}
+	// catch no healtchecks defined
```
hellt, line 1013:
```suggestion
	// catch no healthchecks defined
```

clab/dependency_manager/dependency_manager.go
```diff
@@ -2,101 +2,30 @@ package dependency_manager
...
 	// AddNode adds a node to the dependency manager.
 	AddNode(name string)
 	// AddDependency adds a dependency between depender and dependee.
 	// The depender will effectively wait for the dependee to finish.
-	AddDependency(dependee, depender string) error
+	AddDependency(depender string, dependerState types.WaitForPhase, dependee string, dependeeState types.WaitForPhase) error
 	// WaitForNodeDependencies is called by a node that is meant to be created.
 	// This call will bock until all the nodes that this node depends on are created.
-	WaitForNodeDependencies(nodeName string) error
+	Enter(nodeName string, state types.WaitForPhase) error
```
hellt, line 20: what does `Enter` do? The comment is from the old func

clab/dependency_manager/dependency_manager.go
```diff
@@ -2,101 +2,30 @@ package dependency_manager
...
 	// SignalDone is called by a node that has finished all tasks for the provided State.
 	// internally the dependent nodes will be "notified" that an additional (if multiple exist) dependency is satisfied.
-	SignalDone(nodeName string, state NodeState)
+	SignalDone(nodeName string, state types.WaitForPhase)
+	GetDependerCount(nodeName string, state types.WaitForPhase) (uint8, error)
```
hellt, line 24: why uint8? maybe uint is better to not think about potential constraints

clab/clab_test.go
```diff
@@ -46,14 +55,23 @@ func Test_createStaticDynamicDependency(t *testing.T) {
...
-	dm.EXPECT().AddDependency("node4", "node2")
-	dm.EXPECT().AddDependency("node4", "node3")
-	dm.EXPECT().AddDependency("node5", "node1")
-	dm.EXPECT().AddDependency("node5", "node2")
-	dm.EXPECT().AddDependency("node5", "node3")
+	clab := &CLab{
+		Nodes: nodeMap,
+	}
+
+	err := WithDependencyManager(dm)(clab)
```
hellt, line 62: why not `clab.WithDependencyManager(dm)`?

clab/clab.go
```diff
@@ -527,22 +542,109 @@ func (c *CLab) scheduleNodes(ctx context.Context, maxWorkers int,
...
 					continue
 				}
 
-				// signal to dependency manager that this node is done with creation
-				dm.SignalDone(node.Config().ShortName, dependency_manager.NodeStateCreated)
+				// DeployLinks does not necessarily create all the links already.
+				// veth links might be created by the peer node, if it was not already
+				// deployed properly. Hence we need to wait for all the links to be created.
+				// we just wait if there is actually a dependency on this state, otherwise
+				// we head on.
```
hellt, line 569: everything below this block makes the function very long
let's try to carve it to decompose it

cmd/deploy.go
```diff
@@ -229,14 +225,8 @@ func deployFn(_ *cobra.Command, _ []string) error {
...
-	log.Debug("containers created, retrieving state and IP addresses...")
-	// updating nodes with runtime information such as IP addresses assigned by the runtime dynamically
-	for _, n := range c.Nodes {
-		err = n.UpdateConfigWithRuntimeInfo(ctx)
-		if err != nil {
-			log.Errorf("failed to update node runtime information for node %s: %v", n.Config().ShortName, err)
-		}
-	}
+	// write to log
+	execCollection.Log()
```
hellt, line 229: is there a need to return execCollection to the caller (deploy)
maybe the log should happen where the exec collection is populated/executed?

hellt (changes requested):

clab/dependency_manager/dependency_manager.go
steiler: A Node calls Enter with it's own name and the Phase it tries to Enter... Blocking if there are dependencies that are not met yet.
Considered it a better name then WaitForNodeDependencies. This should also change in the next phase and become a Node bound private func that would not take the name any more.
Will change the comment.

clab/dependency_manager/dependency_manager.go
steiler: ok

clab/clab.go
steiler: resonable, yes

runtime/docker/docker.go
steiler: Done

runtime/docker/docker.go
steiler: Done

clab/dependency_manager/dependency_manager.go
steiler: Done

clab/dependency_manager/dependency_manager.go
steiler: Done -> uint

clab/clab_test.go
steiler: `WithDependencyManager ` yields a `ClabOption`. 
`ClabOption` is defined in the following way:
```
type ClabOption func(c *CLab) error
```
So it will apply itself on the clab instance.
Hence we're also checking the `WithDependencyManager ` option. A `clab.WithDependencyManager(dm)` does not exist.
We could only set the option `manually` via:
```
	clab := &CLab{
		Nodes: nodeMap,
		dependencyManager: dm,
	}
```
But this way it makes perfect sense to me.

cmd/deploy.go
steiler: I wanted to make sure, that the exec output is printed as a collection of outputs en bloc.
So thats why we use the collection here.
As you see, we have the `WaitGroup` just right below, because the main program flow is simply spawning go funcs to create the containers and uses the WaitGroup below, to "sync" with the async container creations. 
So only when that WaitGroup unblocks we can be sure all the output is generated and print it as a consecutive block rather then having it scattered all over the log.

https://github.com/srl-labs/containerlab/blob/6c009855625a3afe9ef05da9835879d6c6d40e16/cmd/deploy.go#L219-L229

clab/clab.go
hellt: postponing for a later enhancements

steiler: Agreed to be tackled in an upcoming PR.

clab/clab.go
```diff
@@ -520,29 +534,123 @@ func (c *CLab) scheduleNodes(ctx context.Context, maxWorkers int,
...
+				count, err := dm.GetDependerCount(node.GetShortName(), types.WaitForCreateLinks)
+				if err != nil {
+					log.Error(err)
+				}
```
hellt, line 579: @steiler I am not sure this GetDependerCount logic is required at all.
Don't we always want to ensure that all links were created?

All other stages require links to be provisioned, so maybe we can just drop this additional logic?

clab/clab.go
```diff
@@ -517,32 +531,126 @@ func (c *CLab) scheduleNodes(ctx context.Context, maxWorkers int,
...
+
+				// health state processing
+				count, err = dm.GetDependerCount(node.GetShortName(), types.WaitForHealthy)
+				if err != nil {
+					log.Error(err)
+				}
+				if count > 0 {
+					// if there is a dependecy on the healthy state of this node, enter the checking procedure
+					for {
+						healthy, err := node.GetRuntime().IsHealthy(ctx, node.Config().LongName)
```
hellt, line 615: @steiler it feels a bit off that IsHealthy is a method of a runtime and not the node itself.
Doesn't it make more sense to have `node.IsHealthy()`?

clab/dependency_manager/dependency_node.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// newDependencyNode initializes a dependencyNode with the given name.
+func newDependencyNode(name string) *dependencyNode {
+	d := &dependencyNode{
+		name:                name,
+		stageBarrier:        map[types.WaitForStage]*sync.WaitGroup{},
+		stageBarrierCounter: map[types.WaitForStage]uint{},
+		depender:            map[types.WaitForStage][]*dependerNodeStage{},
+	}
```
hellt, line 28: Do we need `stageBarrierCounter` at all? Is it different from `len(d.depender[stage])`?

hellt:
@steiler also check out this commit
[580e852](https://github.com/srl-labs/containerlab/pull/1759/commits/580e85252b252ff35ac6d645b214624d6f0a9acc)

I think it makes things clearer, as no extra indirection is involved, no Enter/SignalDone on the depMgr level, these function belong to the dependencyNode

clab/clab.go
steiler: we'll the thing is, that the first node, that is deployed, cannot create any links, because all the other nodes are not there yet. Waiting for all the nodes to complete to then create the links is something we got rid of lately. All the links these days are created by the second node (per link). If we need all the links to be created, we have to explicitly wait for all the links. If there is no dependency on the creation of links, why would we wait, we can simply carry on and speed up the deployment.
What I did was to change the count to a simple boolean. Because I realized the count does not really matter, it is more the "if" (we have to wait) that matters. 
This logic also reapeats itself on the healthy and the exit phase ...
This is crutial, since many containers do not have a healthcheck defined, so we would never be done with the waiting for healthy. So clan deploy would never terminate. Same applies to the exit state. Mostly you will not wait for exit state, which is basically our mechanism that enables init containers. But for all other containers we don't want to wait for the exit state, because clab would "never" terminate.

clab/clab.go
steiler: Agree. We even have a node.GetState() but it does not resembel whats required here. I'll take a look if we can unify it or if we need a seperate function.

steiler: alright did introduce that

clab/dependency_manager/dependency_node.go
steiler: Changed stageBarrierCounter to MustWait, which is only a boolean value, indicating if the given stage is to be waited for or it can be continued with the next phase.

steiler:
> @steiler also check out this commit [580e852](https://github.com/srl-labs/containerlab/pull/1759/commits/580e85252b252ff35ac6d645b214624d6f0a9acc)
> 
> I think it makes things clearer, as no extra indirection is involved, no Enter/SignalDone on the depMgr level, these function belong to the dependencyNode

Yes that is alright. Would be a natural change in the upcoming change, we discussed. But it's fine to do it right now.

hellt (approved):

PR #1768, opened 2023-12-05
Update Border0 FKA Mysocketio Commands + Docs

adrianosela:
## Update Border0 FKA Mysocketio Commands + Docs

As of November 2023, Border0 deprecated programmatic authentication. This changes the code and the corresponding documentation to use their OAuth2.0 "Device Authorization Code" flow.

The code for the flow mostly comes from the Border0 CLI's source code directly: https://github.com/borderzero/border0-cli/blob/main/cmd/login.go#L42

Except its made to use utility functions present in this code base.

### DIFF
--- a/border0_api/border0.go
+++ b/border0_api/border0.go
@@ -11,12 +11,19 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/cenkalti/backoff"
+	"github.com/golang-jwt/jwt"
 	log "github.com/sirupsen/logrus"
+	"github.com/skratchdot/open-golang/open"
 	"github.com/srl-labs/containerlab/nodes"
 	"github.com/srl-labs/containerlab/utils"
 	"gopkg.in/yaml.v2"
@@ -24,8 +31,10 @@ import (
 
 const (
 	apiUrl                       = "https://api.border0.com/api/v1"
+	portalUrl                    = "https://portal.border0.com"
 	ENV_NAME_BORDER0_ADMIN_TOKEN = "BORDER0_ADMIN_TOKEN"
 	ENV_NAME_BORDER0_API         = "BORDER0_API"
+	ENV_NAME_BORDER0_PORTAL      = "BORDER0_PORTAL"
 )
 
 var supportedSockTypes = []string{"ssh", "tls", "http", "https"}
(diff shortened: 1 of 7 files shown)


atoonk:
I tested it on a digital ocean machine; quick test shows it works as I'd expect. Nice work @adrianosela  cc: @hellt 
to confirm, with this change the new login command is  `containerlab tools border0 login` 
which will print:
```
Please navigate to the URL below in order to complete the login process:
https://portal.border0.com/login?device_identifier=123456789abc.........
Login successful
```

the user will be prompted to open the browser, and they'll need to authorize / confirm, the log in attempt.

hellt:
thanks @adrianosela, @atoonk 

@steiler  would you mind reviewing this? Thanks

border0_api/border0.go
```diff
@@ -11,54 +11,179 @@ import (
...
+		i++
 	}
-	// prepare a LoginRequest
-	loginReq := &LoginRequest{
-		Email:    email,
-		Password: password,
-	}
-	// init a LoginResponse
-	loginResp := &LoginResponse{}
+}
```
hellt, line 143: @adrianosela how about using expo backoff and get rid of `i`?

```go
// polling for auth token
for {
    exponentialBackoff := backoff.NewExponentialBackOff()
    exponentialBackoff.InitialInterval = 1 * time.Second
    exponentialBackoff.MaxInterval = 5 * time.Second
    exponentialBackoff.Multiplier = 2

    var token *deviceAuthorizationStatus
    var err error

    err = backoff.Retry(func() error {
        token, err = getDeviceAuthorizationStatus(ctx, deviceAuthToken)
        return err
    }, exponentialBackoff)

    if err != nil {
        log.Fatalf("We couldn't log you in, make sure that you have authenticated using the link above: %v", err)
    }

    if token != nil && token.Token != "" && token.State != "not_authorized" {
        fmt.Println("Login successful")
        return token.Token, nil
    }
}
```

adrianosela: Yeah, that poor logic was copied verbatim from the b0 cli code. I made the change here - and I will be making it there too! Thank you ;) 

I also fixed the lines that made your linter unhappy.

Thanks for getting on this quickly <3

adrianosela:
Rebased onto srl-labs/containerlab `main`

adrianosela:
Here's a short demo of the new flow too with the latest code:

https://github.com/srl-labs/containerlab/assets/16856511/db82851e-0357-435a-acbd-e92790c1caeb

adrianosela:
TL;DR; your failed logins are expected - try without the `-e` flag.

Hey @steiler we've deprecated programmatic logins. Only folks that have requested allowlisting with border0 are allowed to log in with username and password. 

We see your programmatic auth attempts in our API logs and can confirm we blocked them due to your email not being in the aforementioned allowlist.

Note that I hid the email and password flags for this reason. If I allowlist you, you will be able to log in that way, however we ask that all new customers log in with only `./bin/containerlab tools border0 login` (without the `-e` / `--email` flag).

The point of this is to protect our customers against credential stuffing attacks against the Border0 service (e.g. an attacker using credentials from a breach of a different web service against our API). For more: https://auth0.com/blog/what-is-credential-stuffing/

For programmatic access we require generating an access token (i.e. service identity) from the border0 portal -- which should still work with container labs by providing the token via the [`BORDER0_ADMIN_TOKEN` env var](https://github.com/srl-labs/containerlab/blob/ee3ef23449afad09c755065b5ad746b04810e312/border0_api/border0.go#L27) to containerlabs commands.

hellt:
Since there are no touches to the core clab part, and we don't have existing b0 users (cause we had it broken for a while with no cycles to support this integration) I will just roll it in and then later in 2024 we will see how to take "lab exposure" part further.

Thanks all.

adrianosela:
@hellt let us/me know how we can help with that :)

If direct means of communication fail you can reach us at support@border0.com or via our community slack https://join.slack.com/t/border0community/shared_invite/zt-zbx586ls-x44z7I3POLPQfesRWnig7Q

hellt:
@adrianosela I think the first step would be to transition the existing b0 integration to the go API that you recently added.
When we delivered the integration it was not available, so we had to do a quick API ourselves.

Switching to the upstream API would rather make the integration leaner and more straightforward.

adrianosela:
@hellt Let's talk next year. We maaaaay (no promises) have something in the works to simplify the integration and make the experience much nicer for clab users.

hellt:
sounds exciting @adrianosela 
I was also thinking to have an agent for our SR Linux Network OS, something we can talk through next year as well.

PR #1769, opened 2023-12-05
Prevent cleanup on CTRL-C

steiler:
Adjust deploy -> ctrl-c to not cleanup as requested in #1761


cmd/deploy.go
```diff
@@ -386,7 +386,7 @@ func setupCTRLCHandler(cancel context.CancelFunc) {
 		cancel()
 
 		// when interrupted, destroy the interrupted lab deployment with cleanup
```
hellt, line 388:
```suggestion
		// when interrupted, destroy the interrupted lab deployment
```

PR #1770, opened 2023-12-05
SRL: Populate NodeConfig with generated System MAC for it to show up in the json data export

steiler:
Set the generated SRL system Mac in the Nodeconfig for the json export


hellt:
thanks, merging?

steiler:
yes, go ahead

PR #1773, opened 2023-12-08
Fix exec filters with bad filters

### LINKED ISSUE #1748: exec doesn't exit with -1 code when no containers are found matching the filter
Consider the following command:

```
sudo clab exec --label containerlab=greeter --cmd 'ip l'
```

even when no containers are matched by the filter clab silentry exists with rc=0

But this is counter intuitive, if no containers found we must exist with rc=1 and an error that no containers found.

friday963:
Fixes #1748 .

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -97,7 +97,10 @@ func execFn(_ *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-
+	err = c.GlobalRuntime().CheckFilterLength(cnts)
+	if err != nil {
+		return err
+	}
 	// prepare the exec collection and the exec command
 	resultCollection := exec.NewExecCollection()
 
--- a/mocks/mockruntime/runtime.go
+++ b/mocks/mockruntime/runtime.go
@@ -7,7 +7,7 @@ package mockruntime
 import (
 	context "context"
 	reflect "reflect"
-
+	"errors"
 	gomock "github.com/golang/mock/gomock"
 	exec "github.com/srl-labs/containerlab/clab/exec"
 	links "github.com/srl-labs/containerlab/links"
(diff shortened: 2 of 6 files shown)


friday963:
`friday@ubuntu:~/code/containerlab/bin$ sudo ./containerlab exec --label clab-node-kind=linux --cmd "ls"
INFO[0000] Executed command "ls" on the node "clab-apline_test-host1". stdout:
bin
dev
etc
home
lib
friday@ubuntu:~/code/containerlab/bin$ sudo ./containerlab exec --label clab-node-kind=nokia_srlinux --cmd "ls"
Error: filter did not match any containers
friday@ubuntu:~/code/containerlab/bin$ sudo ./containerlab exec --label clab-node-kind=eos --cmd "ls"
Error: filter did not match any containers`

hellt:
you can now refresh the main branch and rebase your commits on top of it to resolve the testing issue with border0 failure.

steiler:
@friday963 there is still a DeepSource issue you need to fix here

steiler:
@friday963 as Roman mentioned in the Discord thread, the implementation is weired.
You're getting the GenericContainers and just have to check for its length. You do not at all need to involve the (global)runtime again in this.
As soon as you copy / paste code you should always ask yourself, what is going wrong. (see: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
Your commit "add tests" does also not add any tests but updates the mock for the Runtime interface.
Mocks are used in tests. But are not tests. If you want to learn more (https://en.wikipedia.org/wiki/Mock_object).
Hope you find these references and explenations useful. 
This PR I will close since I've opened #1786 that honors the above. Take a look. It implements a real test (integration not unit-test in this case.)
Thanks anyways for your effort. Looking forward to further contributions.

PR #1824, opened 2024-01-20
Documentation updates

slinkish:
Fixed issue with sudo/escalation losing env variable declared in line and then made the PR slightly more substantial pulling in an open doc issue.


hellt:
Thanks a lot @slinkish

PR #1853, opened 2024-02-02
Support mounting anonymous volume

mzagozen:
The `docker run ... -v /path/to/dir` syntax creates an anonymous volume and mounts it to the given destination path in the container.

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -497,6 +497,11 @@ func (c *CLab) resolveBindPaths(binds []string, nodedir string) error {
 		// host path is a first element in a /hostpath:/remotepath(:options) string
 		elems := strings.Split(binds[i], ":")
 
+		if len(elems) == 1 {
+			// if there is only one element, it means that we have an anonymous
+			// volume, in this case we don't need to resolve the path
+			continue
+		}
 		// replace special variable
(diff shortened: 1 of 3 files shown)


hellt:
Thanks @mzagozen 

can you add a bit of a context how are you using this anon volume so that we can provide some hints in the docs?

mzagozen:
Sure, we have topologies that include a Cisco NSO container based on [nso-docker], like so:

```yaml
topology:
    nso:
      kind: linux
      image: ${NSO_IMAGE}
      binds:
        - /var/opt/ncs/packages

```
The path `/var/opt/ncs/packages` is the NSO package load path - where it loads user code from. We have an existing development workflow where we use this anonymous volume to then update the code (NSO packages) in the running NSO container by running a sidecar container for compilation. This sidecar attaches to the anonymous volume mounted in the NSO container with the command `docker run --volumes-from ${NSO_CNT} ...`

From https://gitlab.com/nso-developer/nso-docker/-/blob/master/skeletons/system/testenvs/testenv-common.mk?ref_type=heads#L66:
```shell
docker run -t --rm -v $(PROJECT_DIR):/src --volumes-from $${NSO} $(NSO_IMAGE_PATH)cisco-nso-dev:$(NSO_VERSION) bash -lc 'rsync -aEim -- ...
```
[nso-docker]: https://gitlab.com/nso-developer/nso-docker

hellt:
thanks

PR #1864, opened 2024-02-05
Count management interface in CLAB_INTFS_WITH_MGMT env var

### LINKED ISSUE #1851: network-mode:none - clab_intfs calculation logic
Hello, 

I am trying to use network-mode: none with VM-based node, but node does not start:
```
name: vsrx-cluster
topology:
  nodes:
    node0:
      kind: juniper_vsrx
      image: vrnetlab/vr-vsrx:23.2R1.13
      network-mode: none
      env:
        CLAB_INTFS: 0

```
```
| # |           Name           | Container ID |           Image            |     Kind     |  State  |   IPv4 Address   |     IPv6 Address     |
+---+--------------------------+--------------+----------------------------+--------------+---------+------------------+----------------------+
| 1 | clab-vsrx-cluster-node0  | 001d7344ec31 | vrnetlab/vr-vsrx:23.2R1.13 | juniper_vsrx | running | N/A              | N/A                  |

```
```
$ sudo docker logs -f 001d7344ec31
2024-02-01 12:46:49,838: vrnetlab   DEBUG    Creating overlay disk image
2024-02-01 12:46:49,854: vrnetlab   DEBUG    Starting vrnetlab VSRX
2024-02-01 12:46:49,854: vrnetlab   DEBUG    VMs: [<__main__.VSRX_vm object at 0x7fb4aac6e250>]
2024-02-01 12:46:49,860: vrnetlab   DEBUG    VM not started; starting!
2024-02-01 12:46:49,860: vrnetlab   INFO     Starting VSRX_vm
2024-02-01 12:46:49,861: vrnetlab   DEBUG    number of provisioned data plane interfaces is 1
2024-02-01 12:46:49,861: vrnetlab   DEBUG    waiting for provisioned interfaces to appear...

```
look like clab_intfs calculation logic should take into account network-mode:none

```
sudo docker exec -it 001d7344ec31 bash
root@node0:/# env | grep INTF
CLAB_INTFS=1

```
pls also look at this check ethX, where X is >0**, there was an issue:

```
sudo containerlab deploy -t ~/clab/vsrx-cluster/vsrx-cluster.yaml
INFO[0000] Containerlab v0.50.0 started
INFO[0000] Parsing & checking topology file: vsrx-cluster.yaml
**Error: "node0" interface name "eth0" doesn't match the required pattern. It should be named as ethX, where X is >0**

```
yaml:
```
  links:
    - endpoints: ["node0:eth0", "switch:eth1"]
    - endpoints: ["node1:eth0", "switch:eth2"]

```
Thank you for your support!
Br,
D

hellt:
We now introduce a new env var CLAB_INTFS_WITH_MGMT that counts the management interface in when counting calculating the number of interfaces required by clab.

The reason for this change is to make sure that

1) we always check that the management interface is available, not only clab interfaces
2) we respect the `network mode: none` setting when eth0 interface is added manually to the list of endpoints and therefore it shouldn't be implied as an addition to datapath interfaces

Goes together with vrnetlab change https://github.com/hellt/vrnetlab/pull/166

** This is a breaking change for vrnetlab nodes**
Because newer vrnetlab (>=0.15.0) nodes will rely on the new CLAB_INTFS_WITH_MGMT env var, we will add an early error when this env var is not found asking users to upgrade containerlab.

Older vrnetlab nodes will continue to function because they would still use the CLAB_INTFS env var and since the calculation logic for CLAB_INTFS hasn't changed it will work as before.
It will not though support the `network mode: none`.

fix #1851

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -529,8 +529,19 @@ func (c *CLab) resolveBindPaths(binds []string, nodedir string) error {
 func (c *CLab) SetClabIntfsEnvVar() {
 	for _, n := range c.Nodes {
 		// Injecting the env var with expected number of links
+		// also adding one interface for the implicit management interface (eth0)
+		clabIntfs := len(n.GetEndpoints()) + 1
+
+		// when network mode is set to none, the eth0 is explicitly added
+		// so we need to substract it from the number of interfaces
+		// in vrnetlab the provisioned interfaces are matched as eth* glob
+		// so all dataplane + management interfaces are counted
+		if n.Config().NetworkMode == "none" {
+			clabIntfs -= 1
+		}
+
 		numIntfs := map[string]string{
-			types.CLAB_ENV_INTFS: fmt.Sprintf("%d", len(n.GetEndpoints())),
+			types.CLAB_ENV_INTFS: fmt.Sprintf("%d", clabIntfs),
 		}
 		n.Config().Env = utils.MergeStringMaps(n.Config().Env, numIntfs)
 	}
--- a/clab/config.go
+++ b/clab/config.go
@@ -524,24 +524,29 @@ func (c *CLab) resolveBindPaths(binds []string, nodedir string) error {
 	return nil
 }
 
-// SetClabIntfsEnvVar sets CLAB_INTFS env var for each node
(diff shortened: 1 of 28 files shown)


hellt:
not a prio right now

PR #1873, opened 2024-02-09
Docs: fix flag for graphviz generation

alexandrehassan:
The documentation used to say that running without `--srv` would generate the `.dot` file and not serve it. This behavior seems to have changed. 

Updated to follow current behavior (to get a the dot file `graph` must be called with `--dot`.

### DIFF
--- a/docs/cmd/graph.md
+++ b/docs/cmd/graph.md
@@ -73,11 +73,9 @@ When `graph` command is called with the `--mermaid` flag, containerlab will gene
 
 ### Online vs offline graphing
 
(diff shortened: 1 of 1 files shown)


alexandrehassan:
Reading through the graph documentation, the [offline section](https://containerlab.dev/cmd/graph/#online-vs-offline-graphing) is kind of confusing as it unclear that the default behavior is to act as if the `--offline` flag was set unless the lab is running. I think the documentation should read something like:

> If the lab is running containerlab will build the graph by inspecting the running containers which are part of the lab. This method provides additional details (like IP addresses). It is possible to opt out of this behavior by using the `--offline` flag.
>
>If `--offline` flag was not provided and no containers were found matching the lab name, containerlab will use the topo file only (as if offline mode was set).

hellt:
seems good @alexandrehassan 
would you mind adding this correction to this PR?

alexandrehassan:
@hellt Added the correction

PR #1889, opened 2024-02-14
Update install.md

ckuhtz:
the modified commands only work as root (can't chown otherwise) and this version should work as anyone:  Wouldn't it be easier to run the container as a specific user, namely whoever is the current user?  

Also replaced pwd fork for ${PWD}.

Suggestion for the future: the whole mkdir converted is really ugly IMHO; might be better to append a suffix to the file instead.


hellt:
good catch
`converted` is not needed anymore, I removed it entirely

PR #1890, opened 2024-02-15
Update install.md

ckuhtz:
pull isn't necessary as the command in the alias will automatically pull if not found locally.


hellt:
pull is used to ensure that the `dev-latest` tagged image is updated, as it changes with every commit to master.

ckuhtz:
@hellt you can accomplish that by specifying pull policy always with `—pull always` for the docker run command instead of two commands.

hellt:
good call
feel free to raise a PR with `--pull always` added to the run command if you want to add it
or I can add it a bit later

PR #1900, opened 2024-02-19
Check DOCKER-USER chain exists in nf_tables

steiler:
The simple check for the loaded nf_tables kernel module is not enough. It can be loaded whilst nf_tables is still not in use. iptables-legacy then most probably is.
So via this PR, we check if the DOCKER-USER chain exists. If not, we return and iptables will be used.

### DIFF
--- a/runtime/docker/firewall/nftables/client.go
+++ b/runtime/docker/firewall/nftables/client.go
@@ -50,7 +50,7 @@ func NewNftablesClient(bridgeName string) (*NftablesClient, error) {
 		return nil, err
 	}
 	if len(chains) == 0 {
-		log.Debug("nftables does not seem to be in use, no DOCKER-USER chain found.")
+		log.Debugf("nftables does not seem to be in use, no %s chain found.", definitions.DockerFWUserChain)
 		return nil, definitions.ErrNotAvailabel
 	}
 
--- a/runtime/docker/firewall/nftables/client.go
+++ b/runtime/docker/firewall/nftables/client.go
@@ -45,7 +45,7 @@ func NewNftablesClient(bridgeName string) (*NftablesClient, error) {
 		bridgeName: bridgeName,
 	}
 
-	chains, err := nftC.getChains("DOCKER-USER")
+	chains, err := nftC.getChains(definitions.DockerFWUserChain)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


aminato-nokia:
tested in RHEL 7 with this kernel 
[root@kvm-7 tmp]# uname -r
6.7.4-1.el7.elrepo.x86_64
[root@kvm-7 tmp]#

deployment was ok without any error messages

[root@kvm-7 tmp]# containerlab deploy --topo test_ne.yml
INFO[0000] Containerlab v0.49.0 started
INFO[0000] Parsing & checking topology file: test_ne.yml
INFO[0000] Creating lab directory: /tmp/clab-test_oob
INFO[0000] Creating container: "VSR-NRC"
INFO[0000] Adding containerlab host entries to /etc/hosts file
INFO[0000] Adding ssh config for containerlab nodes
+---+-----------------------+--------------+-------------------------------------------+---------+---------+----------------+--------------+
| # |         Name          | Container ID |                   Image                   |  Kind   |  State  |  IPv4 Address  | IPv6 Address |
+---+-----------------------+--------------+-------------------------------------------+---------+---------+----------------+--------------+
| 1 | clab-test_oob-VSR-NRC | 4bb896de89ea | registry.srlinux.dev/pub/vr-sros:23.10.R1 | vr-sros | running | 20.30.40.20/24 | N/A          |
+---+-----------------------+--------------+-------------------------------------------+---------+---------+----------------+--------------+
[root@kvm-7 tmp]#

steiler:
@hellt Take a look if you've got comments, otherwise we can merge.

runtime/docker/firewall/nftables/client.go
```diff
@@ -45,6 +45,15 @@ func NewNftablesClient(bridgeName string) (*NftablesClient, error) {
 		bridgeName: bridgeName,
 	}
 
+	chains, err := nftC.getChains(definitions.DockerFWUserChain)
+	if err != nil {
+		return nil, err
+	}
+	if len(chains) == 0 {
```
hellt, line 52: Can we check for docker user chain and not just some chains?

steiler: Thats what we do. if you look closely you will notice `definitions.DockerFWUserChain` as the parameter to `getChains(...)`.

PR #1906, opened 2024-02-22
Library Usage: Allow to use containerlab as a library

steiler:
To allow using Containerlab as a go library we need to adjust some things, that went wrong in the past.

- Introduce clear seperation of cmd and clab package
- Build a proper facade that provides the high-level calls.


steiler:
What would we consider the right abstraction for topology handover?
Would we provide a file pointer or would we take in a json blob (probably with an additional argument, the working dir, where we would create the lab folder).
My tendency is towards the json blob.

ldelossa:
As per our discussion here: https://github.com/srl-labs/containerlab/discussions/1905

I want to outline the (admittedly few) design requirements Cilium would have for utilizing ContainerLab as an embedded library. 

I'd like to first mention that our initial goal to using ContainerLab is for embedding test case scenarios. Cilium has quite a few complex testing scenarios. Using ContainerLab would allow us to rely on an already mature solution which can setup network topologies for dedicated Cilium datapath testing. We are tracking this issue here: https://github.com/cilium/cilium/issues/27450 

Luckily, our requirements are not too crazy (IMO, you may think otherwise with your mental map of ContainerLab :))  

1. We'd like to define a topology in a JSON file which define a set of network namespaces and connects them via veths (as I understand, this is the core of ContainerLab, we don't really have a scoped focus on other type of network connectivity such as bridge into VMs). 
2. We will need returned to us a list of configured network namespaces and a handle to these namespaces. Whether this is a pinned filesystem path or a FD can be up to the ease of implementation on ContainerLab's side. We just need the handles provided back to us. 
3. I'm discussing this a bit internally, but I also assume we need to understand the created veth devices (netns + ifindex) after the topology is created, since we'd want to apply eBPF programs to these interfaces to facilitate datapath testing. 


> What would we consider the right abstraction for topology handover? 

Can you explain a bit more about what you mean by "topology handover". 
If this is "communicating the resulting topology back to the caller" then I think points 2. and 3. above cover this from our requirements. 

>  Would we provide a file pointer or would we take in a json blob (probably with an additional argument, the working dir, where we would create the lab folder).

It would be possible for the library to take in a generic io.Reader, no? Therefore, the caller can provide any medium (file, in-memory, network-socket, etc...) which implements the io.Reader interface. The implications of this is that the one doing the reader must also do the topology syntax checking at read time.

steiler:
> 2. We will need returned to us a list of configured network namespaces and a handle to these namespaces. Whether this is a pinned filesystem path or a FD can be up to the ease of implementation on ContainerLab's side. We just need the handles provided back to us.

What we do is, that we always create the `/var/run/netns/<container_name>` handle, as this would also be required by the `ip` command. This _contract_ should suffice getting a hold onto the handle. 
(As a sidenote, we already provide json output for most of the clab commands, so you could already get the container names status etc in json as of todays binary)

> 3. I'm discussing this a bit internally, but I also assume we need to understand the created veth devices (netns + ifindex) after the topology is created, since we'd want to apply eBPF programs to these interfaces to facilitate datapath testing.

Would it be sufficient for you to get the netns handle and query the interfaces yourself to figure out the ifindex's. Otherwise we could probably provide these in the results.

> > What would we consider the right abstraction for topology handover?
> 
> Can you explain a bit more about what you mean by "topology handover". If this is "communicating the resulting topology back to the caller" then I think points 2. and 3. above cover this from our requirements.

Like what you answered below, provide a file or an io.reader.

> > Would we provide a file pointer or would we take in a json blob (probably with an additional argument, the working dir, where we would create the lab folder).
> 
> It would be possible for the library to take in a generic io.Reader, no? Therefore, the caller can provide any medium (file, in-memory, network-socket, etc...) which implements the io.Reader interface. The implications of this is that the one doing the reading must also do the topology syntax checking at read time.

Yes thats my thinking, you provide an io.Reader seems to be a good approach.
Well for the checking, if we get an io.Reader, we'd hand that over to the yaml library to read the data and hand back an error if that parsing fails. So I think that should be fine. I'm realizing just now that there should be other means to utilize the library function. One would be to export the topology struct, which would allow the topology definition to be created via structs, but we convert that internally to the specific nodes that implement the nodes.Node interface. So this level of abstraction should probably also be exposed as a subsequent step.

steiler:
@ldelossa Another question: Is your plan to use containerlab also as a means to deploy the container / network namespaces or is it mainly your objective to use containerlab for wiring?

ldelossa:
> What we do is, that we always create the /var/run/netns/<container_name> handle, as this would also be required by the ip command. This contract should suffice getting a hold onto the handle.
(As a sidenote, we already provide json output for most of the clab commands, so you could already get the container names status etc in json as of todays binary)

This sounds completely reasonable to me. As long as the naming is consistent with the container names in the topology file, the user of the library can do the correlating. 

> Would it be sufficient for you to get the netns handle and query the interfaces yourself to figure out the ifindex's. Otherwise we could probably provide these in the results.

When you say "query the interfaces" what do you mean here exactly? 
My concern here is that say we have a very busy network namespace with 4 veths created. 
What is a quick and simple way for us to understand which veth (ifindex/name) goes to which adjacent network? 
However if you mean "query the interfaces" as in, Go interfaces the library will provide, then this seems feasible, just not aware of these interfaces. 

> Yes thats my thinking, you provide an io.Reader seems to be a good approach.
Well for the checking, if we get an io.Reader, we'd hand that over to the yaml library to read the data and hand back an error if that parsing fails. So I think that should be fine. I'm realizing just now that there should be other means to utilize the library function. One would be to export the topology struct, which would allow the topology definition to be created via structs, but we convert that internally to the specific nodes that implement the nodes.Node interface. So this level of abstraction should probably also be exposed as a subsequent step.

100% to this. Especially if you want to modify a topology in Golang, having to go "back-through" the JSON would be inconvenient :). 

> Another question: Is your plan to use containerlab also as a means to deploy the container / network namespaces or is it mainly your objective to use containerlab for wiring?

I'm not sure I'm seeing a difference between "deploying" and "wiring". 
But for our use cases we would want the network names spaces, veths wired up, routing tables configured, etc.. I'm leaning towards "deployed" here, we expect a full end to end network topology setup. From there we will apply eBPF programs to specific interfaces and then run various unit tests across these network namespaces/eBPF programs.

clab/deploy_options.go
```diff
@@ -0,0 +1,79 @@
...
+	skipPostDeploy bool
+	graph          bool
+	maxWorkers     uint
+	exportTemplate string
+}
+
+func NewDeployOptions(maxWorkers uint) (*DeployOptions, error) {
+	d := &DeployOptions{
+		maxWorkers: math.MaxUint,
+	}
```
hellt, line 21: seems like this is not needed
setting maxworkers to the max value is not going to be useful in any case since it is reset in the initWorkerCount

clab/deploy_options.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/tklauser/numcpus"
+)
+
+type DeployOptions struct {
+	reconfigure    bool
+	skipPostDeploy bool
+	graph          bool
+	maxWorkers     uint
+	exportTemplate string
+}
```
hellt, line 16: since Setters are implemented this struct should be unexported.
To prevent users from using it directly.

If direct usage is allowed, then setters are not required

cmd/disableTxOffload.go
```diff
@@ -38,36 +40,17 @@ var disableTxOffloadCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		ctx, cancel := context.WithCancel(context.Background())
-		defer cancel()
-
-		log.Infof("getting container '%s' information", cntName)
 
-		nodeRuntime := c.GlobalRuntime()
+		node, err := c.GetNode("cntName")
```
hellt, line 44: should it be a var, not a string?

cmd/deploy.go
```diff
@@ -107,234 +102,50 @@ func deployFn(_ *cobra.Command, _ []string) error {
...
-		return err
+	if mgmtNetName != "" {
+		opts = append(opts, clab.WithManagementNetworkName(mgmtNetName))
 	}
-
-	c.SSHPubKeys, err = c.RetrieveSSHPubKeys()
-	if err != nil {
-		log.Warn(err)
+	if v4 := mgmtIPv4Subnet.String(); v4 != "<nil>" {
+		opts = append(opts, clab.WithManagementIpv4Subnet(v4))
```
hellt, line 113: maybe there is a better way to check if net.IP is nil?

clab/file.go
```diff
@@ -27,9 +27,9 @@ const (
 	varFileSuffix = "_vars"
 )
 
-// GetTopology parses the topology file into c.Conf structure
+// LoadTopology parses the topology file into c.Conf structure
 // as well as populates the TopoFile structure with the topology file related information.
-func (c *CLab) GetTopology(topo, varsFile string) error {
+func (c *CLab) LoadTopology(topo, varsFile string) error {
```
hellt, line 32: let's name it `LoadTopologyFromFile` to indicate that we work on the file input here

clab/deploy_options.go
steiler: Wait, the DeployOptions must be exported, because it will be handed in from externally.
The settings within are not, they can only be set via the setters... So in my view this is kinda right... not?!

clab/deploy_options.go
steiler: Done

cmd/disableTxOffload.go
steiler: Correct. Done.

cmd/deploy.go
steiler: It's been like this for ages (https://github.com/srl-labs/containerlab/blob/main/cmd/deploy.go#L367-L372). We can check if there is another way, but lets do that seperately.

clab/file.go
steiler: Done

clab/deploy_options.go
hellt: yeah, I didn't notice unexported fields at first

PR #1907, opened 2024-02-22
Fixing Exec race condition

steiler:


tests/01-smoke/stages.clab.yml
```diff
@@ -7,11 +7,11 @@ name: stages
 topology:
   defaults:
     kind: linux
-    image: alpine:3
+    image: ubuntu:latest
   nodes:
     node1:
       exec:
-        - ash -c "date +%s%3N > /tmp/time"
+        - ash -c "date +%s%N > /tmp/time"
```
hellt, line 14: let's use `debian:bookworm-slim` for now

and pre-pull it so that it is not pulled by clab itself to not have effect on container startup time

and exec would read `bash` instead of `ash`

PR #1908, opened 2024-02-22
Replace utils.LinkByNameOrAlias with netlink.LinkByName

steiler:
The netlink library now supports AltNames/Aliases so we use that feature.


steiler:
The netlink library has an issue with querying altnames that exceed the 15 char length.
Refere to: https://github.com/vishvananda/netlink/issues/955
This is blocking us from completing this. Hence we need to wait for a fix.

hellt:
This is sweet! Bonus points for making it into netlink 💪

PR #1912, opened 2024-02-23
k8s-kind: Expose kind command line option via extras field

YutaroHayakawa:
When the default CNI is disabled, the k8s nodes won't become ready forever until CNI is installed. However, the containerlab waits for the cluster readiness which will never succeeds. To solve this issue, don't wait for the cluster readiness and let deploy command finish, so that we can install the CNI after that.

We could conditionally wait for the cluster readiness by checking cluster configuration (e.g., wait only when disableDefaultCNI=false), but it can be a bit confusing from user's standpoint and if users wish to wait, they can always do it by themselves, for example with kubectl wait command.

### DIFF
--- a/docs/manual/kinds/k8s-kind.md
+++ b/docs/manual/kinds/k8s-kind.md
@@ -100,6 +100,22 @@ With `k8s-kind`` nodes it is possible to use the following configuration paramet
 - [image](../nodes.md#image) - to define the kind container image to use for the kind cluster
 - [startup-config](../nodes.md#startup-config) - to provide a kind cluster configuration (optional, kind defaults apply otherwise)
 
+### Extra parameters
+
+In addition to the generic node parameters, `k8s-kind` can take following extra parameters from `extras` field.
+
+```
+topology:
+  nodes:
+    kind0:
+      kind: k8s_kind
+      extras:
+        k8s_kind:
+          deploy:
+            # Corresponds to --wait option. Wait given duration until the cluster becomes ready.
+            wait: 0s
+```
+
 ## Known issues
 
 ### Duplication of nodes in the output
--- a/nodes/k8s_kind/k8s_kind.go
+++ b/nodes/k8s_kind/k8s_kind.go
@@ -77,8 +77,25 @@ func (n *k8s_kind) Deploy(_ context.Context, _ *nodes.DeployParams) error {
 		// set the byteConfig as the config to use
 		cluster.CreateWithV1Alpha4Config(conf),
 		// make the Create call synchronous, but use a timeout of 15 min.
+		// This may be overridden by the user in the extra config.
 		cluster.CreateWithWaitForReady(time.Duration(15)*time.Minute),
 	)
+
+	// Handle extra deploy options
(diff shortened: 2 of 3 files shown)


hellt:
@YutaroHayakawa 

I am hesitant to say that majority of the kind users would install a custom CNI, more meandering to say that most just need a regular cluster with most of the kind defaults (could be wrong).

So maybe it is better to be more explicit here and use the `type` option under the node level that could take two values:

* blocking (default)
* non-blocking

non-blocking would mean the cluster deployment stage won't block waiting for cluster to be rendered ready.

The topology would look like

```yaml
  nodes:
    k01:
      kind: k8s-kind
      startup-config: k01-config.yaml
      type: non-blocking
```

/cc @steiler

YutaroHayakawa:
> I am hesitant to say that majority of the kind users would install a custom CNI, more meandering to say that most just need a regular cluster with most of the kind defaults (could be wrong).

Fair enough. Kind document also says `This is a power user feature with limited support`.

Regarding the solution, I like the direction to make a toggle, but using `type` field sounds a bit hacky. Is it really intended to be used in this kind of case? Also, it only allows having a single string value. We may want to introduce more knobs in the future. For example, `kind create cluster` command can take the flag like below, but we cannot specify them in ContainerLab as of today (`--wait` is one of them). In particular, I'm interested in exposing the `--retain` and `--verbosity` options, which are, again, niche use cases tho.

```
Creates a local Kubernetes cluster using Docker container 'nodes'

Usage:
  kind create cluster [flags]

Flags:
      --config string       path to a kind config file
  -h, --help                help for cluster
      --image string        node docker image to use for booting the cluster
      --kubeconfig string   sets kubeconfig path instead of $KUBECONFIG or $HOME/.kube/config
  -n, --name string         cluster name, overrides KIND_CLUSTER_NAME, config (default kind)
      --retain              retain nodes for debugging when cluster creation fails
      --wait duration       wait for control plane node to be ready (default 0s)

Global Flags:
      --loglevel string   DEPRECATED: see -v instead
  -q, --quiet             silence all stderr output
  -v, --verbosity int32   info log verbosity, higher value produces more output
```

So, instead of using `type,` how about introducing a new generic opaque field to expose node-type-specific options? Something like:

```
  nodes:
    k01:
      kind: k8s-kind
      startup-config: k01-config.yaml
      options:
        wait: 0s
        retain: true
        verbosity: 5
```

I'm happy to help with implementation.

steiler:
I'm sharing your general opinion.
I am not sure if the type field is the most future proof variant for this switch. But I'm also lacking a better solution atm.
So well lets go with this approach for now.

hellt:
We have the `Extras` knob under each node definition https://github.com/srl-labs/containerlab/blob/2d58102bcdf5b20a8da7db3aafb25664bb728873/types/node_definition.go#L68 that could be used to define kind-specific options.

We may leverage it, as it was already used for a similar case with a few other nodes

YutaroHayakawa:
> We have the Extras knob under each node definition

Ah, cool. Yeah, then I'll fix this PR to leverage it.

YutaroHayakawa:
Done

PR #1914, opened 2024-02-23
veth creation: make first node create the veth endpoints

steiler:
This is the PR to tackle the link creation issues described in #1907.

This PR now changes the link creation strategy. For the veth, where there are two endpoints located in two network namespaces. The link is created by the first node that reaches the create-links phases. It creates the link with its two endpoints in the host namespace with randomized interface names. One endpoint (netlink.link) is then transferred to the netns of the container, renamed to the designated name, and set admin up.
The other endpoint remains in the root/host netns (wiht its random clab-<randid> name) until the second node reaches the create-links phase. The second node, instead of creating the whole veth link, only pulls in the existing endpoint, renames it and sets it admin up.

This way, the defined interfaces will exist, as soon as the create-links phase is reached. However, it is not guarantied that the link is already operational. But anyways will this help to perform `ip link` and `ip address` commands and further also prevents deadlocks in case of vrnetlab nodes, where the nodes would only boot the VMs when all the interfaces are available to the network namespace.

Further did I remove the reference from nodes to links. Nodes do own endpoints that belong to links. Hence to access Link attributes, the endpoint should be consulted. This prevents us from drifting state, if an endpoint is added but the link is missed or the other way around.

### DIFF
--- a/links/endpoint_veth.go
+++ b/links/endpoint_veth.go
@@ -20,3 +20,7 @@ func (e *EndpointVeth) Verify(_ *VerifyLinkParams) error {
 func (e *EndpointVeth) Deploy(ctx context.Context) error {
 	return e.GetLink().Deploy(ctx, e)
 }
+
+func (e *EndpointVeth) AutoDeployWithAEnd() bool {
+	return false
+}
--- a/links/endpoint_veth.go
+++ b/links/endpoint_veth.go
@@ -1,5 +1,7 @@
 package links
 
+import "context"
+
 type EndpointVeth struct {
 	EndpointGeneric
 }
@@ -13,4 +15,8 @@ func NewEndpointVeth(eg *EndpointGeneric) *EndpointVeth {
 // Verify verifies the veth based deployment pre-conditions.
 func (e *EndpointVeth) Verify(_ *VerifyLinkParams) error {
 	return CheckEndpointUniqueness(e)
+}
+
+func (e *EndpointVeth) Deploy(ctx context.Context) error {
+	return e.GetLink().Deploy(ctx, e)
 }
--- a/links/endpoint_veth.go
+++ b/links/endpoint_veth.go
@@ -21,6 +21,6 @@ func (e *EndpointVeth) Deploy(ctx context.Context) error {
 	return e.GetLink().Deploy(ctx, e)
 }
 
-func (e *EndpointVeth) AutoDeployWithAEnd() bool {
+func (e *EndpointVeth) IsNodeless() bool {
 	return false
 }
--- a/links/link_veth.go
+++ b/links/link_veth.go
@@ -232,7 +232,7 @@ func (l *LinkVEth) Deploy(ctx context.Context, ep Endpoint) error {
 	// The first node to trigger the link creation will call deployAEnd,
 	// subsequent (the second) call will end up in deployBEnd.
 	switch l.DeploymentState {
-	case LinkDeploymentStateFullDeployed:
+	case LinkDeploymentStateHalfDeployed:
 		return l.deployBEnd(ctx, idx)
 	default:
 		return l.deployAEnd(ctx, idx)
--- a/links/link_veth.go
+++ b/links/link_veth.go
@@ -3,10 +3,10 @@ package links
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
-	"github.com/srl-labs/containerlab/nodes/state"
 	"github.com/srl-labs/containerlab/utils"
 	"github.com/vishvananda/netlink"
 )
@@ -57,8 +57,6 @@ func (r *LinkVEthRaw) Resolve(params *ResolveParams) (Link, error) {
 		}
 		// add endpoint to the link endpoints
 		l.Endpoints = append(l.Endpoints, ep)
-		// add link to endpoint node
-		ep.GetNode().AddLink(l)
 	}
 
 	// set default link mtu if MTU is unset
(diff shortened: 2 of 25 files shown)


robotwalk:
Hi @steiler,

I just tested this PR and it works like a charm. 

 - Boot up duration increased.
 - Most resource hungry vms get healthy first

Topology needed before around 48min to become healthy and there was always the possibilty that the virtual asr9k will never become healthy because there are not enough resources. 
With this PR asr9k become healthy after 17min and the complete topology after 41mins.

If I did the math right speed increased by 15%

links/endpoint_veth.go
```diff
@@ -14,3 +16,11 @@ func NewEndpointVeth(eg *EndpointGeneric) *EndpointVeth {
 func (e *EndpointVeth) Verify(_ *VerifyLinkParams) error {
 	return CheckEndpointUniqueness(e)
 }
+
+func (e *EndpointVeth) Deploy(ctx context.Context) error {
+	return e.GetLink().Deploy(ctx, e)
+}
+
```
hellt, line 23: this seems weird... We have (endpoint).Deploy(), which in its turn calls (link).Deploy.

Why don't we remove all (endpoint).Deploy and instead call (link).Deploy directly?
The link interface has `GetEndpoints` to access associated endpoints

This is the main thing that makes me suspicious here

hellt: to expand on this.
Logically we have the following composition of objects:

* Link
  * EndpointA
  * EndpointA

Thus it makes sense that the calling node has `(node).DeployLink` (this func still exists) - https://github.com/srl-labs/containerlab/pull/1914/files#diff-d25cfdb9022b8a280a3cf87aa5a83926d393e4d5d28c550dd57cca87c9130359R519

What I propose to reconsider is what happens next. I think the default node should be 

```
for _, l := range d.Links {
		err := l.Deploy(ctx)
```

since we deploy a link object. Now how and what is needed to deploy this link is with the link implementation. If a link needs to deploy endpoints separately -- it should happen in the `(link).Deploy()`.

That way we keep the logical wrapping intact and

steiler: > @steiler this seems weird... We have (endpoint).Deploy(), which in its turn calls (link).Deploy.
> 
> Why don't we remove all (endpoint).Deploy and instead call (link).Deploy directly? The link interface has `GetEndpoints` to access associated endpoints
> 
> This is the main thing that makes me suspicious here

The abstraction of having a link attached to a node simply is wrong. The Endpoint belongs to a Node. Regarding the call to Link.Deploy() you see we provide the endpoint as parameter. This is, because it matters which side (Endpoint) of the Link calls the creation. Sure we could search through the list via provided node names bla bla bla ... still it would be the wrong abstraction of the whole thing.

steiler: > to expand on this. Logically we have the following composition of objects:
> 
> * Link
>   
>   * EndpointA
>   * EndpointA
> 
> Thus it makes sense that the calling node has `(node).DeployLink` (this func still exists) - https://github.com/srl-labs/containerlab/pull/1914/files#diff-d25cfdb9022b8a280a3cf87aa5a83926d393e4d5d28c550dd57cca87c9130359R519
> 
> What I propose to reconsider is what happens next. I think the default node should be
> 
> ```
> for _, l := range d.Links {
> 		err := l.Deploy(ctx)
> ```
> 
> since we deploy a link object. Now how and what is needed to deploy this link is with the link implementation. If a link needs to deploy endpoints separately -- it should happen in the `(link).Deploy()`.
> 
> That way we keep the logical wrapping intact and

Again, the node itself needs to specifically know about its side of the link, namely the Endpoint. It contains the name that a link is referenced with etc. So the endpoint belongs to the Node and the Endpoint is part of a Link. Just like that.
We've had these discussions last time and I surrendered, but figured again this time that it was bullshit.
What we did so far was to keep a list of links and endpoints. Which broke my neck because I was not keeping these lists in sync... avoidable by adding just endpoints.
If you where just to store links, whenever you would want to access the Endpoint via the link, you would need to resolve it via a loop, looking up the node name, which is stupid.
So I remain with my opinion that this is the right abstraction for the whole matter.

steiler: > Thus it makes sense that the calling node has `(node).DeployLink` (this func still exists) - https://github.com/srl-labs/containerlab/pull/1914/files#diff-d25cfdb9022b8a280a3cf87aa5a83926d393e4d5d28c550dd57cca87c9130359R519

Thanks for spotting this, I'll rename it to (node).EndpointsDeploy()

links/link_veth.go
```diff
@@ -109,70 +107,145 @@ func (*LinkVEth) GetType() LinkType {
...
+	// first we need to check that the provided ep is part of this link
+	_, err := l.getEndpointIndex(ep)
+	if err != nil {
+		return err
+	}
+
+	// The first node to trigger the link creation will call deployAEnd,
+	// subsequent (the second) call will end up in deployBEnd.
+	switch l.DeploymentState {
+	case LinkDeploymentStateDeployed:
```
hellt, line 242: this variable name seems misleading. At this point in time we haven't deployed the link, but just the A end of it.

Feels like this should be `LinkDeploymentStateAEndDeployed`
then once the B End is deployd the link reaches `LinkDeploymentStateDeployed`?

links/link_veth.go
```diff
@@ -109,70 +107,145 @@ func (*LinkVEth) GetType() LinkType {
...
+	// since each node calls deploy on its links, we need to make sure that we only deploy
+	// the link once, even if multiple nodes call deploy on the same link.
+	l.deployMutex.Lock()
+	defer l.deployMutex.Unlock()
+
+	// first we need to check that the provided ep is part of this link
+	_, err := l.getEndpointIndex(ep)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
steiler: fair point, but then lets only provide the index...

links/link_veth.go
steiler: Yes you are right... the state was there as is and I lazily missused it.

links/link_veth.go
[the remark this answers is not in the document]
steiler: Done

links/link_veth.go
steiler: Done

hellt:
Thanks mate, this is solid
Shipping it

PR #1941, opened 2024-03-14
Added long names of vr-sros and srl

Bzhash:


steiler:
We should incorporate the NodeRegistry here. It knows all the names a certain kind has assigned.

hellt:
the whole config thingy needs to be either scraped or reworked... For now if that can unblock someone I am willing to merge it.

PR #1960, opened 2024-03-27
Add vr-c8000v kind

mzagozen:
This is right now a verbatim copy of the existing `vr-csr` kind. I expect they will diverge in the future though, so it is best to create a copy now rather than encourage branching logic in the same file.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -11,7 +11,7 @@ on:
   workflow_dispatch:
 
 env:
-  GO_VER: "1.20.2"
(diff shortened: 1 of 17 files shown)


hellt:
Thanks, @mzagozen I am merging this as I successfully booted it using the tip of my branch. We can enhance the support for c8000v in the future.
Primarily I think it would be interesting to see if there are performance gains when using `-cpu host` qemu flags and giving it more vcpu `-smp 2` as compared to the default `1vcpu/4GB`

PR #1962, opened 2024-03-28
Add "auto" option to ipv4-subnet and ipv6-subnet

mzagozen:
If we do not care about the assigned IPv4/IPv6 subnet for the container network, we let the container runtime allocate one for us. This guarantees to avoid a collision of the hardcoded range (172.20.20.0/24 and 2001:172:20:20::/64) with any existing container networks. For example this makes it possible to run concurrent isolated topologies without directly modifying the topology files.

This just works for the podman runtime. For the docker runtime we compute the IPv6 subnet manually because it can only pick addresses from the configured (global) subnet.


hellt:
thanks @mzagozen 
I have rebased this PR since in the meantime we switched to go1.21.
I also added the docs/schema and slightly opimized the v6 ULA code.

Let me know if that seems OK to you and we can merge

mzagozen:
Agree on the docs and refactoring. I pushed another rebased branch to fix the diff in the GitHub UI.

PR #1990, opened 2024-04-12
Support for Border0 Connector / Docker Exec

adrianosela:
## Support for Border0 Connector / Docker Exec

This PR introduces the ability to deploy the [Border0 Connector](https://docs.border0.com/docs/border0-connector#connectors) as a node in a containerlab environment (lab)... connecting all nodes without the need to expose nodes' SSH ports or manage/provision any nodes with secrets.

This makes for a seamless experience for containerlab users who wish to expose their labs over the internet (behind their authentication method of choice: Google/GitHub/Microsoft/Email-Link).

The nicest thing about it is users can talk to their nodes from any web browser using our [web client](https://www.border0.com/blogs/introducing-the-border0-client-portal). 

### Video Demo

https://github.com/srl-labs/containerlab/assets/16856511/13fd9f2f-3c2a-469b-a64b-0111997918df

---

I alluded to this feature being in the works in  https://github.com/srl-labs/containerlab/pull/1768:

> @hellt Let's talk next year. We maaaaay (no promises) have something in the works to simplify the integration and make the experience much nicer for clab users.

We are excited to hear your thoughts and collaborate on making the containerlab x border0 experience as best as it can be!

cc:// @steiler @hellt @atoonk

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.21
 require (
 	github.com/a8m/envsubst v1.4.2
 	github.com/awalterschulze/gographviz v2.0.3+incompatible
+	github.com/borderzero/border0-go v1.4.29
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/containernetworking/plugins v1.4.1
 	github.com/containers/common v0.58.1
@@ -61,6 +62,7 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2 // indirect
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.15.5 // indirect
 	github.com/blang/semver/v4 v4.0.0 // indirect
+	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cilium/ebpf v0.12.3 // indirect
 	github.com/cloudflare/circl v1.3.7 // indirect
 	github.com/containerd/cgroups/v3 v3.0.3 // indirect
@@ -208,7 +210,7 @@ require (
 	github.com/goombaio/namegenerator v0.0.0-20181006234301-989e774b106e // indirect
 	github.com/gorilla/mux v1.8.1 // indirect
 	github.com/gorilla/schema v1.2.1 // indirect
-	github.com/gosimple/slug v1.12.0 // indirect
+	github.com/gosimple/slug v1.12.0
 	github.com/gosimple/unidecode v1.0.1 // indirect
 	github.com/hairyhenderson/go-fsimpl v0.0.0-20220529183339-9deae3e35047 // indirect
 	github.com/hairyhenderson/toml v0.4.2-0.20210923231440-40456b8e66cf // indirect
--- a/go.mod
+++ b/go.mod
@@ -5,8 +5,7 @@ go 1.21
 require (
 	github.com/a8m/envsubst v1.4.2
 	github.com/awalterschulze/gographviz v2.0.3+incompatible
-	github.com/borderzero/border0-go v1.4.29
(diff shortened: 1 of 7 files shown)


hellt:
thanks @adrianosela 
this is cool. Will have to try it out somewhen in the next few weeks

go.mod
```diff
@@ -1,10 +1,13 @@
 module github.com/srl-labs/containerlab
 
-go 1.21
+go 1.22
+
+toolchain go1.22.2
```
hellt, line 5: @adrianosela it looks like `github.com/borderzero/border0-go` requires go1.22.
Is it mandatory in your case?
In containerlab we use `goN-1` version to allow for broader adoption.

Maybe go1.22 is not really needed by border0-go and you can declare go1.21

adrianosela: Downgraded `github.com/borderzero/border0-go` to `go1.21` 😄

adrianosela:
Hey @hellt - just FYI I pushed another commit that uses the Border0 authentication flow from our SDK, which maintains backwards compatibility with the legacy (email + password) flow.

I mostly did this hoping that removing code from here would make codecov's diff happy :D But that did not happen.

What are your contributing guidelines? Do we need to satisfy test coverage here since we are just calling out to our SDK? (which has plenty of tests there).

hellt:
No worries @adrianosela, it is fine to have gaps in test coverage for the parts related to b0 integration.

So for my understanding, is email/pass-based auth is completely legacy and everything can be done with the new flow? If yes, then I am in favor of removing the email/pass code parts completely

adrianosela:
@hellt

With respect to the email + password flow: the email + password (programmatic) authentication IS formally deprecated. However, at the time of deprecating such logins, we opted to allowlist a handful of users who were using the feature to continue to use it in order to prioritize speed of rolling out the new (more secure) authentication flow. Some of these users (at least one of which is definitely a containerlab user) are still active today. We should figure out together if we want to:
- remove support for the legacy auth flow from this repo (and optionally reach out to them to let them know)
- simply keep it in here since its hidden from the command tree and docs + new users cannot use it anyways.

---

I'd be happy to strip out a lot of code from here, but I think we should have a live conversation after you've given the new [Border0 Connector](https://docs.border0.com/docs/border0-connector#connectors) based flow a try and we'd figured out together exactly how you'd/we'd like containerlab users to manage their policies and sockets with the new flow.

It seems historically in containerlab you would define Border0 policies and sockets in the config yaml directly. We should figure out if you want to keep supporting that and if so we'd have to make some adjustments to the code here. With the connector model (as in the docs linked above) everything is remotely managed, instead of static-config managed. However we could add a border0 command here to create and link the resources in static config to a connector anyways.

Let's follow on our ongoing email thread until we have clarity. cc:// @atoonk

hellt:
@adrianosela after playing with this a little bit I wonder why do we want to tie this in to containerlab via the tools command?
This seems to be a generic connector of type "docker host" that makes a lot of sense, but it got me wondering if this should be coupled to containerlab at all?

What I mean is I think it would make a lot of sense if `clab tools border0 setup` emit a `docker run` command instead of a yaml blob. Maybe even start a container using the docker client that is part of the clab already.

I might be missing something here, but I think a single b0 container providing a connector is really all we need. It is easier to setup and requires fewer steps.

adrianosela:
Hey @hellt - that sure sounds more reasonable! Since the B0 container isn't really part of the lab's topology (and if a customer wants to do that they can still do it themselves).

Our internal priorities have shifted quite a bit. I don't anticipate any work here in the next ~2 weeks. But we'll get back to it soon enough.

hellt:
we will get back to it later

PR #1997, opened 2024-04-15
add cookie consent and option to manage cookies

ankudinov:
Containerlab MkDocs is using Google Analytics, which is not necessarily GDPR compliant. While this goes against engineering common sense and is very theoretical concern - you never know if it will hit or not. Legal world is weird. I'd suggest at least adding an explicit cookie consent and an option to change cookie settings as it's documented in [MkDocs privacy section](https://squidfunk.github.io/mkdocs-material/setup/ensuring-data-privacy/).  
GDPR and Google Analytics is a long read that can certainly not be covered in a single PR.
To add more fun, German court decided Jan 2022, that Google Fonts are not GDPR compliant as well. There is a MkDocs knob for that (privacy plugin): https://github.com/squidfunk/mkdocs-material/issues/739
However I'll not add it as part of this PR as it looks a bit theoretical to me and every MkDocs site is using gFonts anyway.  
I believe that Containerlab community is sane and have nothing to worry about, but as it's also quite popular - some basic guards may be helpful.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -262,7 +262,7 @@ extra:
       - accept
       - manage
       - reject
-    description: >- 
+    description: >-
       We use cookies to recognize your repeated visits and preferences, as well
       as to measure the effectiveness of our documentation and whether users
       find what they're searching for. With your consent, you're helping us to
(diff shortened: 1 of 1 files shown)


ankudinov:
I hope, I can also win the most weird cLab PR crown by opening that. 😃

hellt:
I wonder why squidfunk doesn't have this on https://squidfunk.github.io/mkdocs-material/

But yeah, it is better to be safe

Issue #2013, opened 2024-04-24
cRPD 23.1+ version specific changes


mzagozen:
I made two changes in the pre-deploy steps for cRPD version 23.1 and newer:
1. The license file is no longer expected in `/config/license/safenet/junos_sfnt.lic`, but rather in the parent directory. So it becomes just `/config/license/junos_sfnt.lic`.
2. It is no longer required to edit the `/etc/sshd_config` directly and restart SSHD. The startup config for enabling SSH server and allowing *root* login are honored when the control plane starts.

Implementing this is straightforward, but I am having trouble coming up with a way of keeping backwards compatibility by conditioning behavior on the version. The current cRPD version is available the container image in `/etc/junos/release.txt` as a string on a single line. But reading this file requires a running container, so not available yet in the pre-deploy stage. Starting a (separate) container just to read the file and grab the would work, but feels very hacky ... WDYT?

hellt:
Regarding the license, it seems it should be possible to use the cli command (https://github.com/srl-labs/containerlab/issues/1350) to load one.
Maybe this is version-agnostic enough to work with pre 23 and post 23 versions?

hellt:
done in #2014

PR #2014, opened 2024-04-24
Always use CLI command to load cRPD license

### LINKED ISSUE #1350: Change license provisioning for crpd
Crpd v22+ doesn't load a license automatically from the well known path. 
To overcome that we should mount a license to /tmp/lic.key and execute "exec cli request system license add /tmp/filename"

mzagozen:
The expected license file location was changed in Junos 22 and later. To keep the code backwards and forwards compatible we use the CLI command to add a license file from a stable location.

Closes #1350

### DIFF
--- a/nodes/crpd/crpd.go
+++ b/nodes/crpd/crpd.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/srl-labs/containerlab/clab/exec"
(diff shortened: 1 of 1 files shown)


hellt:
thanks @mzagozen 
this lgtm
I presume you tried it with the crpd 23+? I think in general containerlab strives to support the latest releases, to go in a lockstep with the current NOS offerings; with that in mind I think what is important is that newer version runs smoothly, and older ones might be less prioritized in terms of the support

mzagozen:
Right, I only have 23.4R1.9 available for testing. I remember using the CLI command way back anyway :)
Like I mentioned in #2013, the SSH config massaging code is now unnecessary, But I guess since it does not interfere with the current latest it may as well remain in place for older versions.

PR #2028, opened 2024-04-29
Add knob to skip adding extended ACL to labdir

steiler:
This PR adds the "--skip-labdir-acl" knob to the deploy call. Setting it will skip adding extended ACL to the lab dir.


steiler:
@hellt can you maybe add a quick test. THX

hellt:
thanks for the quick turnover @steiler!
I tested locally and it works as advertised. didn't add tests though, just added the flag to one of the deployments so that we know it doesn't break things

Issue #2033, opened 2024-05-04
Issue with Arista version in quickstart topology


james-m-henderson:
I went through the [quickstart](https://containerlab.dev/quickstart) guide and have noted a couple of issues.

The most significant is that the suggested version of Arista (4.25.0F) is not present on the Arista website for download. The closest I could find was 4.25.11M.

After downloading this version and running, containerlab seemed to hang indefinitely after doing `containerlab deploy`.

I noticed that while the docker VMs were up, if I try to use the Cli command on Arista, it doesn't work:

```
jmh@the-pot:~/clab-quickstart$ docker exec -it clab-srlceos01-ceos bash
bash-4.2# exit
jmh@the-pot:~/clab-quickstart$ docker exec -it clab-srlceos01-ceos Cli
OCI runtime exec failed: exec failed: unable to start container process: exec: "Cli": executable file not found in $PATH: unknown
```

In order to fix, I tried downloading the latest available Arista image, 4.32.0F. This works without flaw, topology starts up and the above Cli command works.

There was one other small flaw I noticed. The command given to connect to the srl container is wrong.

It is listed as:

```
docker exec -it clab-srlceos01-srl1 bash
```

It should be:

```
docker exec -it clab-srlceos01-srl bash
```

james-m-henderson:
Here is the topology file with the "good" version activated and the "bad" version commented out:

```
# topology documentation: http://containerlab.dev/lab-examples/srl-ceos/
name: srlceos01

topology:
  nodes:
    srl:
      kind: nokia_srlinux
      image: ghcr.io/nokia/srlinux
    ceos:
      kind: arista_ceos
      image: ceos:4.32.0F
      #image: ceos:4.25.11M

  links:
    - endpoints: ["srl:e1-1", "ceos:eth1"]
```

hellt:
thanks @james-m-henderson 
do you want to raise a PR with those fixes and get into the prestigious contributors ranks? =)

thank you!

james-m-henderson:
Hi Roman,

I already started on one, but I wanted to know if I should update all
arista lab examples and docs as well? In addition some are on a different
version, which I haven’t tested yet. Would it make sense to just
standardize on the newest?

Thanks,

James

hellt:
Feel free to update all of them to the version you tested. We don't have sophisticated labs  that exercise advanced platform features, so you can safely bump them all to the 4.32.0F

james-m-henderson:
[...]

PR #2034, opened 2024-05-04
Standardize on ceos:4.32.0F

james-m-henderson:
Fixes https://github.com/srl-labs/containerlab/issues/2033

Arista (4.25.0F) is not present on the Arista website for download. The closest available seems to be 4.25.11M.

In testing 4.25.11M did not have the Cli come up, and it would hang while initializing. The standardizes all Arista labs and docs to reference version 4.32.0F, which is working.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -223,7 +223,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Pull ceos image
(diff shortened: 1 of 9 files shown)


hellt:
thanks @james-m-henderson 
can you confirm the method you loaded the ceos image?
was it `docker load -i <tar.gz>` or `docker import <tar.gz>`?

james-m-henderson:
Docker import. I imported both images exactly the same way, two different
lab docker instances, same behaviour each time. If there are some debug or
log output you need I can get it for you Monday.

Thanks,

James

hellt:
thanks @james-m-henderson 
I have updated the CI to use the same ceos image, the tests passed.

PR #2055, opened 2024-05-28
Implement Dummy Interface

steiler:
The Linux Dummy interface is the /dev/null of network interfaces.
It can be used if you need a certain amount of interfaces. They can be set to operational up without requiring a peer endpoint.


hellt:
thanks mate
for context for me when getting to the docs https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking#dummy

PR #2062, opened 2024-06-03
simplify dev container setup to MVP

ankudinov:
- don't use dockerfile
- use MS image


hellt:
Hi @ankudinov 
I am bringin back dockerfile since I wanted to use some elaborated zsh config in the future, as well as to install clab from a package repo to avoid pull rate limits from docker

PR #2085, opened 2024-06-12
Vmx lab

iproute-net:
add topology file to the vmx lab


hellt:
Hi @iproute-net 

Before you take this PR further I must inform you that containerlab docs already has lab examples for vMX and cEOS 

* https://containerlab.dev/lab-examples/vr-vmx/
* https://containerlab.dev/lab-examples/srl-ceos/

The containerlab docs are meant to provide examples for different kinds, but it is not a collection of labs with different permutations of the NOSes.
Users are free to create the labs in their own repos or at clabs.netdevops.me

iproute-net:
I made this one because the device kinds are changed in the newer container lab and the "juniper_vmx" kind is no longer supported.

PR #2089, opened 2024-06-12
Devcontainer docs

hellt:


hellt:
Hi @ankudinov 
feel free to give this doc a look. As with any other doc piece in containerlab it is a first iteration of them, will add more as we go.

Just wanted to make sure the initial load is sane enough.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+The `image` field points to the Containerlab Dev Container image that would define your Codespaces environment. Containerlab provides devcontainer images, and you can see all available tags on the [package' page][clab-devcontainer].
+
+The image tag corresponds to the containerlab release version that is pre-installed in the image. You can choose the version that you want to use in your lab.
+
+### Host requirements
+
+Another important part of the `devcontainer.json` file is the `hostRequirements` field that defines the machine type that Codespaces environment will run on. Codespaces offer a small selection of machine types that differ in the number of CPUs, RAM, and storage. You can choose the machine type that fits your lab requirements.
+
+<figure markdown>
+![machines](https://docs.github.com/assets/cb-35329/mw-1440/images/help/codespaces/change-machine-type-choice.webp){width=80%}
```
ankudinov, line 143: Make sure you add a note somewhere that by default 4 cpus is the maximum. And bigger machines have to be requested from GH support. They usually still give them for free however. Something like:

By default only 2 and 4-core machines are available. If you need bigger machines, please contact [Github support](https://support.github.com/) to enable them. Keep in mind though, that Github may ask for a payment option to enable them and free time will run out faster with bigger instances.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+
+Using the machine types displayed above you can tune the `hostRequirements` section by choosing the machine type that fits the requirements of your lab.
+
+/// note
+Codespaces VMs support nested virtualization, so you can run [VM-based kinds](../manual/vrnetlab.md) :partying_face:
+///
+
+### Mounts
+
+Lastly we specify the folder name we want our repo to be cloned in, this is governed by the `workspaceFolder` field. The `workspaceMount` field is used to mount the local workspace folder to the Codespaces environment and is static for all labs.
```
ankudinov, line 154: For the specific case described in this doc I'd skip `workspaceMount` and `workspaceFolder` in devcontainer.json definition. By default the entire repository will be mounted as a workspace and there is no need to define anything. I'm normally using workspaceMount/workspaceFolder when I have to mount subdirectory (multiple labs in the same repo) or similar.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Testing the environment
+
+Once you added the `.devcontainer/devcontainer.json` file to your lab repository, you can test the environment locally and in Codespaces.
+
+#### Local testing
+
+Testing the environment locally is a litmus test to ensure that the devcontainer.json file is correct and the environment can be started. But since the environment runs locally, it doesn't test the Codespaces-specific settings like the machine type and CPU/RAM requirements.
+
+To deploy the environment locally, use the VS Code command panel (`Cmd/Ctrl+Shift+P`) and execute `Dev Containers: Rebuild And Reopen In Container` action. This will trigger the VS Code to build the container and open the environment in the container.
```
ankudinov, line 164: Mention that Dev Containers VSCode extension must be installed: ms-vscode-remote.remote-containers
This may not be obvious for everyone.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+You'll get the environment up and running in a couple of minutes and you can test it to ensure that it works as expected.
+
+### Launching the environment
+
+Once you are satisfied with the environment, you can add a nice button to the README file that will allow users to start the environment with a single click.
+
+/// tab | Button
+<div align=center markdown>
+<a href="https://codespaces.new/srl-labs/srlinux-vlan-handling-lab?quickstart=1">
+<img src="https://gitlab.com/rdodin/pics/-/wikis/uploads/d78a6f9f6869b3ac3c286928dd52fa08/run_in_codespaces-v1.svg?sanitize=true" style="width:50%"/></a>
```
ankudinov, line 183: Can I reuse this .svg in my demos and labs? Like it a lot. ;)

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+
+You'll get the environment up and running in a couple of minutes and you can test it to ensure that it works as expected.
+
+### Launching the environment
+
+Once you are satisfied with the environment, you can add a nice button to the README file that will allow users to start the environment with a single click.
+
+/// tab | Button
+<div align=center markdown>
+<a href="https://codespaces.new/srl-labs/srlinux-vlan-handling-lab?quickstart=1">
```
ankudinov, line 182: May be add some explanation to explain the deeplink magic or (better) refer to Github docs directly: https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/setting-up-your-repository/facilitating-quick-creation-and-resumption-of-codespaces

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+4. And configure the Codespace settings manually.
+
+## Dev Container
+
+The key pillar behind Codespaces is the Containerlab' Dev Container image that defines the environment in which the lab will run. The Dev Container image is a Docker image that contains all the necessary tools and dependencies to run Containerlab and other tools that you might need in the lab.
+
+The following two files define Containerlab's Dev Container image:
+
+1. [devcontainer.json](https://github.com/srl-labs/containerlab/blob/main/.devcontainer/devcontainer.json) - the Dev Container configuration file that defines how the environment is built, configured and launched.
+2. [Dockerfile](https://github.com/srl-labs/containerlab/blob/main/.devcontainer/Dockerfile) - the Dockerfile that the Dev Container is built from.
```
ankudinov, line 227: Dockerfile is not mandatory part of dev container. The core of dev containers are pre-build images, features and dev container spec. May be mention Dockerfile as an option, I think we have to change it later anyway. Refer to dev container spec somewhere: https://containers.dev/implementors/spec/

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+
+The resulting Dev Container image contains the following tools and dependencies:
+
+- containerlab binary installed via the deb repository
+- docker in docker setup
+- [gNMIc](https://gnmic.openconfig.net) and [gNOIc](https://gnoic.kmrd.dev) tools
+- Go SDK
+- Python 3 with pyenv
```
ankudinov, line 235: No pyenv please!! =) I think, this can hurt in future. Happy to be wrong though.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+- `gh` CLI tool
+- zsh shell with oh-my-zsh configuration
+- VS Code plugins
+
+## Tips, tricks and known issues
+
+### Authenticating with ghcr.io container registry
+
+While we still live in a world where vendor' Network OS images are not available in the public container registries, you might want to push the images you've built to your GitHub container registry as a `private` image. That way you get to benefit from building the image once and reusing it across different labs while keeping the image private.
```
ankudinov, line 244: NO-NO-NO! Please remove that before some legal department saw this. =)
The fact that some images are not on container registries is fundamentally wrong, but we can not advise that to users. Even though this is very theoretical - they are legally liable if they upload images anywhere as normally it's covered by EULA agreements etc.
We can however advice upload directly to a codespace and I can add a section about using [EOS downloader](https://pypi.org/project/eos-downloader/) later.

ankudinov (changes requested):
It would be great to make some changes in the doc as I mentioned in the comments. Especially the private registry part MUST be removed.

docs/manual/codespaces.md
ankudinov: Maybe also add a section where other people/vendors can index their repositories with Codespaces support and provide corresponding instructions. I'll be very happy to list mine.
If there is nothing listed - the community can poke the vendor with a stick if there will be demand.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Adding codespaces to your lab
+
+By now you should be willing to try running your labs in Codespaces. To our luck, it is super simple, all you need to do is create a `.devcontainer/devcontainer.json` file in your lab repository that will define the Codespaces environment. The file should look similar to this:
+
+```json
+{
+    "image": "ghcr.io/srl-labs/containerlab/clab-devcontainer:0.55.0",
+    "hostRequirements": {
+        "cpus": 2,
```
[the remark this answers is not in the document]
hellt: I wouldn't put this if I haven't tested. I have two labs succesfully running with 2 vcpus

- https://github.com/srl-labs/srlinux-vlan-handling-lab
- https://github.com/srl-labs/srl-acl-lab

The community has a high number of FRR based labs, and FRR is even more lightweight, so I will not rule out 2vcpu machine types

docs/manual/codespaces.md
hellt: this is not the case today. I haven't asked anyone, and had 2/4/8/16 vcpu machines available, which seems to be inline with the public docs.

I will put a note that you still can ask GH to add more beefy machines by writing to the support email

docs/manual/codespaces.md
hellt: sure

docs/manual/codespaces.md
ankudinov: That sounds great and means it was improved meanwhile. I think a generic note about "beefy machines" should be sufficient in this case.

docs/manual/codespaces.md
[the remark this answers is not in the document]
ankudinov: Agree 2 cpus should be sufficient for this size. To clarify - I know that Containerlab community is smart enough to figure out simple things, but prefer to put safer recommendation as people are often blindly copy-pasting. So, 4 cpus is a safer copy-paste.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,281 @@
...
+
+You will be prompted to authenticate with your GitHub account and the new token will be generated for you. Then you can login to the registry using the newly acquired token:
+
+```bash
+gh auth token | \
+docker login ghcr.io -u $(cat /home/vscode/.config/gh/hosts.yml | \
+grep user: | awk '{print $2}') --password-stdin
+```
+
+<small>Of course you can also install tailscale or any other 0-tier VPN to access any other self-hosted private registry.</small>
```
ankudinov, line 268: Replace lines 248 to 268 with:

If the image you are planning to use is not available on any public container registry, you can easily upload it to the Codespace when container starts using Codespace GUI or Github CLI. (Optional: we can add some instructions on how to upload images with GH CLI)  
After uploading the image, you can import it with `docker import ${IMAGE_FILENAME} ${IMAGE_NAME}`  
Alternatively you can use vendor provided tools to automate image download.

Everything written above makes total technical sense. However we still live in the world where most vendors resist publishing containers on any open (or private) registry and disrespect any disobedience.  
After all - let people poke vendors and add some pressure. This will not change if we use workarounds.

hellt: this is not about the vendors, this is how to make use of the collection of private images one might have on their ghcr.io registry.

What you are proposing to add can be another subsection under the tips section.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+
+The best part about Codespaces is that it has a suitable free tier. GitHub offers **120 cpu-hours/month and 15 GB storage for free** to all users. This means that you can run a Codespace environment for 120 cpu-hours per month without any charges. This is a compelling offer for those who
+
+- want to spin up a lab provided by others to get through a tutorial or a demo
+- don't need to run the labs 24/7
+- want to demo labs on-the-go without managing the local environment
+
+You can select which GitHub machine type you want to use for your project; each machine type is characterized by the amount of CPUs/RAM/Storage that it is equipped with, and based on that you can calculate how many cpu-hours you would consume running a lab with a chosen machine type.
+
+![pic](https://gitlab.com/rdodin/pics/-/wikis/uploads/fc7bfa89c921f965ce61f6eb5e86f312/image.png){.img-shadow}
```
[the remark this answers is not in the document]
hellt: done in [48d3723](https://github.com/srl-labs/containerlab/pull/2089/commits/48d3723db620cefcc115ec3ed7791b038c63813a)

docs/manual/codespaces.md
ankudinov: OK, got it know "private image" != "vendor image". I think is neutral enough. LGTM especially with vendor subsection for clarity.

docs/manual/codespaces.md
```diff
@@ -0,0 +1,273 @@
...
+
+///
+
+#### Billing
+
+It is always a good idea to periodically check how much of the cpu-hours you've consumed and check the remaining quota. Your billing information is available in the [Billing settings][billing].
+
+![billing](https://gitlab.com/rdodin/pics/-/wikis/uploads/4bf6eb89bd3b4f2d4b05ecdc3ab22675/image.png){.img-shadow}
+
+The screenshot shows that 10 cpu-hours out of 120 available were consumed in the current month' period and the codespaces environments occupy 8.15 GB of storage out of 15 GB included. So far it is all well within the free tier limits.[^5]
```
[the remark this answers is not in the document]
hellt: The space thingy is complex, [the way gh counts that is mystical](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-github-codespaces#about-billing-for-storage-usage), and I am not yet sure how to read the data usage properly.

If I launch the codespaces from this repo https://github.com/srl-labs/srl-telemetry-lab I get a codespaces that measures as 7.39GB

image

If I enter into this VM and use `df -h` than I see that 8GB.

If I start the lab from here https://github.com/srl-labs/srlinux-vlan-handling-lab/ which uses the same image for codespaces, it reports 3.3GB. And I don't quite get it at this moment...

image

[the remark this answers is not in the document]
ankudinov: I think the best way to find out what's happening is to go to your profile -> Billings and Plans and click "Get usage report" right to "Usage this month" section. Github will send and email with a link to a CSV explaining all the billing calculations. Also expand Storage section under Codespaces and make sure you don't see "Prebuild Storage" used.  
Here is my usage as of today (feel free to use it as a healthy example):
image
As you can see, CPU is ahead of storage and that's what I expect at all times for a proper Codespace configuration.  
And here is an example Codespace:
image
My understanding is that if I run this Codespace constantly for a month - I'd be charged for 3.36 GB-month. However I always stop it quick and storage usage is at minimum. Prebuild (if enabled) would change that and start consuming storage 24/7. But we don't need it as we are relying on Github packages and pre-build container images that are free for public repos.  

Here is the relevant quote from GH docs:

For example, if you have one codespace that uses 100 GB of storage and has existed for one hour you will have used 100 / (24 * 30) = 0.1388 GB-months of storage in a 30-day month. If your use of GitHub Codespaces during a 30-day month consists of two 100 GB codespaces that both existed for three full days then there will be 24 * 3 hourly reports for the storage of these codespaces, giving a total of (24 * 3) * 200 / (24 * 30) = 20 GB-months.

So, I strongly suspect you have prebuilds enabled. Although it could be something else that will be visible in CSV report anyway.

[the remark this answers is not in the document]
hellt: The report shows some numbers, and I guess they totalling to the 8GB of storage quota consumed, but as I mentioned, the calculation is complex to me

image

docs/manual/codespaces.md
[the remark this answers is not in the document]
hellt: Fair, I changed the cpu number to 4 - [5afe38a](https://github.com/srl-labs/containerlab/pull/2089/commits/5afe38abd54815de93dcf5beda1c4d0a5728693a)

docs/manual/codespaces.md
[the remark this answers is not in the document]
ankudinov: OK. Looks like you have not deleted some Codespaces after starting them, that would explain storage utilization on some days without compute.  
TLDR; I'd add a section to the doc about deleting Codespaces from https://github.com/codespaces after the workshop and maybe suggest lowering the retention period to 1 day.
I assume 901 hours on 11.06 is a bug. That would use all free compute if true.

docs/manual/codespaces.md
hellt: agree, I added it initially to have a short and sweet path

`/my-lab-name` vs `/workspaces/my-lab-name`

but since now the prompt is on two lines it is not so bad visually, so I removed it [2608f99](https://github.com/srl-labs/containerlab/pull/2089/commits/2608f99d61a021922a90c67f81b41c6d55372f0e)

docs/manual/codespaces.md
hellt: good call. Done in [12395ab](https://github.com/srl-labs/containerlab/pull/2089/commits/12395ab4b87c50440ffb7ac4d5b3f2ab580ceefa)

docs/manual/codespaces.md
hellt: This section is about Containerlab' dev container setup, which uses the Dockerfile

docs/manual/codespaces.md
hellt: what harm you see in having pyenv?
If one doesn't need it - they are free not to use it. The pyenv itself is a lightweight wrapper, it doesn't consume anything from the resources.

docs/manual/codespaces.md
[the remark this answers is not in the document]
hellt: yep, definitely some bugs in calculation for a couple of runs.

I added a tip about 1 day retention period - [e1f2b7f](https://github.com/srl-labs/containerlab/pull/2089/commits/e1f2b7fcbcc770b4d38e8c1cdef155bef231f08b)

docs/manual/codespaces.md
ankudinov: Totally up to you. My point of view is that container must be complete and closed environment with all dependencies installed inside and fixed path to everything. venv/pyenv open... possibilities. =) You can have Python3.10 on a Python3.11 container, pointing to a different path and sometimes even sourced from the repo directly. I have seen quite a few weird things happening while helping to troubleshoot. Therefore my preference to have all dependancies as part of container build is very strong. However in ideal case we should still provide some options by building multiple containers with different tags, Python versions, etc. As that will be separate matrix PR somewhere in the future - we can remove Pyenv once there is an alternative.

docs/manual/codespaces.md
hellt: yes, thanks, done in [31c91f8](https://github.com/srl-labs/containerlab/pull/2089/commits/31c91f863447354e3c33333d28d4ffce1fab412e)

docs/manual/codespaces.md
hellt: I agree, we need to look at the `-slim` images next where everything dev related is stripped completely.

For now the `<version>` tag will contain the dev tools such as pyenv

PR #2107, opened 2024-06-24
Add Ostinato traffic generator as a supported node

pstavirs:
Added documentation on how to use the Ostinato traffic generator with Containerlab.

A future PR will include a lab example


hellt:
Thanks @pstavirs 
Lgtm

PR #2111, opened 2024-06-26
Add Ostinato lab example ost-srl

pstavirs:
~~This PR is work-in-progress awaiting #2110 to be merged first so that the diagram URL can be updated in the lab example before merging this PR~~

#2110 is merged. This PR is ready for merge now.

PR #2112, opened 2024-06-26
Fix bullet list indentation in ost-srl lab example

pstavirs:
The editorial changes in #2111 inadvertently removed indentation of some bullet list items


hellt:
gotcha, had to fix the linting rule

PR #2115, opened 2024-06-29
Fix XRv9k username and password env vars

kaelemc:
Currently supplying `USERNAME` and `PASSWORD` environment variables in the topology definition file do nothing. 

Current code doesn't uses the hardcoded credentials (clab:clab@123) for the command argument, the proposed change fixes this and respects the value of the environment variable.

(was discussed in the containerlab discord)


hellt:
thanks!

PR #2119, opened 2024-07-02
Specify in XRv9K documentation that at least one dataplane interface is required.

kaelemc:
If the XRv9k doesn't have at minimum one data-plane interface defined in the topology file, the bootstrap process will fail and infinitely loop meaning the node will not boot properly and can't be accessed via ssh.

This adds a callout in the docs for the xrv9k which specifies this requirement.

Should this be an 'info' or 'note' callout?


hellt:
thanks @kaelemc 
I changed it to the `note` admonition.

PR #2124, opened 2024-07-04
Interface aliases

vista-:
This PR introduces the interface aliasing feature.

## What's this feature about?

Network OS's like to use their own naming conventions, with interfaces referred to as `GigabitEthernet1/3`, `1/2/1`, `ge-0/0/4`, and many other formats. While interfaces in containerised NOSes generally conform to Linux interface naming standards out of necessity, we have no such luck with virtualised NOSes, such as those running in `vrnetlab` containers.

Currently, users of `vrnetlab`-based images need to manually figure out the mapping between container interfaces (which Containerlab creates and uses the `ethX` naming convention) and NOS's VM interfaces, to which the `vrnetlab` launch script maps. This is also the case for containerised NOSes where there is a disconnect between the interface names in the container and within the NOS itself, which might perform some mapping internally.

Interface aliases help by allowing users to directly use interface names matching NOS naming conventions in Containerlab. This not only simplifies creating topologies (and leads to less wiring errors and quicker transition from ideas to actual topologies), but also enables easy topology generation based on modelled networks, or even network inventories.

### As an end-user of Containerlab

Interface aliases are used in the same way that you would declare any link in Containerlab. The documentation should provide users information about the interface naming standards of a Kind.

Existing naming conventions are also preserved for all Kinds, allowing for backward compatibility with existing topologies, even allowing for mixed use of existing interface names and interface aliases in the same topology.

As an example, imagine we want to create a lab with three different `vrnetlab`-based Kinds: vEOS, CSR1000v and vSRX, cabled like this:

|           A side | B side           |
|-----------------:|------------------|
|    vSRX ge-0/0/2 | vEOS Ethernet1/2 |
|     CSR1000v Gi5 | vSRX ge-0/0/5    |
| vEOS Ethernet1/3 | CSR1000v Gi3     |

Using the `ethX` interface naming convention, the topology would look like this:

```
  links:
    - endpoints: ["vSRX:eth3", "vEOS:eth2"]
    - endpoints: ["CSR1000v:eth4", "vSRX:eth6"]
    - endpoints: ["vEOS:eth3", "CSR1000v:eth2"]
```

Note the three different kinds of offset used here on the three different NOSes!

Using aliased interface names, the topology definition becomes much more straightforward:

```
  links:
    - endpoints: ["vSRX:ge-0/0/2", "vEOS:Ethernet1/2"]
    - endpoints: ["CSR1000v:Gi5", "vSRX:ge-0/0/5"]
    - endpoints: ["vEOS:Ethernet1/3", "CSR1000v:Gi3"]
 ```
 
Both topology definitions result in the same lab being deployed, but the latter is easier to write and to understand.

### As a Containerlab developer

In most cases, if you would like to add interface alias support to an existing or new node (Kind) type, these are the key takeaways:

- `InterfaceRegexp` (*regexp.Regexp) should be a regular expression which matches on the NOS' interface naming conventions.  
The default `VRNode` interface mapping logic expects a named capture capture group called 'port' capturing the port number  
Example: `regexp.MustCompile(``(?:Gi|GigabitEthernet)1/(?P<port>\d+)$``)`  
- `InterfaceOffset` (int) is a numeric negative offset of the port number compared to `eth1`, the first data port in the containers; by default, this is set to `0`  
If using a 0-based port numbering system, such as Junos (`ge-0/0/[0]`), this should be `0`.  
For 1-based port numbering, such as on Cisco and Arista boxen (`Et1/[1]`), this should be `1`.  
In case the first port is reserved as the management interface (when there's no separate management interface), and data port numbering begins with 2, this should be `2`.  
- `InterfaceHelp` (string) gets displayed to the user if the interface naming check function in `VRNode` fails
This should be a concise explanation about the expected interface naming formats.  
Example: `"(et|xe|ge)-0/0/X (where X >= 0) or ethX (where X >= 1)"`  

these three variables should be set in the `Init` function of the node (Kind).

Additional variables help with customising the way interface aliasing works for a specific node type:
- `InterfaceMappedPrefix` (string) is the prefix on which the calculated interface index gets tacked onto. Default: "eth"
- `FirstDataIfIndex` (int) is the index of the first data interface of the container, and acts as a positive offset. This is useful as many NOSes reserve the first interface in the container/VM as the management interface. Default: `0`

For convenience, vrnetlab-based nodes should be based off of the `VRNode` type, which automatically sets the correct `InterfaceMappedPrefix` ("eth") and `FirstDataIfIndex` (`1`).

For more complex/advanced interface alias mapping, where the DefaultNode-provided implementation falls short, developers can override the interface aliasing/mapping behaviour by overriding the following two functions:

- `CalculateInterfaceIndex(ifName string) (int, error)` parses the interface name and calculates the mapped interface index. This is called by `GetMappedInterfaceName` for each interface that matches `InterfaceRegexp`.
- `GetMappedInterface (ifName string) (string, error)` returns with the mapped interface name for the provided interface alias. This is called by `AddPoint` during the node deployment process for each interface that matched `InterfaceRegexp`.

Additionally, `AddEndpoint` and `CheckInterfaceName` can also be overridden. The default implementation for the latter also ensures that no overlapping interfaces exist within the node (e.g. when using a mixed `ethX` and interface alias in the topology definition.)

When adding new interface aliases, you're encouraged to write tests to ensure the new interface aliasing logic works properly and that it won't break existing topologies' `ethX`-based interface naming conventions.

## How is this implemented?

The interface aliasing feature is made available in `DefaultNode`. The actual interfaces created in the container images still use the same, compatible interface names, with the interface aliases getting transparently mapped during the process of adding each `Endpoint` to the node. The mapped interface name ends up being the actual `IfaceName`, while the interface alias is saved as `IfaceAlias`. The mapped interface index gets calculated based on (at least in the default `DefaultNode) implementation):

- The extracted port index in the interface alias (with the help  of `InterfaceRegexp`)
- The offset of the data index compared to the first available data interface's index, as set in `InterfaceOffset`
- The first available data interface's index, by default set to `0` in `FirstDataIfIndex`. In `VRNode`, this is set to `1` by default, which should  be true in general for `vrnetlab`-based containers

Using the three NOSes in the example above, we can do the following calculations to show off how this works for each VRNode-based NOS's **second** usable data interface:
```
vSRX: ge-0/0/[2] --> 2 is the extracted port index, 0 for 0-based interface numbering, 1 as the first `ethX` data interface
2 - 0 + 1 = 3 --> `eth3`
vEOS: Ethernet1/[3] --> 3 is the extracted port index, 1 for 1-based interface numbering, 1 as the first `ethX` data interface
3 - 1 + 1 = 3 --> `eth3`
CSR1000v: Gi[4] --> 4 is the extracted port index, 2 for 1-based interface numbering AND first port is taken for management, 1 as the first `ethX` interface
4 - 2 + 1 = 3 --> `eth3` 
```

For ease of use, the interface alias is preserved as... well, an actual interface alias! Netlink allows us to set arbitrary text as "interface aliases" (think of them as descriptions) on links:

```
324: eth1@if325: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether aa:c1:ab:07:c6:2f brd ff:ff:ff:ff:ff:ff link-netnsid 1
    alias ge-0/0/0
    altname ge-0-0-0
326: eth5@if327: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether aa:c1:ab:2c:66:b6 brd ff:ff:ff:ff:ff:ff link-netnsid 3
    alias ge-0/0/4
    altname ge-0-0-4
328: eth3@if329: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether aa:c1:ab:7d:6e:0f brd ff:ff:ff:ff:ff:ff link-netnsid 2
    alias ge-0/0/2
    altname ge-0-0-2
```

This allows us to side-step two major limitations in Linux interface naming conventions: the 15 character limit (which Linux interface altnames also allow us to sidestep) and the forbidden character '/', which many NOS' interface naming conventions favor (and altnames still don't allow us to use them).

An interface altname is also added with illegal characters replaced with a '-'.

The command line tools dealing with interfaces (`generate` and `netem`) have also been updated to be interface alias-aware.

## Supported node types

The following nodes (Kinds) have been updated to support interface aliases:

### vrnetlab-based
- Aruba AOS-CX
- Cisco C8000v
- Cisco CSRv1000
- FreeBSD
- Fortinet Fortigate
- Cisco FTDv
- Cisco Nexus9000v
- OpenBSD
- Palo Alto PA-VM
- Mikrotik RouterOS
- Nokia SR-OS (only basic mapping, port index only, errors out on interfaces that don't look like 1/1/X)
- Arista vEOS
- Juniper vJunosEvolved
- Juniper vJunos-router
- Juniper vJunos-switch
- Juniper vMX
- Juniper vQFX
- Juniper vSRX
- Cisco XRv
- Cisco XR9000v
### container-based
- Nokia SR Linux (full support for linecard/port/channel mapping)

What is **not** supported at the moment:

- Checkpoint CloudGuard (not enough information available about interface naming/inconsistent?)
- IPInfusion OcNOS (not enough information available about interface naming)
- SONiC (not enough information available about interface naming)

### CODE REVIEW

**Comment by vista-** [comment]
Checked the DeepSource errors: the remaining one is erroneously reported.

**Comment by vista-** [comment]
I see why the unit tests are failing... I'll push a fix for all the unit tests.

**Comment by vista-** [comment]
Boom, made graph interface alias-aware and fixed a bug affecting longer interface names!
Screenshot 2024-07-04 at 23 10 19

File: docs/manual/kinds/vr-csr.md [hunk 1]
```diff
@@ -35,12 +35,19 @@ Cisco CSR1000v node launched with containerlab can be managed via the following
 !!!info
     Default user credentials: `admin:admin`
 
+## Interface naming
+
+CSR1000v nodes use the interface naming convention `GigabitEthernetX` (or `GiX`, both are accepted), where X denotes the port number.
+
+!!!warning
+    Data port numbering starts at `2`, as `Gi1` is reserved for management connectivity. Attempting to use `Gi1` in a containerlab topology will result in an error.
+
```
hellt, line 44: Hi @vista- 
regarding this note, I think @ipspace tied this to an issue with the way csr1000v was handling changing mac addresses

https://github.com/hellt/vrnetlab/pull/206

it seems with https://github.com/hellt/vrnetlab/pull/206 merged we should expect csr1000v to map `eth1` to `Gi1`

ipspace: CSR interfaces start with Gi1, which is the management interface (mapped into eth0), so it's true that the first data-plane interface is Gi2 (mapped into eth1). The problem I fixed in in hellt/vrnetlab#206 was CSR mapping the first interface into Gi2 (because of changed MAC address)

hellt: ah, I see, thanks for clarifying @ipspace

links/link.go
```diff
@@ -426,6 +426,13 @@ func SetNameMACAndUpInterface(l netlink.Link, endpt Endpoint) func(ns.NetNS) err
 			}
 		}
 
+		if endpt.GetIfaceAlias() != "" {
+			err := netlink.LinkSetAlias(l, endpt.GetIfaceAlias())
+			if err != nil {
+				return err
+			}
+		}
+
```
hellt, line 435: @vista- 
instead of link aliases we should use link AltNames

We switched from using Aliases to using AltNames (https://github.com/srl-labs/containerlab/pull/1908) since this makes it easier to query links by name with a single netlink call, while altnames provide all the benefits we are looking for (long names)

https://x.com/ntdvps/status/1763493657804427299

hellt: mmm. I see why you did that

```
and the forbidden character '/', which many NOS' interface naming conventions favor (and altnames still don't allow us to use them)
```

I didn't know about altnames not allowing `/`

cmd/tools_netem.go
```diff
@@ -123,7 +124,16 @@ func netemSetFn(_ *cobra.Command, _ []string) error {
...
+		netemIfLink, err := netlink.LinkByName(netemInterface)
+		if err != nil {
+			// Look for aliased interfaces as well
+			netemIfLink, err = netlink.LinkByAlias(netemInterface)
+			if err != nil {
+				return err
+			}
+		}
+		netemIfName := netemIfLink.Attrs().Name
+		link, err := net.InterfaceByName(netemIfName)
```
hellt, line 136: What I propose we should do with interface names (in general) is:

1. if an interface name contains `/` we swap it with `-` and always set this as altname (while keeping the alias)
2. that way, we can always use `netlink.LinkByName(sanitizedInterfaceName)` when fetching a link, regardless if it is long or short. With a single API call.
3. This would eliminated the logic to fetch a link via alias or name.

we can use smth like this

```go
func (s *Socket) SanitizeName() {
    var sb strings.Builder // Efficient way to build strings
    sb.Grow(len(s.Name))   // Allocate enough memory to avoid reallocation

    for _, char := range s.Name {
        switch char {
        case '.', ' ', '_':
            sb.WriteRune('-') // Replace with '-'
        default:
            sb.WriteRune(char) // Keep the original character
        }
    }

    s.Name = sb.String()
}
```

Should happen here https://github.com/vista-/containerlab/blob/7cb1875ea02b0647f9471beb61255d8e563ca73b/links/link.go#L414

vista-: Added!

hellt:
Let's ship it, thanks @vista- !

vista-:
Thanks for all your help and the effort you put into this PR too @hellt! :shipit:

PR #2137, opened 2024-07-10
Add suport for EL distros in one-liner install script.

kaelemc:
This PR enables the one liner install script to support RHEL & other EL counterparts like Fedora, Rocky and CentOS to name a few.

Currently tested on Rocky 9.3. This is in draft PR status while I test other distros & versions.

### DIFF
--- a/utils/quick-setup.sh
+++ b/utils/quick-setup.sh
@@ -7,6 +7,8 @@ function check_os {
             DISTRO_TYPE="debian"
         elif [ "$ID" = "ubuntu" ]; then
             DISTRO_TYPE="ubuntu"
+        elif [ "$ID" = "rocky" ]; then
+            DISTRO_TYPE="rhel"
         else
             echo "This is not Debian or Ubuntu"
         fi
@@ -16,13 +18,13 @@ function check_os {
 }
 
 function install-docker {
-    # check OS to determine how to install docker
-    check_os
 
     if [ "${DISTRO_TYPE}" = "debian" ]; then
         install-docker-debian
     elif [ "${DISTRO_TYPE}" = "ubuntu" ]; then
         install-docker-ubuntu
+    elif [ "${DISTRO_TYPE}" = "rhel" ]; then
+        install-docker-rhel
     else
         echo "Cannot determine the operating system"
         exit 1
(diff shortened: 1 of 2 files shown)


kaelemc:
Confirmed working on RHEL 9 (Developer edition) and CentOS 9.

utils/quick-setup.sh
```diff
@@ -73,32 +75,75 @@ function install-docker-ubuntu {
...
+function install-gh-cli-rhel {
+    sudo dnf install -y 'dnf-command(config-manager)'
+    sudo dnf config-manager -y --add-repo https://cli.github.com/packages/rpm/gh-cli.repo
+    sudo dnf install git -y
+    sudo dnf install -y gh --repo gh-cli
+}
 
-    sudo apt update && sudo apt install -y containerlab
+function install-containerlab {
+    sudo bash -c "$(curl -sL https://get.containerlab.dev)"
```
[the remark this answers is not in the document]
kaelemc: Fixed

kaelemc:
Confirmed basic containerlab by running [srl-quickstart](https://github.com/srl-labs/containerlab/blob/main/lab-examples/srl-quickstart/srl01.clab.yml) on:

- Rocky 9.3
- Rocky 8.8
- Fedora Server 40
- RHEL 9 w/Developer Subscription
- Ubuntu 22.04 Server

hellt:
Thanks @kaelemc 

Would you mind adding the tested systems here as well?

https://containerlab.dev/install/#quick-setup

I also see that dnf is used to install packages, I guess this rules out ancient systems like centos7 that don't have dnf client? Or maybe it does?

kaelemc:
Oh yes, I will add the relevant documentation. Thanks for the reminder.

Yes `dnf` is used to install github cli and for the Fedora docker installation. 

RHEL 7 is going EOL this year, and I personally think it's unlikely people would be installing clab on a RHEL 7 based distro so we are probably fine on that front.

kaelemc:
Scrap that, RHEL7/CentOS already is EOL as of June 2024.

kaelemc:
Just want your opinion on two things:

1. Should the script perform (or the documentation specify) that users can/should perform the [docker postinstall](https://docs.docker.com/engine/install/linux-postinstall/). It makes things easier and cleaner for the user (assuming the machine only runs containerlab because of the security implications etc.), . Effectively means newer users might find it easier to use containerlab out-of-the-box. 

2. Should the script install `make` which is required for vrnetlab? Or should we leave that up to the user to do?

hellt:
@kaelemc 
yes, I always wanted to add the docker user group to this script. I see that it is part of the postinstall, so we should definitely do this while we are at it.

and yes, make we should install as well

kaelemc:
@hellt Added those two things. Confirmed working on Ubuntu & Rocky.

utils/quick-setup.sh
```diff
@@ -73,33 +76,136 @@ function install-docker-ubuntu {
...
+
+    # diverges from the instructions. This means docker daemon starts on each boot.
+    sudo systemctl enable --now docker
+}
+
+function post-install-docker {
+    # instructions from:
+    # https://docs.docker.com/engine/install/linux-postinstall/
+    sudo groupadd docker
+    sudo usermod -aG docker "$SUDO_USER"
```
hellt, line 129: I think it is useful to add this

```
newgrp docker
```

I once had a problem with login/logout not helping the case. The docs say that sometimes a reboot is needed, but also mention that this command might help

kaelemc: I tried this in the shell script along with `su`. It won't/doesn't work due to the script executing as root.

kaelemc: We could add to the docs that the user can either log out & back in _or_ execute `newgrp docker`?

hellt: yes, lets put in a note then

kaelemc: Done.

hellt:
Thanks again @kaelemc

PR #2167, opened 2024-08-18
Don't crash if a Docker password contains a colon

remram44:
One of my passwords saved in `~/.docker/config.json` contains a colon. Because of this, every attempt to use `containerlab deploy` would fail with:

```
INFO[0000] Containerlab v0.56.0 started                 
INFO[0000] Parsing & checking topology file: sonic01.clab.yml 
Error: unexpected auth string
```


hellt:
thanks!

PR #2168, opened 2024-08-18
Allow overriding data installation prefix

remram44:
Allow changing the default installation directory where containerlab expects to find templates, using $CLAB_PREFIX.

Alternatively there could be a $CLAB_TEMPLATES_DIR or similar.

### DIFF
--- a/clab/config/template.go
+++ b/clab/config/template.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"text/template"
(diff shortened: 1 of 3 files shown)


hellt:
Hi @remram44 
I wonder, what use case do you have that warrants this customization? Out of these three artifacts (cfg templates, json template path, graph source templates) which ones do you care about being at a custom path and why?

remram44:
The graph templates (+static).

Having a single installation path makes it difficult to use different versions of containerlab or to run on shared machines.

The current location also doesn't match the [filesystem hierarchy standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html), it should be in `/usr/share` or `/usr/local/share`.

Another option would be to use the [embed package](https://pkg.go.dev/embed) so you don't have to distribute other files with the `containerlab` binary at all.

hellt:
I think it makes sense for the static artifacts used for graphing to enjoy the embed package.

Today, for the dev purposes, or multi-user purposes where users might want to leverage different export templates, the `--template` flag can be used https://containerlab.dev/cmd/graph/#render-graph-using-a-custom-html-template

hellt:
superseded by #2169

PR #2169, opened 2024-08-18
Embed templates into binary

remram44:
I can't tell what `templates/graph/exp/` is for.

There is also no way to use `full.tmpl` rather than `auto.tmpl` in the current setup


clab/export.go
```diff
@@ -41,10 +42,18 @@ type TopologyExport struct {
...
 // exportTopologyDataWithTemplate generates and writes topology data file to w using a template.
 func (c *CLab) exportTopologyDataWithTemplate(_ context.Context, w io.Writer, p string) error {
-	n := filepath.Base(p)
-	t, err := template.New(n).
+	var n string
+	if len(p) != 0 {
+		n = filepath.Base(p)
+	} else {
+		n = "auto.tmpl"
+	}
```
hellt, line 55: I think, in this case, the template name can be unconditionally named "export" without having the if/else branching.

clab/export.go
```diff
@@ -55,8 +64,13 @@ func (c *CLab) exportTopologyDataWithTemplate(_ context.Context, w io.Writer, p
...
 			},
-		}).
-		ParseFiles(p)
+		})
+	var err error
+	if len(p) != 0 {
+		_, err = t.ParseFiles(p)
+	} else {
+		_, err = t.Parse(defaultExportTemplate)
+	}
```
hellt, line 73: To let people use the embedded full.tmpl config, let's say that if a user provides a string value `__full` as the template path, this would trigger the usage of the full.tmpl

hellt:
thanks @remram44

PR #2187, opened 2024-09-03
Don't crash if encoded Docker password contains +

remram44:
This is a different bug from #2167. The decoding is done using URL encoding (`A-Za-z-_`) instead of standard encoding (`A-Za-z+/`).

This would cause a crash if a Docker password happened to encoding to a string with a `+` or `/`, as was my case with my quay.io account:

```
$ containerlab deploy
INFO[0000] Containerlab v0.57.0 started                 
INFO[0000] Parsing & checking topology file: frr.clab.yml 
Error: illegal base64 data at input byte 15
```


hellt:
appreciate!

PR #2199, opened 2024-09-20
removing whitespaces and linefeed from NodeType parameter.

flosch-dev:
When using custom-sros-variant in clab topology definition, the trailing \n is added as parameter to vrnetlab. Since https://github.com/srl-labs/containerlab/pull/2012 the parameter is escaped with %q which result in "n" is added as NodeType parameter to vrnetlab. vrnetlab also passes the same to vsim instannce which always fails to boot with incorrect type.

Not sure if this is best way to fix this issue. As a workaround, I added white space at the end of my type definition in topology yaml.

### DIFF
--- a/types/topology.go
+++ b/types/topology.go
@@ -1,10 +1,11 @@
 package types
 
 import (
+	"strings"
+
 	"github.com/docker/go-connections/nat"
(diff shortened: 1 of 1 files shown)


hellt:
thnx

PR #2208, opened 2024-09-29
added datapath field for SR Linux SXR

hellt:


hellt:
this datapath is not available in the container

PR #2211, opened 2024-09-30
Add Cisco IOL kind

kaelemc:
Relevant to https://github.com/hellt/vrnetlab/pull/256. This adds Cisco IOL as a kind in containerlab.

The kind is name `cisco_iol` let me know if you want this changed. I've also created the node as a `VRNode` type, I figured that'd be the easiest way to get the nice interface names.

IOL also doesn't support NETCONF so configuration saving isn't possible by those means, but the user is able to perform `copy run start` or `write memory` locally on IOL and it'll save to the nvram (which is stored under the labdirectory). That way config persistence is achieved.

Still working on documentation and other things.


hellt:
Thanks @kaelemc!

nodes/iol/iol.go
```diff
@@ -0,0 +1,70 @@
...
+// Register registers the node in the NodeRegistry.
+func Register(r *nodes.NodeRegistry) {
+	r.Register(kindnames, func() nodes.Node {
+		return new(iol)
+	}, defaultCredentials)
+}
+
+type iol struct {
+	nodes.VRNode
+}
```
hellt, line 36: @kaelemc somehow these iol images do not work for me.

This is what I get in the logs

```
❯ docker logs -f clab-iol-switch 
2024-10-03 10:42:12,883 INFO Set uid to user 0 succeeded
2024-10-03 10:42:12,887 INFO RPC interface 'supervisor' initialized
2024-10-03 10:42:12,887 INFO supervisord started with pid 1
2024-10-03 10:42:13,890 INFO spawned: 'entrypoint' with pid 7
2024-10-03 10:42:14,892 INFO success: entrypoint entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2024-10-03 10:42:14,892 INFO exited: entrypoint (exit status 0; expected)
2024-10-03 10:42:24,154 INFO spawned: 'iol' with pid 39
2024-10-03 10:42:24,156 INFO success: iol entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2024-10-03 10:42:24,222 INFO reaped unknown pid 8 (exit status 120)
```

I believe you don't have the process reaped as I do?

hellt: also I noticed that interfaces of the container do not have v4 addresses
something you remove in the entrypoint?

```
root@router2:/opt/iol# ifconfig 
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet6 2001:172:20:20:42:acff:fe14:1405  prefixlen 64  scopeid 0x0<global>
        ether 02:42:ac:14:14:05  txqueuelen 0  (Ethernet)
        RX packets 359  bytes 39937 (39.0 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 102  bytes 13978 (13.6 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```

hellt: as discussed in discord:
- ip addresses are claimed by the IOL, hence they are not visible in the default netns

kaelemc:
@hellt IOL type has been added, can be either `iol` (default) or `l2`. The configuration will change accordingly.

I've also made it so interfaces can now be defined in a non-contiguous manner. For example you can now have something like this
```yaml
links:
  - endpoints: ["router:Ethernet1/2", "switch:e0/4"]
```

previously interfaces had to be added sequentially (if you wanted to use e0/4 it meant that e0/1, e0/2, e0/3 had to be defined first).

I have one problem. IOL interfaces can only have 4 interfaces per slot. That means if I wanted to use `eth6` that'd  be `e1/2`. The regex for interface aliasing doesn't seem to support the slot formatting but only `<port>`. So technically the user could define `e0/6` but that wouldn't match with the CLI.

kaelemc:
I think we should stick with the contiguous interface naming requirement even though the code can handle otherwise. Even if a user uses a random interface without declaring/using the previous, all interfaces will show up in the CLI anyways which might just create extra confusion.

The regex issue is still a problem as soon as a user wants more than 3 data-plane interfaces in their topology.

kaelemc:
I'm not sure if this was the right way to do it but interfaces are assigned to the container following `ethX` schema.

`GetInterfaceMapping()` (heavily copied from the SRL definition) will map the interface slot/port into an `ethX` based endpoint. For example `e1/2` becomes `eth6`.

Interfaces are assigned in groups of four so each slot has max 4 interfaces, because of this the slot is multiplied by 4, the port index is added and then we add them and get out `ethX` interface index.

The topology will still only accept the CLI style interface names and NOT `ethX` interface naming.

I have done my 'famous' OSPF adjacency test using the following topology:
```yaml
name: iol
topology:
  nodes:
    r1:
      kind: cisco_iol
      image: vrnetlab/cisco_iol:17.12.01
    r2:
      kind: cisco_iol
      image: vrnetlab/cisco_iol:17.12.01
    sw1:
      kind: cisco_iol
      image: vrnetlab/cisco_iol:l2-17.12.01
      type: l2
  links:
    - endpoints: ["r1:e0/1", "sw1:e0/1"]
    - endpoints: ["r2:e0/1", "sw1:e1/2"]
 ```

Edgeshark shows the mapped interface
image

OSPF verification:
image

PR #2219, opened 2024-10-04
fix: updates JSON schema for publish to accept udp

nlgotz:
Adds udp to the options for the publish command in Docker.

Without this, publishing any UDP based services (DNS, SNMP, etc.) causes a validation error.


nlgotz:
Going to cleanup separately

hellt:
@nlgotz before you jump on it, I think you don't need `publish`
the `publish` stuff was done for border0 initially, something that exposes the lab via a public tunnel, this is not mapped docker' expose field

Issue #2225, opened 2024-10-09
Variables defined in `cc.clab_vars.yml` file are not parsed on `destroy` ?


adam-kulagowski:
Hi,

Assuming the following node definition in `cc.clab.yml`:

```
[...]
    CE-red:
      kind: linux
      labels:
        device-type: network-device
      image:  {{ $.registry_url }}/vyos:202409160007
      mgmt-ipv4: 172.24.0.100
      binds:
        - config/vyos/ce-red.boot:/opt/vyatta/etc/config/config.boot
      cmd: /sbin/init
```
And following `cc.clab_vars.yml` content:
```
registry_url: "[valid url to local container repository]"
```
The `sudo container deploy` goes nicely. All resources gets created. However the `sudo containerlab -d destroy` ends with:
```
DEBU[0000] trying to find topology files automatically  
DEBU[0000] topology file found: cc.clab.yml             
DEBU[0000] env runtime var value is                     
DEBU[0000] Running runtime.Init with params &{Timeout:2m0s GracefulShutdown:false Debug:false KeepMgmtNet:false VerifyLinkParams:<nil>} and &{Network: Bridge: IPv4Subnet: IPv4Gw: IPv4Range: IPv6Subnet: IPv6Gw: IPv6Range: MTU:0 ExternalAccess:<nil>} 
DEBU[0000] Runtime: Docker                              
DEBU[0000] initialized a runtime with params &{config:{Timeout:120000000000 GracefulShutdown:false Debug:false KeepMgmtNet:false VerifyLinkParams:0xc0007e9e7a} Client:0xc000a68460 mgmt:0xc00077e090} 
Error: yaml: line 73: did not find expected key, failed to parse topology file
```

Line 73 points to one line before the `image: ` value.

Adding `--vars` option do not help. 

**Speculation:** It looks like the `vars` file, upon destruction is loaded way later in the code. The working debug output (without using the `$.registry_url` looks like this:

```
[fidor@Marvin-PC clab]$ sudo containerlab -d destroy -t cc.clab.yml --vars cc.clab_vars.yml 
DEBU[0000] env runtime var value is                     
[...]

PR #2226, opened 2024-10-10
Dedup TopoFileLoading

steiler:
Related to #2225 this dedups the code for loading topologies, where we've had a seperate implementation for the Destroy action.


hellt:
great catch. ty!

PR #2227, opened 2024-10-10
N9kv docs refresh

hellt:
* add a note on 9kv-lite and mem tuning
* use new block formatting
* use template variables


rohit04saluja:
When will the refresh on the website happen with this change?

hellt:
With the next release. Likely next week

Issue #2233, opened 2024-10-14
Need a nerd knob to disable auto-restart policy of containers of Linux kind


ipspace:
Linux containers (which include FRRouting nodes) are restarted after a server reboot, but other containers are not, usually resulting in a broken lab. The configuration is gone anyway, and the connectivity (veth pairs) is most probably missing.

Worst case, if a Linux-based container manages to crash your server (if anyone can explain why I'm experiencing that I'll be forever grateful), you could get stuck in an infinite loop and own an expensive brick.

I understand the reasoning behind restarting Linux containers, but it would be great if that would be configurable. Please note that this is different from the **auto-remove** option. I would love to see why a container crashes (great for troubleshooting), I would just need a mechanism to influence its restart.

Finally, FWIW, it looks like https://github.com/srl-labs/containerlab/blob/794e3cb36401bedd2fae73816963c42661fde371/runtime/docker/docker.go#L506 sets **on-restart** policy to **on-failure**, but Docker nonetheless restarts them after the system crash (contrary to https://docs.docker.com/engine/containers/start-containers-automatically/#use-a-restart-policy)

steiler:
Oh no, again, kind specific config in a general purpose struct...
I'm trying hard since quite a while to erase those types of polution.

steiler:
Hi @ipspace,
well seeing the code I am sad again, having these kind specific knobs in the general runtimes is just ugly and I try since quite a while to get rid of them (although a blame shows my name besides that line .... which I think is due to moving the func around).
I have a pretty hacky solution for you.
Check it out and let me know if that works fine for you or not.
```
mava@server01:~/projects/containerlab3$ cat lab-examples/linux.clab.yml
name: linuxfrr

topology:
  kinds:
    linux:
      image: alpine
  nodes:
    frr1:
      kind: linux
    frr2:
      kind: linux
    host:
      kind: host
      stages:
        configure:
          wait-for: 
          - node: frr1
            stage: configure
          - node: frr2
            stage: configure
          exec:
            on-enter:
              - docker update --restart=on-failure clab-linuxfrr-frr1
              - docker update --restart=on-failure clab-linuxfrr-frr2
mava@server01:~/projects/containerlab3$ 
mava@server01:~/projects/containerlab3$ sudo clab deploy -t /home/mava/projects/containerlab3/lab-examples/linux.clab.yml --reconfigure
INFO[0000] Containerlab v0.57.2 started                 
[...]

PR #2238, opened 2024-10-15
Adding HostExec for stages and RestartPolicy

steiler:
Adding the RestartPolicy as a knob, also cleaning up the docker runtime from referencing specific node kinds.
Further implements host-exec per stage, which executes commands in the host namespace as part of the stages of a specific node.
Tackling the two topics that surfaced in #2233 

```
name: linuxfrr

topology:
  kinds:
    linux:
      image: alpine
  nodes:
    frr1:
      kind: linux
      stages:
        configure:
          host-exec:
            on-enter:
              - docker update --restart=no clab-linuxfrr-frr1
    frr2:
      kind: linux
    frr3:
      kind: linux
      restart-policy: no
```

The json schema update is missing so far, as well as docs update.

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -197,6 +197,7 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 		Memory:          c.Config.Topology.GetNodeMemory(nodeName),
 		StartupDelay:    c.Config.Topology.GetNodeStartupDelay(nodeName),
 		AutoRemove:      c.Config.Topology.GetNodeAutoRemove(nodeName),
+		RestartPolicy:   c.Config.Topology.GetRestartPolicy(nodeName),
 		Extras:          c.Config.Topology.GetNodeExtras(nodeName),
 		DNS:             c.Config.Topology.GetNodeDns(nodeName),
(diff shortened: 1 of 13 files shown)


hellt:
thanks @steiler 
I have added docs/schema/tests and did one slight modification by putting the host execs target ahead of the container exec.

The idea is that host execs are typically preparation commands, so they would benefit form running first, and then the container stage execs would follow

steiler:
> The idea is that host execs are typically preparation commands, so they would benefit form running first, and then the container stage execs would follow

I've also thought about orders. The other option would be to reorg that part and make it more like
```
exec:
  - command: <command>
    target: <container/host>
```
That would allow to run them in any order ...

hellt:
@steiler 

> > The idea is that host execs are typically preparation commands, so they would benefit form running first, and then the container stage execs would follow
> 
> I've also thought about orders. The other option would be to reorg that part and make it more like
> 
> ```
>   nodes:
>     frr1:
>       kind: linux
>       stages:
>         configure:
>           exec:
>             on-enter:
>               - command: <command>
>                 target: <host/container>
> ```
> 
> That would allow to run them in any order ...

This is way better, I agree
I merged the current version to save the current state, but if you know how to implement this - it would be much cleaner

PR #2239, opened 2024-10-17
Ensure IOL nodes have unique system base MACs

kaelemc:
IOL generates the base system MAC based on the PID that the IOL binary is launched as. 

Previously IOL nodes all used a static PID of 1. This meant multiple IOL switch nodes would have the same (duplicate) base MAC which means things like STP all have the same system ID.

This change uses the node index so that each launched IOL node has a unique PID. An environment variable `IOL_PID` is set to this node index which the IOL container entrypoint uses to launch IOL with that environment variable as the PID number.

This ensures the base system MAC is unique per each IOL node.

Relevant vrnetlab PR: https://github.com/hellt/vrnetlab/pull/270

This PR also changes the MAC OUI to be in the IANA assigned [documentation MAC space](https://www.iana.org/assignments/ethernet-numbers/ethernet-numbers.xhtml).

### DIFF
--- a/nodes/iol/iol.go
+++ b/nodes/iol/iol.go
@@ -40,7 +40,7 @@ var (
 	IOLCfgTpl, _ = template.New("clab-iol-default-config").Funcs(
 		gomplate.CreateFuncs(context.Background(), new(data.Data))).Parse(cfgTemplate)
 
-	IOLMACBase = "00:53:01"
+	IOLMACBase = "00:00:5E"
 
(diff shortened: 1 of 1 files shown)


kaelemc:
Sorry I made an error, I read the RFC and the defined documentation space uses the generic IANA OUI of `00:00:5e`. The MAC I had in there was wrong and used the low-order part of the MAC which is incorrect. 

The commit I just made changes the OUI to `00:00:5e` which fixes that.

PR #1944, opened 2024-03-15
Decouple `cmd` package from the `clab` application part2

steiler:
continuation of #1906

PR #2244, opened 2024-10-17
Move all Kinds into the package kinds

robertvolkmann:
Continuation of #1944


hellt:
Hi @robertvolkmann 
thanks for spending time on this
what do you achieve with packing kinds under `kinds/types`?

robertvolkmann:
You're right, I can also move them one level up. In the end, the registration of the kinds is no longer part of the `clab` package and is now independent of the `node` package as well. I followed the idea from @steiler's draft.

hellt:
I think the end goal is twofold

1. make clab package importable so that it can be called as an api (high level goal)
2. make sure that the `cmd` pkg doesn't contain any specifics that drive the clab package

The clab package registering kinds is OK, because it does not prevent points 1 and 2 above to be achieved.

steiler:
Yes, so in #1944 the plan was to have a specific kind_registry that you can query with kind name and retrieve additional information about that kind. One of the issues was e.g. the creadentials, they are set as a global var in the <kind> package.
By creating the kind_registry the idea was to use that as a central point where the kinds would register themselfes and also provide additional infos that can be queried from it in the deployment or and other of the workflows.

PR #2256, opened 2024-10-24
Add network aliases to node

mzagozen:
A node may have additional network aliases defined in the topology file, like so:

```yaml
topology:
  nodes:
    srl1:
      kind: srl
      aliases:
        - srl1-alias1
```

This is directly influenced by docker-compose
aliases (https://docs.docker.com/reference/compose-file/services/#aliases). The difference is that containerlab aliases are defined on a per-node basis, not per-network.

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -202,6 +202,7 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 		DNS:             c.Config.Topology.GetNodeDns(nodeName),
 		Certificate:     c.Config.Topology.GetCertificateConfig(nodeName),
 		Healthcheck:     c.Config.Topology.GetHealthCheckConfig(nodeName),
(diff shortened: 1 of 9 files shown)


hellt:
thanks @mzagozen 
what is the use case for the name alias?

mzagozen:
> thanks @mzagozen what is the use case for the name alias?

We have a controller in our topology where the tenant is determined by the domain name used to connect to the controller. So we set up aliases like `tenant-1.example.com`, `tenant-2.example.com`. We then use these names from our NMS which is also deployed with containerlab in the same network. Of course this could be solved by adding `/etc/hosts` entries in the NMS container, but since the container runtime already supports aliases, it feels better to use it?

hellt:
thnx @mzagozen 
no, I agree that they are useful, I was just making sure you know that the nodes are accessible by their shortnames among each other.
But since you have custom expected fqdns, then I can see how having those is beneficial.
Thank you for the PR, I will get to it next week

PR #2262, opened 2024-10-26
Remove IOL MAC generation

kaelemc:
IOL generates it's own hardware MAC addresses for its interfaces. It turns out we can use these instead of generating our own MAC addresses.

This PR removes:

- The MAC address generation in containerlab.
- references in the template config to any generated MACs.
- Any other variables/structs vars which store the generated MAC addresses.


kaelemc:
Tested this and it works, we can wait for @DanPartelly to test as well.

DanPartelly:
Tested it as well. Working as intended.

PR #2263, opened 2024-10-26
Add secondary IOL NVRAM bind.

kaelemc:
IOL actually has two locations where it stores the NVRAM (essentially the saved configuration files). 

Previously only one of the locations was bound from the container to the lab directory.

This PR adds the second NVRAM location as a bind to the container (as well as the logic check if we have to create the file so that we don't overwrite it).

### DIFF
--- a/nodes/iol/iol.go
+++ b/nodes/iol/iol.go
@@ -28,6 +28,10 @@ import (
 const (
 	typeIOL = "iol"
 	typeL2  = "l2"
+
+	iol_workdir = "/iol"
+	nvram_file1 = "nvram_00001"
+	nvram_file2 = "nvram_00003"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


nodes/iol/iol.go
```diff
@@ -95,7 +95,8 @@ func (n *iol) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
 
 	n.Cfg.Binds = append(n.Cfg.Binds,
 		// mount nvram so that config persists
-		fmt.Sprint(path.Join(n.Cfg.LabDir, "nvram"), ":/iol/nvram_00001"),
+		fmt.Sprint(path.Join(n.Cfg.LabDir, "nvram_00001"), ":/iol/nvram_00001"),
+		fmt.Sprint(path.Join(n.Cfg.LabDir, "nvram_00003"), ":/iol/nvram_00003"),
```
hellt, line 99: @kaelemc would you mind introducing constants here for nvram1 and nvram2?

kaelemc: Fixed.

nodes/iol/iol.go
```diff
@@ -28,6 +28,10 @@ import (
 const (
 	typeIOL = "iol"
 	typeL2  = "l2"
+
+	iol_workdir = "/iol"
```
hellt, line 32: this seems to be not used?

kaelemc: @hellt Lines 102 & 103

PR #2264, opened 2024-10-27
Make IOL NVRAM filename generate based on IOL PID

kaelemc:
Sorry @hellt, #2263 was not truly a fix.

It turns out, yet again the PID is related to something IOL generates. 

This PR changes the behaviour so the NVRAM filename is generated dynamically based on the node PID (which is the index of the node + 1, since index starts from 0).

I have tested with multiple nodes, and it works (ignore that I said it worked for #2263 😛).

@DanPartelly Maybe a test to confirm I haven't missed anything? Thanks in advance.

### DIFF
--- a/nodes/iol/iol.go
+++ b/nodes/iol/iol.go
@@ -30,8 +30,6 @@ const (
 	typeL2  = "l2"
 
 	iol_workdir = "/iol"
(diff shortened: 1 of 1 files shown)


hellt:
Thank you @kaelemc !

PR #2279, opened 2024-11-05
invoke key update, closes #2278

### LINKED ISSUE #2278: Fix GitHub CLI GPG key expiration in older clab-devcontainer releases
**Description**
Users working with older releases of clab-devcontainer (e.g., version 0.56.0) are encountering issues due to an expired GitHub CLI GPG key. When these older containers are launched in environments like GitHub Codespaces, particularly when the devcontainer.json includes commands that use `apt-get`, the container fails to initialize properly and falls back to recovery mode.

The error appears as:

```
GPG error: https://cli.github.com/packages stable InRelease: The following signatures were invalid: EXPKEYSIG 23F3D4EA75716059 GitHub CLI [opensource+cli@github.com](mailto:opensource+cli@github.com)
E: The repository 'https://cli.github.com/packages stable InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
```

**Impact**
- Affects older releases of clab-devcontainer (<=0.56.0)
- Causes Codespaces to fall back to recovery mode when initializing these older containers
- Particularly impacts users who need to maintain or work with older versions for compatibility reasons

**Current Implementation**
In these older releases, the Dockerfile sets up the GitHub CLI repository using:
```dockerfile
RUN sudo mkdir -p -m 755 /etc/apt/keyrings \
    && wget -qO- https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo tee /etc/apt/keyrings/githubcli-archive-keyring.gpg > /dev/null \
    && sudo chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null```
```
**Proposed Solution**
Update the Dockerfile in a new release to use the githubcli-archive-keyring package for managing GPG keys:

``` # Install the githubcli-archive-keyring package to handle key rotations
RUN apt-get update && apt-get install -y --no-install-recommends \
    github-cli \
    githubcli-archive-keyring
```

FloSch62:
Closes #2278 

I have not found a better solution to keep the gpg keys updated/rotated automatically.
The gpg keys from the current release are valid until 2026, but with PR it ensures older Docker images of the clab-devcontainer can still function correctly even if the GPG key for the GitHub CLI repository has expired. The goal is to make the container update the GPG key as needed without requiring users to rebuild the image or manually intervene, allowing apt update to succeed seamlessly.

By configuring apt to run /usr/local/bin/update-github-cli-key.sh before any update operation, its ensure that the GPG key is refreshed every time apt update is invoked. This happens automatically and does not require any manual intervention from the user.
If the key installed in the image is expired, the script will fetch the latest key from the GitHub servers, updating the keyring so that apt can verify the repository signatures correctly.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -5,11 +5,30 @@ ARG CLAB_VERSION
 RUN echo "deb [trusted=yes] https://netdevops.fury.site/apt/ /" | \
     tee -a /etc/apt/sources.list.d/netdevops.list
 
-# setup keyring for github cli
-RUN sudo mkdir -p -m 755 /etc/apt/keyrings \
-    && wget -qO- https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo tee /etc/apt/keyrings/githubcli-archive-keyring.gpg > /dev/null \
-    && sudo chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
-    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+# Create the update-github-cli-key.sh script
+RUN echo '#!/bin/bash\n'\
+'set -e\n'\
+'echo "Updating GitHub CLI GPG key..."\n'\
+'curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | \\\n'\
+'sudo dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg\n'\
+'sudo chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg\n' | \
+sudo tee /usr/local/bin/update-github-cli-key.sh > /dev/null
+
+# Make the script executable
+RUN sudo chmod +x /usr/local/bin/update-github-cli-key.sh
+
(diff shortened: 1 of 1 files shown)


hellt:
@FloSch62 I am actually thinking "screw this gpg thingy". All the bad things come from security :D 
Let's instead curl the latest binary from the releases instead and forget all that nonsense


something like

```dockerfile
RUN bash -c 'VERSION=$(curl -s https://api.github.com/repos/cli/cli/releases/latest | grep -o "\"tag_name\": \".*\"" | sed -E "s/.*: \"(.*)\"/\1/") && \
    CLEAN_VERSION=${VERSION#v} && \
    DOWNLOAD_URL="https://github.com/cli/cli/releases/download/${VERSION}/gh_${CLEAN_VERSION}_linux_amd64.tar.gz" && \
    curl -L "$DOWNLOAD_URL" | tar xz -C /tmp && \
    mv /tmp/gh_${CLEAN_VERSION}_linux_amd64/bin/gh /usr/local/bin/ && \
    chmod +x /usr/local/bin/gh && \
    rm -rf /tmp/gh_${CLEAN_VERSION}_linux_amd64'
```

FloSch62:
@hellt I think you are right, I applied your proposal and tested it.

PR #2285, opened 2024-11-09
Use default vCPU and RAM settings from vrnetlab

kaelemc:
2vCPU is frankly not enough for this node. Cisco docs say 4vCPU is the minimum required, with 4 vCPU I am seeing greatly improved boot times (still absurdly long though 😁).

https://www.cisco.com/c/en/us/td/docs/routers/virtual-routers/configuration/guide/b-xrv9k-cg/b-xrv9k-cg_chapter_011.html

### DIFF
--- a/nodes/vr_xrv9k/vr-xrv9k.go
+++ b/nodes/vr_xrv9k/vr-xrv9k.go
@@ -59,7 +59,7 @@ func (n *vrXRV9K) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
 		"USERNAME":           defaultCredentials.GetUsername(),
 		"PASSWORD":           defaultCredentials.GetPassword(),
 		"CONNECTION_MODE":    nodes.VrDefConnMode,
-		"VCPU":               "2",
+		"VCPU":               "4",
 		"RAM":                "16384",
 		"DOCKER_NET_V4_ADDR": n.Mgmt.IPv4Subnet,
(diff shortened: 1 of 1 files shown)


kaelemc:
Sounds good, so I should remove the env vars from the clab end so that they are only passed when the user manually sets them?

nodes/vr_xrv9k/vr-xrv9k.go
```diff
@@ -59,7 +59,7 @@ func (n *vrXRV9K) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
 		"USERNAME":           defaultCredentials.GetUsername(),
 		"PASSWORD":           defaultCredentials.GetPassword(),
 		"CONNECTION_MODE":    nodes.VrDefConnMode,
-		"VCPU":               "2",
+		"VCPU":               "4",
 		"RAM":                "16384",
```
kaelemc, line 63: As simple as removing these two lines I guess?

hellt: Yes, but first ensuring that the vrnetlab has these defaults set

kaelemc: Yep gotcha 👍

kaelemc:
I'll make the changes on the vrnetlab side (maybe after the scrapli work is finished) then make the changes on the containerlab end.

kaelemc:
Not relevant for me right now. Can handle this later.

PR #2291, opened 2024-11-12
graph drawio: Pull strategy and Interactive mode support

FloSch62:
This PR introduces two major improvements to the clab-io-draw functionality:

1. Automatic Image Updates
- When using the `latest` tag, containerlab now automatically checks and pulls 
  the newest version of clab-io-draw
- Users are informed about the update status through clear log messages
- Ensures users always have the most recent version when using `latest` tag

2. Native Interactive Mode Support
- Adds native support for the `-I` (interactive) mode directly in containerlab
- Users can now utilize the interactive 

```bash
# Start interactive cli ui
containerlab graph -t st.clab.yml --drawio --drawio-args="-I --theme nokia_modern_dark"
```


FloSch62:
@steiler I think I tried it when I originally added it but I was failing to get it work. If you want to move it to the docker lib, feel free to do it ;)

PR #2295, opened 2024-11-15
Add script which prevents XRd from having multiple IPv6 addresses on mgmt interface

### LINKED ISSUE #2277: XRd can have duplicate IPv6 addresses on management interface
Documenting here in case any other users run into the issue. Working on a fix..

- When redeploying a topology __with saved configs__ XRd can have multiple IPv6 addresses on it's mgmt interface. 
    - Saved configs as in: committed config within XR CLI.
- On redeploy the IPv6 address from the old deployment are present in the saved configs, they are not removed.
- The new management IPv6 address is simply added, unlike the IPv4 management address which overrides the old one in the config.
    - Have to find out the root cause, might be that the startup config loads the new mgmt ipv6 and the existing config has the old ipv6 and the startup config gets merged into the saved config.
- This means that XRd nodes have both their old and new management IPv6 address, meaning possible overlaps which could mean:
    - A different mapping between hostname and node itself (the wrong IPv6 address is replying, SSH prefers IPv6).
    - This could render some nodes unreachable since you can get different replies from duplicate addresses.

kaelemc:
Closes #2277.

This PR adds a basic script which removes all IPv6 address from the management interface (MgmtEth0/RP0/CPU0/0), and then assigns the correct management IPv6 address according to clab.

This fixes a problem that users have reported having of 'incorrect device mapping to ip address'. No work is required for IPv4 as the management interface is only allowed a single IPv4 address and XRd automatically overwrites the old one.

- Environment variable `XR_EVERY_BOOT_SCRIPT` points to the path to the script on the containers filesystem. When this env var is populated XRd runs the script at that path on well.. every boot 😀.
- An empty file for the script is created in the `createXRDFiles()` function, so that docker doesn't automatically create a folder for the bind mount. ACLs are also edited on the file since XR needs the file to be executable.
- Script populates IPv6 address via gomplate. This has to be done in PostDeploy since we don't know the container IPv6 address in PreDeploy (I think?).
- XRd executes the script AFTER first boot or startup configuration has been applied.

I've tested and ensured the script is working as it should and confirmed that *without* the script i'm getting the duplicate address behaviour.

### DIFF
--- a/nodes/xrd/mgmt_intf_v6_addr.sh.tmpl
+++ b/nodes/xrd/mgmt_intf_v6_addr.sh.tmpl
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+source /pkg/bin/ztp_helper.sh
+
+xrapply_string "interface MgmtEth0/RP0/CPU0/0\n no ipv6 address\n{{- if .MgmtIPv6Addr }}ipv6 address {{ .MgmtIPv6Addr }}/{{ .MgmtIPv6PrefixLen }}\n{{- end}}"
--- a/nodes/xrd/xrd.go
+++ b/nodes/xrd/xrd.go
@@ -5,6 +5,7 @@
 package xrd
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"fmt"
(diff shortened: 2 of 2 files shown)


hellt:
do you want someone else to test it first? or is it pretty much tested by you?

kaelemc:
I've tested it, I think it's good to go 👍

PR #2311, opened 2024-11-23
Change quickstart doc piece installation script to use the 'quick start' script.

kaelemc:
Previously the [Quickstart](https://containerlab.dev/quickstart/) page had the one-liner script that only installed containerlab only and not any of the dependencies.

I think the quicksetup script makes more sense instead of the clab-only install script. This PR does the following:

- Admonition on the install page under the quicksetup script is now of the 'tip' type (was 'info' before).
- Quickstart doc page was updated to use the proper quicksetup script (via snippets).
- Changed the text/explainer slightly to more accurately reflect the quicksetup script. The hyperlink now links to the quicksetup heading as well.

Let me know if you think the admonition should be changed back to info, or something that catches attention more like 'warning'. In fairness the content is important as if the user doesn't follow those instructions their docker install will not be working properly.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -30,14 +30,14 @@ The script officially supports the following OSes:
 * Rocky Linux 9.3, 8.8 (should work on any 9.x and 8.x release)
 
 To install all components at once, run the following command on any of the supported OSes:
-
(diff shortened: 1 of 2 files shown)


hellt:
thnx @kaelemc

PR #2347, opened 2024-12-13
Fix management interface and startup configs for IOL

kaelemc:
Relevant to IOL and the discussion in [Discord](https://discord.com/channels/860500297297821756/1316963741216014421/1317027972682612786).

### DIFF
--- a/docs/manual/kinds/cisco_iol.md
+++ b/docs/manual/kinds/cisco_iol.md
@@ -122,9 +122,61 @@ Ethernet0/2            unassigned      YES unset  administratively down down
 Ethernet0/3            unassigned      YES unset  administratively down down
 ```
 
+## Startup configuration
+
+When -{{ kind_display_name }}- is booted, it will start with a basic configuration which configures the following:
+
+- IP addressing for the Ethernet0/0 (management) interface.
+- Management VRF for the Ethernet0/0 interface.
+- Default route(s) in the management VRF context for the [management network](../network.md#management-network).
+- SSH server.
+- Sets all user defined interfaces into 'up' state.
+
+On subsequent boots (deployments which are not the first boot of -{{ kind_short_display_name }}-), -{{ kind_short_display_name }}- will take a few extra seconds to come up, this is because Containerlab must update the management interface IP addressing and default routes for the management network.
+
+### User-defined config
+
+-{{ kind_display_name }}- supports user defined startup configurations in two forms:
+
+- Full startup configuration.
+- Partial startup configuration.
+
+Both types of startup configurations are only be applied on the **first boot** of -{{ kind_short_display_name }}-. When you save configuration in IOL to the NVRAM (using `write memory` or `copy run start` commands), the NVRAM configuration will override the startup configuration.
+
+#### Full startup configuration
+
+The full startup configuration is used to fully replace/override the default startup configuration that is applied. This means you must define IP addressing and the SSH server in your configuration to access -{{ kind_short_display_name }}-.
+
+You can use the template variables that are defined in the [default startup confguration](https://github.com/srl-labs/containerlab/blob/main/nodes/iol/iol.cfg.tmpl). On lab deployment the template variables will be replaced/substituted.
+
+```yaml
+name: iol_full_startup_cfg
+topology:
+  nodes:
+    sros:
+      kind: cisco_iol
+      startup-config: configuration.txt
+```
+
+#### Partial  startup configuration
+
+The partial startup configuration is appended to the default startup configuration. This is useful to preconfigure certain things like loopback interfaces or IGP, while also taking advantage of the startup configuration that containerlab applies by default for management interface IP addressing and SSH access.
+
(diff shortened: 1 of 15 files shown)


kaelemc:
I've added a few things:

- Startup and partial configurations are now supported for IOL. Like SROS partial configs are supplied with the `.partial` file extension. If a startup config is supplied *without* .partial extension, then the config will override the default configuration we have, however like XRd and Cisco 8000 Emulator (and probably a few other nodes) this config can still take advantage of the template fields.

- The `UpdateMgmtIntf` function has been added, it runs in `PostDeploy()` with a 7 second delay. This takes advantage of a new `WriteToStdinNoWait` func that i've implemented only for the Docker runtime.

As per the name, `WriteToStdinNoWait` just blindly sends data to the containers stdin. In this case for IOL we send configuration to correct the management interface IP addressing. This works very very reliably for IOL, at this stage I don't think we need to scrape and send like in vrnetlab.

To refresh your memory, the reason we must do this is because in IOL the startup configuration file is ignored as soon as the nvram file exists (where user configuration is stored once a user does `write memory` or `copy run start`).

So the issue would be that:

1. A user starts a lab
1. A user SSHes into IOL, does some config and saves it (it gets saved to NVRAM)
1. On next deployment of the lab, IOL will load configuration from the NVRAM file.. which has the OLD management IP
1. User can't SSH into IOL properly since the management interface has the wrong IP address configured.

kaelemc:
@hellt Currently my change only alters the management IP on the mgmt interface. 

Is there a possible case at all where the management gateway or subnet could change without having to do a deployment with `--reconfigure`/`-c` flag?

If yes, I'll add configuration to update the static default route for the mgmt VRF.

hellt:
> Is there a possible case at all where the management gateway or subnet could change without having to do a deployment with `--reconfigure`/`-c` flag?

the eth0 interface of a container can change when you do a regular deploy operation. In other words, it doesn't matter if reconfigure is used or not, the IP might change if it becomes allocated to another node between two deploys

kaelemc:
@hellt Yes, understood. That is what has already been fixed in this PR.

What I mean to ask is if the entire subnet can change, hence constituting a change of the next hop IP for the static default mgmt route.

For example:

On initial deployment the subnet is 192.168.10.0/24 and the next hop for the static route is 192.168.10.1
Is there any case where on redeployment (without the `-c` flag) the subnet could change.. ie. to 192.168.20.0/24, thus meaning the next hop is now 192.168.20.1 (in this case the route in IOL must be updated).

Hope that explained it well enough :)

kaelemc:
Apologies, fixing the build errors as well, thought I had everything covered.

kaelemc:
@hellt Kindly requesting your help with the MockRuntime stuff. On this front I am out of my depth so I'll be a bit slow.

I didn't have access to my PC earlier, hence why I asked re: the mgmt subnet. But yes I have confirmed that the static route must be added as well since this case is possible.

kaelemc:
Ok, should be good to go (bar the tests).

Not proud of the 7 second delay. But this is an artifact of IOL having a 5 second boot delay. We can probably remove the 5sec boot delay (this is on the vrnetlab side).. but it means everyone has to rebuild their containers.

Also the `mgmt_str` is awfully long as the multi-line raw string didn't seem to play nice when writing to the stdin.

hellt:
Thanks @kaelemc 

I think to sleep well at nights we might want to add a short e2e test for iol startup config.
The testing orchestration is done via Robot Framework, and for iol it is defined here https://github.com/srl-labs/containerlab/tree/main/tests/10-basic-cisco_iol 

hit me up on discord if you want to have some introduction to the testing infra if you hit roadblocks

kaelemc:
I see you generated the mocks via the Makefile.. didn't know that it was that simple 😅.

Anyways i've added two fairly basic tests to confirm partial and full startup configurations are loading correctly. Full config also tests the template variable replacement. This is not supported in the partial config use case.

Tested it all locally on my machine, Everything seemed to work fine. 

Let me know if I got anything wrong.

hellt:
thanks @kaelemc 
tests look good. The last piece before the merge is to add the doc section that explains how the startup config is handled. You can check the sros docs for reference

hellt:
it seems that the `updateMgmtIf` function in the iol code base is not tested. What are the conditions for this code branch to run?

kaelemc:
It runs if it's **not** the first boot of IOL. 

If the NVRAM file already exists, we know this is not the first boot -- so update the management interface addressing.

kaelemc:
Documentation has been added regarding the startup configuration. Let me know if you see any improvements, clarifications or further info I should add/change.

I am not so familiar with the testing aspect, I am happy to take a crack at it, but will need some info & assistance. Is the codecov report based on the robotframework tests, or is there something else in play there?

How do you think we should handle this since for the mgmt intf IP addressing to be updated (the code branch in question, `updateMgmtIntf`), we deploy IOL lab -> destroy IOL lab -> deploy it again.

hellt:
thanks @kaelemc 
stellar work

PR #2351, opened 2024-12-17
adding no_proxy env variable by default to each clab node

### LINKED ISSUE #2350: Add NO_PROXY env var to the lab nodes by default
When http proxy is in place to access internet resources containerlab users typically have to set the http_proxy/https_proxy env vars in their topologies to make the applications running in the containers use the proxy:

```yaml
name: anysec-macsec

mgmt:
  network: anysec
  ipv4-subnet: 172.50.50.0/24

topology:
  kinds:
    linux:
      image: ghcr.io/srl-labs/network-multitool
      env:
        http_proxy: ${http_proxy}
        https_proxy: ${https_proxy}
        no_proxy: pe1,pe2,p3,p4,ce5,ce6,client7,client8,gnmic,prometheus,grafana,localhost,172.50.50.0/24
```

Note, how every lab node, lab mgmt network and localhost has to be called out in the `no_proxy` env var to ensure that the nodes when communicating with each other will not try to reach the proxy.

Containerlab should set these values automatically as a default env var set:

- all node names as provided in the topology
- mgmt network
- localhost
- 127.0.0.1
- *.local addresses

toweber:
Fix #2350 

If `no_proxy` or `NO_PROXY` variable is already set, existing values are kept, missing default values (127.0.0.1, localhost,...) are added and clab mgmt nets, all clab nodenames and IPs (if available) are added.
For the clab node, both `no_proxy` and `NO_PROXY` are set to the same value

Example for https://github.com/srl-labs/srl-telemetry-lab/
`no_proxy=localhost,127.0.0.1,::1,*.local,172.80.80.0/24,172.80.80.11,172.80.80.12,172.80.80.13,172.80.80.21,172.80.80.22,172.80.80.31,172.80.80.32,172.80.80.33,172.80.80.41,172.80.80.42,172.80.80.43,172.80.80.45,172.80.80.46,client1,client2,client3,gnmic,grafana,leaf1,leaf2,leaf3,loki,prometheus,promtail,spine1,spine2`

### DIFF
--- a/clab/config.go
+++ b/clab/config.go
@@ -260,8 +260,12 @@ func (c *CLab) createNodeCfg(nodeName string, nodeDef *types.NodeDefinition, idx
 	}
 
 	// add mgmt subnet range for the sake of completeness - some OS support it, others don't
-	noProxyList = append(noProxyList, c.Config.Mgmt.IPv4Subnet)
-	noProxyList = append(noProxyList, c.Config.Mgmt.IPv6Subnet)
+	if c.Config.Mgmt.IPv4Subnet != "" {
+		noProxyList = append(noProxyList, c.Config.Mgmt.IPv4Subnet)
+	}
+	if c.Config.Mgmt.IPv6Subnet != "" {
+		noProxyList = append(noProxyList, c.Config.Mgmt.IPv6Subnet)
+	}
 
 	// sort for better readability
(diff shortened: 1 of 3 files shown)


hellt:
Thanks @toweber 
great work!

PR #2363, opened 2024-12-24
docker go sdk for clab-io-draw

### LINKED ISSUE #2361: graph.go GenerateDrawioDiagram
GenerateDrawioDiagram runs the clab-io-draw container should using the official Docker SDK.

"github.com/docker/docker/api/types/container"
"github.com/docker/docker/client"

instead of using exec commands. With that it's truly supports interactive usage by setting OpenStdin/AttachStdin + TTY in the container config.
I already have it mostly working, some pieces are still missing. With that I can use css based tui's like Textualize

image

I will create a PR for it

FloSch62:
This closes #2361 

Using docker go sdk
Fix smoke test for drawio generation.

### DIFF
--- a/clab/graph.go
+++ b/clab/graph.go
@@ -19,6 +19,7 @@ import (
 	"syscall"
 
 	"github.com/awalterschulze/gographviz"
+	"github.com/google/shlex"
 	log "github.com/sirupsen/logrus"
 	e "github.com/srl-labs/containerlab/errors"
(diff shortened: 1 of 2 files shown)


FloSch62:
I was thinking to use the implemations from runtime/docker, but I was not able to use them, so I implemented the docker client in graph.go . It's a bit weird to have func forcePull in graph.go, but thats my current limit of my go skills...

PR #2373, opened 2025-01-02
Remove "spin up" from docs for `clab destroy`

Lachstec:
Just a minor fix regarding wording in the docs for the `destroy` subcommand. It currently says:

 ```With the global --topo | -t flag a user sets the path to the topology definition file that will be used to spin up a lab.```

That confused me a bit since I did not know if I were reading the docs for `deploy` or for `destroy`. This PR changes the docs for destroy so that the part with "spin up a lab" is removed.


hellt:
thanks!

PR #2381, opened 2025-01-08
quick-setup: allow running script without setup-sshd function

jklare:
* although the setup-sshd function seems useful for certain cases, it is generally unrelated to what the rest of the script does and strictly not needed for most people
* to allow people to run the script without also modifying the sshd config on their servers (which might be considered a security violation in some contexts), this patch introduces an additional environment variable "SETUP_SSH", which can be set to "false" (or any other value than "true") to disable running the setup-sshd function
* to not break backward compatibility and keep the current functionality the same, the default value for this variable is set to "true"


hellt:
Thanks, @jklare 
can you please make one slight change: `SETUP_SSH` -> `SETUP_SSHD`

jklare:
@hellt Thanks for the quick review, done!

PR #2382, opened 2025-01-08
nokia_sros: handle CRLF and tabs in partialCfg

jcpvdm:
SROS partial configs with CRLF (\r\n) or tabs (\t) can fail to apply. This patch adds automatic conversion to avoid the issue:
- Replace CRLF with LF.
- Replace tabs with 4 spaces.

### DIFF
--- a/nodes/vr_sros/vr-sros.go
+++ b/nodes/vr_sros/vr-sros.go
@@ -262,8 +262,10 @@ func (s *vrSROS) applyPartialConfig(ctx context.Context, addr, platformName,
 		return err
 	}
 
+	configContentStr := string(configContent)
+
 	// check file contains content, otherwise exit early
-	if strings.TrimSpace(string(configContent)) == "" {
+	if strings.TrimSpace(configContentStr) == "" {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


nodes/vr_sros/vr-sros.go
```diff
@@ -315,8 +317,14 @@ func (s *vrSROS) applyPartialConfig(ctx context.Context, addr, platformName,
 			}
 		}
 	}
+
+	// Replace CRLF with LF
+	configContentStr = strings.ReplaceAll(configContentStr, "\r\n", "\n")
+	// Replace tabs with 4 spaces
+	configContentStr = strings.ReplaceAll(configContentStr, "\t", "    ")
```
hellt, line 324: let optimize this bit with a single-pass replaces

```go
replacer := strings.NewReplacer(
    "\r\n", "\n",
    "\t", "    ",
)
configContentStr = replacer.Replace(configContentStr)
```

jcpvdm: thanks @hellt, also updated comments

PR #2384, opened 2025-01-08
json schema refactor.

asadarafat:
1. `additionalProperties`: false is used in more place to forbid extra fields and so we can have strict validation.
2. Removed `minItems` from scalar properties (e.g., for string or number). Because thing could go haywire as keyword `minItems` should be used for array, while the keyword `minItems` used in type string and number in previous schema
3. Unified enum values for "`mage-pull-policy` and `restart-policy` to avoid duplicates and different case variants.
4. Factored out complex regex pattern for ports and publish into separate definitions (port-pattern and publish-pattern).
5. Switched kinds from many hardcoded properties to a single patternProperties rule for flexibility in topology.kinds.
6. AllOf logic remains intact for kind-conditional type constraints, but ensured they don’t contradict each other.

### DIFF
--- a/cmd/tools_netem.go
+++ b/cmd/tools_netem.go
@@ -6,10 +6,12 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"net"
 	"os"
+	"os/exec"
 	"strconv"
 	"time"
 
@@ -24,6 +26,7 @@ import (
 	"github.com/srl-labs/containerlab/internal/tc"
 	"github.com/srl-labs/containerlab/links"
 	"github.com/srl-labs/containerlab/runtime"
+	"github.com/srl-labs/containerlab/types"
 	"github.com/vishvananda/netlink"
 )
 
@@ -35,6 +38,7 @@ var (
 	netemLoss       float64
 	netemRate       uint64
 	netemCorruption float64
+	netemFormat     string
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


schemas/clab.schema.jsonc
```diff
@@ -0,0 +1,924 @@
+{
+    "$id": "https://containerlab.dev/clab.schema.json",
```
hellt, line 2: @asadarafat what is this file for?

asadarafat: I assume you are referring to `schemas/clab.schema.jsonc`. It is essentially the same as `clab.schema.json`, but with added comments. The `.jsonc` format is supported by VSCode, allowing for better readability and enhanced documentation within the file.

hellt:
this is too big to be merged in one go
we need to raise it in smaller chunks

asadarafat:
Can you elaborate what do you mean by smaller chunk, perhaps with some example of the chunks?Cheers/AsadAm 31.01.2025 um 23:34 schrieb Roman Dodin ***@***.***>:﻿
this is too big to be merged in one go
we need to raise it in smaller chunks

hellt:
> Can you elaborate what do you mean by smaller chunk, perhaps with some example of the chunks?

the changes should be scoped to a smaller amount of existing schema, so it would be easier to test the changes.
Take one improvement out of many and create it as a PR instead of having all of them tucked together as an example

asadarafat:
Got it.Cheers/AsadAm 01.02.2025 um 00:26 schrieb Roman Dodin ***@***.***>:﻿

Can you elaborate what do you mean by smaller chunk, perhaps with some example of the chunks?

the changes should be scoped to a smaller amount of existing schema, so it would be easier to test the changes.
Take one improvement out of many and create it as a PR instead of having all of them tucked together as an example

PR #2387, opened 2025-01-11
Export ssh public keys under ssh-pub-keys in topo json

hellt:
For external systems acting on containerlab topologies and creating automations around the nodes spawned with Containerlab it is beneficial to export the SSH **public** keys that containerlab extracted during its run.

This will enable tools to read the topology json and potentially add those keys at the post-deploy stage using automation tools available to a user


FloSch62:
Cool, I can use that for the clab connector too

hellt:
yup, I did it for the NodeUser generation specifically

Issue #2307, opened 2024-11-21
Containerlab should be able to run as non-root user (without `sudo`) on supported platforms


vista-:
At the moment, Containerlab checks whether its running as root (uid==0) when running certain privileged commands, such as `clab deploy`:

https://github.com/srl-labs/containerlab/blob/aaee747c7a12cb323af1997471729dd912fdbc80/cmd/root.go#L70

It would be nice to be able to remove this requirement to allow users to run Containerlab instances without giving them sudo access.

## Motivation

There are some administrators who would like to give the ability to users to run Containerlab without necessarily granting them full sudo rights. They can, in theory, allow a given user group to perform sudo and run Containerlab, but it would be great if this could be done out of the box, without having to perform additional system administration tasks.

## Initial observations

- The user should be able to spin up Docker containers.
- The user needs to create network interfaces, adjust their attributes and such, which might involve poking `sysctl`s
- The user needs to be able to interact with nftables/iptables, including loading their respective kernel modules
- The user performs a write to `sysfs`, to enable LLDP on the Linux bridge created for management
- The user creates a directory (if it does not exist yet) and symlinks network namespaces from `procfs` into `/run/netns/`
- Container names and their assigned IPs are written to /etc/hosts
- Finally, the default system SSH config is expanded in `/etc/ssh/ssh_config.d/` with the node SSH information

## Possible solutions

Breaking this down with some potential solutions other than running Containerlab as root:
- The first requirement is satisfied by having the user running Containerlab in the appropriate docker user group (in most cases, `docker`)
- A `CAP_NET_ADMIN` capability set on the executable would give us the ability to manipulate network interfaces + `nftables`.
- `CAP_SYS_MODULES` capability lets us load the `ip_tables` and `ip6_tables` kernel modules, if they are not loaded yet.
- Using tmpfiles.d would give us the ability to ensure `/run/netns` directory always exists and is writable to all users
- Moving the SSH config from /etc/ssh/ssh_config.d/ to be under the user's home instead would help us separate containerlabs in a multi-user environment.

## Caveats

This still leaves a few unsolved issues, where there's no easy fix from what I can tell:
- The sysctl write allowing LLDP on Linux bridges
-  /etc/hosts
- The workaround for `/run/netns` requires an additional dependency to be installed, which also has to run as a system service

## Alternative workaround

[...]

PR #2408, opened 2025-01-17
Sudo-less operation

vista-:
This PR adds sudo-less operation to Containerlab. See #2307 for more information on this feature.

## Sudo-less?

The sudo-less operation is implemented via setting the SUID (or sticky) bit on the Containerlab binary. The changes include handling of SUID-based running of Containerlab, and properly grabbing the correct user for lab directory ACL/ownership management when not run with `sudo`. 

## Security?

Root privileges are not used where they are not needed, as Containerlab immediately drops to regular user privilege on startup, and only escalates to root privilege where needed (this is currently done on a per-command granularity).

For use-cases where sudo-less operation is required (for example, Containerlab hosts and labs shared between colleagues, or in education), it is not necessarily desired for every user to be able to perform privileged operations on containerlabs, after all, being able to spin up arbitrary containers is the same as granting `sudo` rights. However, operations that don't require such privileges, such as saving configs, or creating topology graphs might still be useful for these users.

This feature also includes an optional check for group membership of the `clab_admins` group:
- If the group does not exist, the check is bypassed
- If the group exists, a check is performed whether the user running Containerlab is a member of the `clab_admins` Unix group before escalating to root privileges

The `clab_admins` group is created and the user running the command is added to it by default by the upgrade script, the quick install script and the post-install script.

## Potential breakages

- `groupadd/usermod` might not work on the post-install script for Alpine without `shadow` utils installed -- this platform is not officially supported by Containerlab, but nfpm is still configured to emit an `apk` file.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -90,9 +90,12 @@ func preRunFn(cmd *cobra.Command, _ []string) error {
 	// setting output to stderr, so that json outputs can be parsed
 	log.SetOutput(os.Stderr)
 
-	err := common.DropRootPrivs()
-	if err != nil {
-		return err
+	// Rootless operations only supported for Docker runtime
+	if rt == "" || rt == "docker" {
+		err := common.DropRootPrivs()
+		if err != nil {
+			return err
+		}
 	}
 
 	return getTopoFilePath(cmd)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -20,15 +20,6 @@ import (
 	"github.com/srl-labs/containerlab/utils"
 )
 
-<<<<<<< HEAD
-=======
-const (
-	CLAB_AUTHORISED_GROUP = "clab_admins"
-	ROOT_UID              = 0
-	NOMODIFY              = -1
-)
-
->>>>>>> b1098918 (containerlab: Add shorthands for root UID and no-modify flags for readability)
 var (
 	debugCount int
 	debug      bool
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -90,9 +90,13 @@ func preRunFn(cmd *cobra.Command, _ []string) error {
 	// setting output to stderr, so that json outputs can be parsed
 	log.SetOutput(os.Stderr)
 
+	err := common.DropRootPrivs()
+	if err != nil {
+		return err
+	}
 	// Rootless operations only supported for Docker runtime
-	if rt == "" || rt == "docker" {
-		err := common.DropRootPrivs()
+	if rt != "" && rt != "docker" {
+		err := common.CheckAndGetRootPrivs(cmd, nil)
 		if err != nil {
 			return err
 		}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -20,6 +20,15 @@ import (
 	"github.com/srl-labs/containerlab/utils"
 )
 
+<<<<<<< HEAD
+=======
+const (
+	CLAB_AUTHORISED_GROUP = "clab_admins"
+	ROOT_UID              = 0
+	NOMODIFY              = -1
+)
+
+>>>>>>> b1098918 (containerlab: Add shorthands for root UID and no-modify flags for readability)
 var (
 	debugCount int
 	debug      bool
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -14,6 +14,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/srl-labs/containerlab/cmd/common"
 	"github.com/srl-labs/containerlab/cmd/version"
 	"github.com/srl-labs/containerlab/git"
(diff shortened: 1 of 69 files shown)


kaelemc:
Brilliant! sudoless working great on `deploy`/`destroy`/`redeploy`. Tested that the group check is working correctly as well.

> Windows/minGW! I'd appreciate if someone could test this out.

I see `mingw` in the `get.sh` but surely this isn't supported for clab on windows @hellt? The blame says it's from 5 years ago too. If I'm missing something here and it is supported, I'll happily test.

Let me know if there's anything specific you want tested.

hellt:
correct, windows is not supported. Windows users should use WSL

vista-:
I have added docs about sudoless operations and made the creation of the `clab_admins` group only happen on the first install/upgrade past v0.6.3.
This is done by creating a file under `/etc/containerlab/` to signal that the setup was already done.

This way, admins who would prefer to have no `clab_admins` group won't have it created on every upgrade.

utils/postinstall.sh
```diff
@@ -32,3 +32,9 @@ elif type "wget" > /dev/null 2>&1; then
     wget -T 2 -q -O /dev/null "$REPO_URL" || true
     exit 0
 fi
+
+if [ ! -f /etc/containerlab/suid_setup_done ]; then
+    groupadd -r clab_admins
+    usermod -aG clab_admins "$SUDO_USER"
+    touch /etc/containerlab/suid_setup_done
```
hellt, line 39: @vista- I wonder, maybe instead of creating an empty file in /etc/containerlab we can check that the `clab_admins` group exists?

get.sh
```diff
@@ -250,7 +250,15 @@ downloadFile() {
...
     echo "Preparing to install $BINARY_NAME ${TAG_WO_VER} into ${BIN_INSTALL_DIR}"
-    runAsRoot cp "$TMP_ROOT/$BINARY_NAME" "$BIN_INSTALL_DIR/$BINARY_NAME"
+    # If SUID setup is running for the first time
+    if [ ! -f /etc/containerlab/suid_setup_done ]; then
+        # Set SUID bit on the containerlab binary
+        runAsRoot install -m 4755 "$TMP_ROOT/$BINARY_NAME" "$BIN_INSTALL_DIR/$BINARY_NAME"
+        # Add clab admins group and add current user to it
+        runAsRoot groupadd -r clab_admins
+        runAsRoot usermod -aG "$SUDO_USER" clab_admins
+        runAsRoot touch /etc/containerlab/suid_setup_done
```
hellt, line 260: same comment as in https://github.com/srl-labs/containerlab/pull/2408/files#r1929715066

docs/install.md
```diff
@@ -333,5 +336,65 @@ or more globally:
...
+```
+# Set SUID bit on Containerlab binary
+sudo chmod u+s `which containerlab`
+# Create clab_admins Unix group
+sudo groupadd -r clab_admins
+# Add current user to clab_admins group
+sudo usermod -aG "$USER" clab_admins
+```
+
+Users who manage their Containerlab installation via `deb/yum/dnf` package managers will have the sudo-less functionality automatically enabled during the first upgrade from pre-`0.6.3` versions.
```
hellt, line 356: pre `0.63.0`

cmd/root.go
```diff
@@ -68,11 +78,84 @@ func init() {
...
-func sudoCheck(_ *cobra.Command, _ []string) error {
-	id := os.Geteuid()
-	if id != 0 {
-		return errors.New("containerlab requires sudo privileges to run")
+func checkAndGetRootPrivs(_ *cobra.Command, _ []string) error {
+	_, euid, suid := unix.Getresuid()
+	if euid != 0 && suid != 0 {
+		return fmt.Errorf("this containerlab command requires root privileges or root via SUID to run, effective UID: %v SUID: %v", euid, suid)
+	}
+
```
hellt, line 86: @vista- in https://github.com/srl-labs/containerlab/pull/2420 a light refactoring has been done to the version command which is now in its own package under `cmd`

Also, the sudoCheck moved to the `cmd/common` package. If you rebase this PR on the `main` and align with these changes it would be fantastic

utils/postinstall.sh
vista-: That wouldn't be an equivalent operation -- if someone wants to let all users use Containerlab, they can simply delete the group and the check is skipped. This way, they would have to delete the group after every upgrade.

docs/install.md
vista-: fixed

cmd/root.go
vista-: Rebased PR on the refactor

hellt:
I think what would be good is to change one of the integration tests to use sudo-less commands. Since the binary that we build for integration tests is built with make, we just need to add chmod command to it, I think, and then remove sudo from the smoke test suite

PR #2426, opened 2025-01-29
Remove gomplate

### LINKED ISSUE #2414: remove `gomplate`
Gomplate is a heavy beast that drags in some heavy deps we don't need. The original reason we added gomplate is because we wanted to leverage extra templating functions that gomplate comes with.

But I would entertain the idea of removing it and instead add those missing functions manually as we identify that we need them.

hellt:
fix #2414 

let's see if anyone used it


hellt:
@kaelemc can you give this PR a try with iol? I think the failures are caused by the flakiness in gh actions, because I don't think you used any gomplate templates in the iol template, at lest they don't jump out at me

kaelemc:
@hellt I took a look, and both IOL + XRd are generating templates correctly. I do however get this error (both on IOL and XRd)

image

kaelemc:
Adding one more thing. The XRd *configuration* template is broken (`n.GenerateConfig(nodeCfg.ResStartupConfig, cfgTemplate)`), Looks like the mgmt interface addressing is not being populated.

axxyhtrx:
Screenshot 2025-01-30 at 11 56 00
Screenshot 2025-01-30 at 11 56 59

Wow.... 
Just look to the size of binaries.

70M before and 36M after removing `gomplate`.

hellt:
@kaelemc looks like iol tests are back, and I haven't changed much...

Regarding the xrd I don't think this PR caused any trouble, since the xrd template did not use any of the gomplate functions...

kaelemc:
@hellt Don't mind me, I was just hallucinating 😆. XRd isn't even supposed to populating mgmt ip in the template config anyways. Looks like all is working well and good 👍

PR #2437, opened 2025-02-02
Add `inspect interfaces` command, refactor cmd package

vista-:
This PR adds the `inspect interfaces` subcommand, which can be used to list the network interfaces of deployed nodes of a lab. The command provides information about the operational state, type, and interface alias of interfaces of lab nodes.

This is a privileged Containerlab command, as it requires entering network namespaces, which only privileged users can perform.

It can either output with pretty-printed table or JSON formatting.
 
The `inspect` and the new `inspect interfaces` command has been moved into a sub-package of `cmd`.

This PR also comes with a large refactor of the `cmd` package, moving common (or shared) variables within the `common` package inside `cmd`.


kaelemc:
@vista- with `--format json` do you think we could hide the log messages to align with `clab ins -a -f json`?

This just makes parsing the json a little easier so we don't have to strip out any logging

hellt:
I think the idea with json output and log messages that we used in other commands was that log messages are directed to stderr and json was printed to stdout

hellt:
thanks a bunch @vista- 
lgtm

PR #2460, opened 2025-02-12
Feature: Enabling Promiscuous Mode in general on MacVlan interfaces

steiler:
The interface is set to Promisc On right before being pushed into the containers network namespace.


hellt:
thanks!

PR #2461, opened 2025-02-12
Feature: Introduce Oper-State flag per Link Endpoint.

steiler:
The oper-state can be set to up/down or enable/disable to defien the expected state of an endpoint after deployment. 

Example:
```
...
  links:
    - type: veth
      endpoints:
       - node: srl1
         interface: e1-15
         mac:  5C:80:B6:aa:aa:aa
       - node: srl2
         interface: e1-5
         mac:  5C:80:B6:bb:bb:bb
         oper-state: down                                                  # setting the oper state.
```

### DIFF
--- a/docs/manual/topo-def-file.md
+++ b/docs/manual/topo-def-file.md
@@ -248,15 +248,17 @@ The veth link is the most common link type used in containerlab. It creates a vi
 links:
   - type: veth
     endpoints:
-      - node: <NodeA-Name>                  # mandatory
-        interface: <NodeA-Interface-Name>   # mandatory
-        mac: <NodeA-Interface-Mac>          # optional
-      - node: <NodeB-Name>                  # mandatory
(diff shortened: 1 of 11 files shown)


hellt:
@steiler I am not sure this PR would work for the issue it is raised against as it sets the admin state of the underlying link, while I think what @jbemmel wanted to alter is the config setting in the cisco template based on a user input.

If we were to assume that `no shut` in cisco config on IOL has effect on the underlying link status, then it would bring interface up again. If it does not do that, then in order to set the interface up something would need to do an equivalent of `ip link set dev xxx up`

steiler:
ok, I see. Do you consider this usefull anyways? Or are we simply dropping this PR?

hellt:
Since there were no use cases (so far) that needed to have some links in the "down" admin-state I would propose to shelf this PR until there is a need to add this capability.

PR #2470, opened 2025-02-18
fix setup qemu action

ankudinov:
Setup-qemu action is based on tonistiigi/binfmt image that is broken after the latest release. Every ARM build will fail.
Check https://github.com/tonistiigi/binfmt/issues/240 for the details.
The workaround is to fix image at tonistiigi/binfmt:qemu-v7.0.0-28. Anything above is broken.


hellt:
thanks!

PR #2487, opened 2025-02-24
Always set the corruption field

FloSch62:
I do not see any reason why the setCorruption was untouched when >0 
That also meant, a set to 0 was not possible


hellt:
thnx!

Issue #2505, opened 2025-03-06
clab version upgrade does not work when proxy is needed


toweber:
Based on the error output it seems the upgrade command is not taking the proxy env vars into account

```bash
clab version upgrade --log-level debug
DEBU[0000] Changed running UIDs to UID: 123 GID: 123
DEBU[0000] Dropped root privileges
DEBU[0000] Group membership check passed
DEBU[0000] Changed running UIDs to UID: 0 GID: 123
DEBU[0000] Obtained root privileges
Failed to retrieve latest containerlab version
Failed to install containerlab
        For support, go to https://github.com/srl-labs/containerlab/issues
Error: upgrade failed: exit status 1
```

PR #2507, opened 2025-03-06
source the proxy address from environment variables

steiler:
implementing #2505


steiler:
@toweber can you try this out, please?

hellt:
thanks!

PR #2517, opened 2025-03-21
Add vr-openwrt kind with overlay bind mount for OpenWRT persistence

Takalele:
This PR adds a new `vr-openwrt` kind to Containerlab, based on the existing `rare` kind.

### Key behavior:
- Automatically bind-mounts the node's lab directory as `/overlay` inside the container
- This allows persistent configuration storage across lab restarts

### Implementation notes:
- If desired, a more generic `vr` kind with overlay support could be created
  → This would require changes in `vrnetlab.py` so that overlay images are always created in `/overlay` instead of `/`

### Related work:
- Requires updated `vrnetlab/openwrt` image that supports `/overlay` handling
- See related image PR: https://github.com/hellt/vrnetlab/pull/339
---

#### Example usage:
```yaml
topology:
  nodes:
    openwrt:
      kind: vr-openwrt
      image: vrnetlab/openwrt_openwrt:24.10.0
      
```

BR
Takalele

### DIFF
--- a/nodes/vr_openwrt/vr_openwrt.go
+++ b/nodes/vr_openwrt/vr_openwrt.go
@@ -13,7 +13,7 @@ import (
 	"github.com/srl-labs/containerlab/utils"
 )
 
-var kindNames = []string{"vr-openwrt"}
+var kindNames = []string{"openwrt"}
 
 const (
 	generateable     = true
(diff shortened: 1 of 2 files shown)


nodes/vr_openwrt/vr_openwrt.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/srl-labs/containerlab/nodes"
+	"github.com/srl-labs/containerlab/types"
+	"github.com/srl-labs/containerlab/utils"
+)
+
+var kindNames = []string{"vr-openwrt"}
```
hellt, line 16: thanks @Takalele 
could you rename the kind to just `openwrt`? Thanks.

Takalele: @hellt 
sure, youre welcome.

PR #2535, opened 2025-04-01
Add logging to CLI topo option resolution

### LINKED ISSUE #2525: Improve logging for topology file sources in nested directories
When deploying a lab using a specified topology file, the source of the resolved topology file (or lack thereof) is typically logged:
```bash
$ clab deploy -d -t nonexistent.clab.yaml
14:33:03 DEBU Changed running UIDs to UID: 1001 GID: 1001
14:33:03 DEBU Dropped root privileges
14:33:03 DEBU Group membership check passed
14:33:03 DEBU Changed running UIDs to UID: 0 GID: 1001
14:33:03 DEBU Obtained root privileges
14:33:03 INFO Containerlab started version=0.66.0
Error: failed to fetch http(s) resource: https://nonexistent.clab.yaml
```
---

However, when carrying out a similar `deploy` using a topology file that's in a nested directory, the error is far less helpful:
```bash
$ clab deploy -d -t folder/nonexistent.clab.yaml
14:34:57 DEBU Changed running UIDs to UID: 1001 GID: 1001
14:34:57 DEBU Dropped root privileges
Error: authentication required: Repository not found.
```

This can be confusing if the path was mistyped, as it sounds more like the container image pull failed instead of the resolution of the topology file.

Weirdly, the error seems to appear as normal using `../nonexistent.clab.yaml`, `../folder/nonexistent.clab.yaml`, or `folderA/folderB/nonexistent.clab.yaml`. It seems to only appear when checking one directory below the current one. It seems to have this behavior whether the folder exists or not (`nonexistent-folder/nonexistent.clab.yaml` or `existent-folder/nonexistent.clab.yaml`)

---
This seems like a two-part issue:
1. There should be some logging as to the finally resolved location for the topology file, e.g.
```bash
00:00:00 DEBU Using topology file found locally at: folder/existent.clab.yaml
```
or 
```bash
00:00:00 DEBU Failed to find topology file locally
00:00:00 DEBU Using topology file found remotely at: https://folder/existent.clab.yaml
```
 
2. Passing a path to the `--topo` flag that's one directory deep should be handled consistently.

matthewCmatt:
- Adds debug-level logging for the resolution logic for `--topo` CLI option
- Expands the CLI `--help` option description for `--topo`
- Fixes a typo I noticed

Fixes #2525

This shouldn't have any functional changes 🤞

### DIFF
--- a/clab/clab.go
+++ b/clab/clab.go
@@ -207,12 +207,14 @@ func (c *CLab) ProcessTopoPath(path string) (string, error) {
 
 	switch {
 	case path == "-" || path == "stdin":
+		log.Debugf("interpretting topo %q as stdin", path)
 		file, err = readFromStdin(c.TopoPaths.ClabTmpDir())
 		if err != nil {
 			return "", err
 		}
 	// if the path is not a local file and a URL, download the file and store it in the tmp dir
 	case !utils.FileOrDirExists(path) && utils.IsHttpURL(path, true):
+		log.Debugf("interpretting topo %q as remote URL", path)
 		file, err = downloadTopoFile(path, c.TopoPaths.ClabTmpDir())
 		if err != nil {
 			return "", err
@@ -222,6 +224,7 @@ func (c *CLab) ProcessTopoPath(path string) (string, error) {
 		return "", fmt.Errorf("provide a path to the clab topology file")
 
 	default:
+		log.Debugf("interpretting topo %q as file path", path)
 		file, err = FindTopoFileByPath(path)
 		if err != nil {
 			return "", err
--- a/cmd/inspect/inspect.go
+++ b/cmd/inspect/inspect.go
@@ -170,7 +170,7 @@ func toTableData(contDetails []types.ContainerDetails) []tableWriter.Row {
 }
 
 // getTopologyPath returns the relative path to the topology file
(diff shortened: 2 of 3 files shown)


clab/clab.go
```diff
@@ -207,12 +207,14 @@ func (c *CLab) ProcessTopoPath(path string) (string, error) {
...
 	switch {
 	case path == "-" || path == "stdin":
+		log.Debugf("interpretting topo %q as stdin", path)
 		file, err = readFromStdin(c.TopoPaths.ClabTmpDir())
 		if err != nil {
 			return "", err
 		}
 	// if the path is not a local file and a URL, download the file and store it in the tmp dir
 	case !utils.FileOrDirExists(path) && utils.IsHttpURL(path, true):
+		log.Debugf("interpretting topo %q as remote URL", path)
```
hellt, line 217: 👋 @matthewCmatt 

maybe also add the check that the string does not contain a dot char in the `utils.IsHttpURL` to make sure we don't treat `abc/def.clab.yml` as a file url?

matthewCmatt:
@hellt I looked into adding the dot exception, but got stuck trying to get the environment up and running on my Mac + devcontainers. Thanks for finishing this one up!

PR #2537, opened 2025-04-02
Schema improvments

FloSch62:
This change will make the stage-exec valid for:

```
      healthcheck:
        test: ["CMD-SHELL", "curl -f http://netbox:8000/api/ || exit 1"]
        timeout: 10
        retries: 100
      stages:
        create:
          wait-for:
            - node: postgres
              stage: create
            - node: redis
              stage: create
        healthy:
          exec:
            - command: bash /config/netbox-init.sh
              target: container
              phase: on-exit
      binds:
        - config/configuration.py:/config/configuration.py
        - config/netbox-init.sh:/config/netbox-init.sh
 ```
 
 Addionally, I added: 
 "additionalProperties": false
 
 which makes it stricter on the allowed items. For example
 
 ``` 
 mgmt:
  ipv5-subnet: 172.30.21.0/24
  
  or
  
    kinds:
    unknwon:
      image: ghcr.io/nokia/unknown:latest
  ```
  
  is than not allowed anymore.

### DIFF
--- a/schemas/clab.schema.json
+++ b/schemas/clab.schema.json
@@ -189,18 +189,24 @@
                     "uniqueItems": true
                 },
                 "env": {
-                  "type": "object",
-                  "description": "environment variables",
-                  "markdownDescription": "[environment variables](https://containerlab.dev/manual/nodes/#env)",
-                  "patternProperties": {
-                    ".*": {
-                    "anyOf": [
-                        { "type": "string" },
-                        { "type": "number" },
-                        { "type": "boolean" }
-                      ]
+                    "type": "object",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: schemas/clab.schema.json [hunk 1]
```diff
@@ -193,15 +193,16 @@
                     "description": "environment variables",
                     "markdownDescription": "[environment variables](https://containerlab.dev/manual/nodes/#env)",
                     "patternProperties": {
-                        ".+": {
+                        ".*": {
```
hellt, line 196: the change from `.+` to `.*` was required?

PR #2538, opened 2025-04-02
Add support for the 6WIND VSR

nicoharnois:
- Fix the if-wait script
- Add configurable parameters for the docker runtime to configure devices, capabilities and the shared memory size
- Add the support of [6WIND VSR](https://doc.6wind.com/new/vsr-3/latest/vsr-guide/) with a dedicated *6wind_vsr*  kind


hellt:
Thanks for a complete PR @nicoharnois

hellt:
I needed to push some things on top, but your branch won't allow it, so I created a new one that has your commits + mine on top
#2540

PR #2540, opened 2025-04-06
Add support for 6WIND VSR

hellt:
additions to #2538


hellt:
Thanks @nicoharnois 
I am merging this.

Once the vsr image is GA we should add an integration test for it to have some coverage

nicoharnois:
Thanks a lot Roman. Understood for the testing coverage, I will check this.

PR #2542, opened 2025-04-06
New template functions (`conv.Join`, `strings.Split`, `strings.ReplaceAll`, `conv.ToInt`)

hellt:
New template functions (copy from gomplate project):

- `conv.Join`
- `conv.ToInt`
- `strings.Split`
- `strings.ReplaceAll`


ubaumann:
Thanks a lot!

PR #2560, opened 2025-04-22
Allow 0-length port array

matthewCmatt:
In a `.clab.yaml` file, specifying a node `ports` list, the YAML schema hinting provided in VSCode gives an error for empty lists. For instance:

```yaml
topology:
  nodes:
    service1:
      kind: linux
      image: alpine:3.14
      ports: []
      entrypoint: /bin/bash -c "echo 'hello world'"
```

Gives a yaml-schema error:

```
Array has too few items. Expected 1 or more.yaml-schema: Containerlab
```

However, this is legal and will deploy properly in ContainerLab.

This MR reduces the limit to allow empty lists. We might consider reducing the limit for other lists defined in the schema, provided the behavior is as expected.

---

For context, Docker compose makes no such limitation on ports. Here's a snippet from the Docker Compose schema:

```json
  "ports": {
    "type": "array",
    "items": {
      "oneOf": [
        {
          "type": "number"
        },
        {
          "type": "string"
        },
        ...
      ]
    },
    "uniqueItems": true
  }
```

I ran into this as I was trying to use Environment Variables to conditionally expose ports, e.g.
```yaml
topology:
  nodes:
    service1:
      kind: linux
      image: alpine:3.14
      ports: ${EXPOSE_PORTS}
      entrypoint: /bin/bash -c "echo 'hello world'"
```

where `EXPOSE_PORTS` can be either `[]` or `["8080:80"]`. This was the only way I could use environment variables to get this behavior.


hellt:
thanks @matthewCmatt 

I was wondering indeed why you would need that, but an underfined env var makes sense to me. Thanks for the fix

PR #2562, opened 2025-04-25
docs/quickstart: Use latest and greatest SR Linux version

vista-:
At the moment, we use a topology with SRL 24.3.3 in the quickstart example, so anyone following the guide on Ubuntu 23+ will run into the syslog AppArmor issue.

Probably best to bump the version to latest in the topology, so anyone using the guide will get the latest version of SRL to play with.

Also updated the topology with a more intuitive interface alias.


vista-:
A follow-up PR to this will update the drawio diagram embedded in the quickstart guide.

hellt:
Let's fix it with 24.10.4 - our latest lts to guarantee the outputs

hellt:
@vista- now that you added the major version as a tag to the srlinux image you have `ghcr.io/nokia/srlinux:24.10` available that we can use in this PR

PR #2569, opened 2025-05-02
Add VPP containerlab kind

pimvanpelt:
This is a work in progress to add VPP to Containerlab.
It uses a docker image currently published on [docker.io](docker.io/pimvanpelt/vpp-containerlab) with tag latest.

It needs to configuration files generated:
1.   /etc/vpp/startup.conf which will contain a CLI prompt that contains the node ShortName
2.  /etc/vpp/clab.vpp which will be executed on the CLI after VPP starts - its purpose is to bind the vethparh from Linux into the Dataplane.

Where the end facing documentation will live, I've added a placeholder (developer) documentation. I will clean this up once we have a working PR.

This PR is currently draft!


pimvanpelt:
Replaced by #2571

PR #2571, opened 2025-05-03
Add VPP node kind

pimvanpelt:
This is a work in progress to add VPP to Containerlab. It uses a docker image currently published at [IPng's Gitea](https://git.ipng.ch/ipng/vpp-containerlab), pushed to git.ipng.ch/ipng/vpp-containerlab:latest

It needs one configuration file to be generated by go template:
1.     /etc/vpp/startup.conf which will contain a CLI prompt that contains the node `.ShortName`

The docker container generates `/etc/vpp/clab.vpp` by scanning veth pairs upon startup. See `files/container-init.sh` in the docker repo above. Where the end facing documentation will live, I've added a placeholder (developer) documentation. I will clean this up once we have a working PR.

This PR is currently draft! But it does work end to end.


hellt:
It is lights out and away we go. Thanks @pimvanpelt

PR #2574, opened 2025-05-07
lab-examples/vsrx01: junoser convert the set-format

junousi:
After https://github.com/hellt/vrnetlab/pull/209 it seems the set-format config is less relevant. To encourage users using up-to-date vrnetlab, replace it with identical config generated with junoser from the set-based data.


hellt:
thanks @junousi

Issue #2580, opened 2025-05-08
Env vars support in YAML anchored embedded startup-configs


junousi:
Hello! I was wondering, is this type of variable expansion out-of-scope for the [envsubst feature](https://containerlab.dev/rn/0.65/#env-vars-support-in-startup-configs) ?

```
topology:
  defaults:
    config:
      vars:
        rtr-config: &rtr-config |
          interfaces {
              ge-0/0/0 {
                  unit 0 {
                      family inet {
                          address ${RTR_CORE_CIDR:=172.16.0.3/24};
                      }
                  }
              }
              ge-0/0/1 {
                  unit 0 {
                      family inet {
                          address ${RTR_CUST_CIDR};
                      }
                  }
              }
...omitted...
  nodes:
    rtr1:
      env:
        RTR_CORE_CIDR: '172.16.0.1/24'
        RTR_CUST_CIDR: '1.2.3.1/24'
      kind: juniper_vjunosrouter
      image: vrnetlab/juniper_vjunos-router:24.2R1-S2.5
      startup-config: *rtr-config
    rtr2:
      env:
        RTR_CORE_CIDR: '172.16.0.2/24'
        RTR_CUST_CIDR: '5.6.7.1/24'
      startup-config: *rtr-config
```

Env vars are present:

```
clab1:~$ docker logs clab-mcast-dense-rtr1|grep RTR_
2025-05-08 18:04:46,718: vrnetlab   	INFO RTR_CORE_CIDR: 172.16.0.1/24
2025-05-08 18:04:46,718: vrnetlab   	INFO RTR_CUST_CIDR: 1.2.3.1/24
```

Env var contents don't end up in the rendered config, though the default value is honored:

```
clab1:~$ docker exec -it clab-mcast-dense-rtr1 /bin/bash -c 'grep -A3 "^    ge-0/0" config/startup-config.cfg'
    ge-0/0/0 {
        unit 0 {
            family inet {
                address 172.16.0.3/24;
--
    ge-0/0/1 {
        unit 0 {
            family inet {
                address $RTR_CUST_CIDR;
```

hellt:
Hi. The envs are taken from your host, not the container

junousi:
So IIUC what happens here:

```
clab1:~$ docker logs clab-mcast-dense-rtr1 | grep -E 'startup\-|START|RTR_CORE'
[...]

PR #2581, opened 2025-05-10
tests/02-basic-srl: Test node-specific env vars

junousi:
Instead of hardcoding system info location, try to retrieve it from node-specific env vars.

See also: #2580


junousi:
PR can be closed if this is not the intended use case for container namespace env vars.

PR #2588, opened 2025-05-13
Extended link config schema

pobk:
This PR does the following:

1. Moves some commonly used properties into reusable defitions (ipv4-addr, ipv6-addr, mtu, vars, labels) and references them in the original properties.
2. Creates a new endpoint element that represents the common endpoint for the link-configs
3. Creates link configs for each of the extended link configurations and moves the short-form into link-config-short.

Testing
This PR has not been extensivly tested against edge-case or complex topologies yet.


hellt:
thanks @pobk

PR #2589, opened 2025-05-13
DevContainer Dockerfile updated to prevent non-existent group error on build

pobk:
Updated DevContainer Dockerfiles to create the group before attempting to add the vscode user to the group to prevent non-existant group errors


.devcontainer/Dockerfile
```diff
@@ -44,7 +44,7 @@ RUN mkdir -p /etc/sudoers.d && echo 'Defaults env_keep += "SSH_AUTH_SOCK"' > /et
 
 # Add vscode user to clab_admins group so that it can run sudo-less clab commands
 # the group is created when clab is installed via the installation script
-RUN usermod -aG clab_admins vscode
```
hellt, line 47: the group creation happens when containerlab is installed via the installer script on ln 22 of this dockerfile

Issue #2504, opened 2025-03-05
Make `groups` work as they were designed to


hellt:
Containerlab topology definition file had a notion of a `group` [forever.](https://containerlab.dev/manual/nodes/#group) But we never implemented them in the way they were envisioned to be used.

A `group` should allow a user to define a common set of node-level parameters that a certain node can then inherit via the `group` reference.

For example, the below topology defines a group `leafs` and adds some node-specific parameters to it. Then a node in the topology references this group to adopt the specified parameters:

```yaml
name: group-demo
topology:
  groups:
    leafs:
      labels:
        someLabel: someValue
      type: ixrd3l

  nodes:
    node1:
      group: leafs

    node2:
```

`node1` will inherit `labels` and `type` as defined in a group, whereas `node2` will not.

evilhamsterman:
I'm just starting with container lab. I think rather than having groups defined in one place and then tag the nodes as part of the groups, why not have groups be subsets of nodes. This could also allow for multiple layers of inheritance and maybe even groups in groups. Kind of like how Ansible handles yaml inventories

```yaml
name: groups-demo
topology:
# General nodes
  nodes:
    node1:
      kind: linux

# Grouped nodes
  groups:
    lab1:
      kind: ceos
      labels:
        cloud: az
      nodes:
        lab1a:

    lab2:
      kind: nokia_srlinux
      groups:
        clients:
          nodes:
            lab2a:
            lab2b:
        spine:
           nodes:
             spine1:
               type: ixrd31
```

PR #2593, opened 2025-05-16
Add Nornir group support

mpfz0r:
Like with Ansible, use a label `nornir-group` to add nodes to a Nornir group in the inventory:

https://nornir.readthedocs.io/en/latest/tutorial/inventory.html#Inventory

Nornir would actually support multiple groups per node, but with the label approach we cannot do that nicely.

I wonder if we couldn't find a better way to do this. Like use a more generic `groups` notation?
https://github.com/srl-labs/containerlab/issues/2504

### DIFF
--- a/clab/inventory.go
+++ b/clab/inventory.go
@@ -185,6 +185,7 @@ type NornirSimpleInventoryKindProps struct {
 // the node registry.
 type NornirSimpleInventoryNode struct {
 	*types.NodeConfig
+	NornirGroup string
 }
 
 // NornirSimpleInventory represents the data structure used to generate the nornir simple inventory file.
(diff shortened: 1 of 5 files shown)


clab/inventory.go
```diff
@@ -243,6 +243,7 @@ func (c *CLab) generateNornirSimpleInventory(w io.Writer) error {
 			}
 
 		}
+		nornirNode.NornirGroup = n.Config().Labels["nornir-group"]
```
hellt, line 246: @mpfz0r you can probably allow setting multiple groups with comma separated values

mpfz0r: Yes that's a possibility, albeit a hackish one.
I wanted to see if a nicer way is possible. But I can also implement it this way...

mpfz0r: another way would be to collect all labels that start with `nornir-group`
eg.
```
labels:
  nornir-group-1: foo
  nornir-group-2: bar
```

hellt: yeah, thats is good as well

Issue #2604, opened 2025-05-26
Docs: Cumulus Linux VX 5.12 will be the last version


robinchrist:
FYI, it seems like Nvidia has silently discontinued support for Cumulus VX after 5.12.

From the docs:
https://docs.nvidia.com/networking-ethernet-software/cumulus-linux-513/Whats-New/

> Cumulus VX
> NVIDIA no longer releases Cumulus VX as a standalone image. To simulate a Cumulus Linux switch, use [NVIDIA AIR](https://docs.nvidia.com/networking-ethernet-software/nvidia-air/).

I think this might be worth documenting? If yes, I can create a docs PR..

hellt:
Yeah that's good to know for sure. Thanks for updating and if you come to raising a PR that'd be great

PR #2605, opened 2025-05-28
Docs: Document that Cumulus VX as been discontinued after version v5.12.1

robinchrist:
See https://github.com/srl-labs/containerlab/issues/2604

From the Cumulus Linux 5.13 docs:
https://docs.nvidia.com/networking-ethernet-software/cumulus-linux-513/Whats-New/

> Cumulus VX
> NVIDIA no longer releases Cumulus VX as a standalone image. To simulate a Cumulus Linux switch, use [NVIDIA AIR](https://docs.nvidia.com/networking-ethernet-software/nvidia-air/).

### DIFF
--- a/docs/manual/kinds/cvx.md
+++ b/docs/manual/kinds/cvx.md
@@ -11,6 +11,10 @@ CVX nodes launched with containerlab come up with:
 * the management interface `eth0` is configured with IPv4/6 addresses as assigned by either the container runtime or DHCP
 * `root` user created with password `root`
 
+!!! note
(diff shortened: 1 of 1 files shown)


ipspace:
FWIW, you should have also added the "vagrant account is valid for 6 months" FUBAR that will make even the 5.12.1 images useless after a few months: https://github.com/ipspace/netlab/blob/dev/docs/labs/cumulus_nvue.md?plain=1#L50

robinchrist:
> FWIW, you should have also added the "vagrant account is valid for 6 months"

Excuse me, WHAT? 😑

Good to know.

PR #2610, opened 2025-06-03
Pass interface specs on Fortigate node creation

jonapap:
Fixed small bug where the interface specs where not passed, making it impossible to create any links with this kind node.


vista- (approved):
Thanks for catching this issue and for contributing the fix! Looks good to me.

vista-:
In case you need it, here's a link to a build that contains your fix: https://github.com/srl-labs/containerlab/actions/runs/15521916829/artifacts/3284252873

We should have a release with some additional fixes coming later this month.

PR #2617, opened 2025-06-11
Add cJunosEvolved

kmo7:
* adding cJunosEvolved to containerlab
* adding the cJunosEvolved lab example and it's doc
* adding the cJunosEvolved description manual

### DIFF
--- a/docs/manual/kinds/cjunosevolved.md
+++ b/docs/manual/kinds/cjunosevolved.md
@@ -1,0 +1,120 @@
+---
+search:
+  boost: 4
+kind_code_name: juniper_cjunosevolved
+kind_display_name: Juniper cJunosEvolved
+---
+# Juniper cJunosEvolved
+
+[Juniper cJunosEvolved](https://www.juniper.net/documentation/product/us/en/cjunosevolved/) is a containerized Junos OS Evolved router identified with `-{{ kind_code_name }}-` kind in the [topology file](../topo-def-file.md). It is a KVM based container that can emulate either of these PTX platforms:
+
+* `PTX10002-36QDD`- A fixed form factor 800G transport router based on Juniper's Express 5 (aka BX) ASIC.
+* `PTX10001-36MR` - A fixed form factor 400G transport router based on Juniper's Express 4 (aka BT) ASIC.
+
+The above platforms are selected via these environmental variables being specified in the topology YAML file as shown in the `lab-examples` directories for cjunosevolved:
+
+ `CPTX_COSIM: "BX"`- For PTX10002-36QDD
+ `CPTX_COSIM: "BT"`- For PTX10001-36MR
+
+Juniper cJunosEvolved nodes launched with containerlab can be provisioned to enable SSH, SNMP, NETCONF and gNMI services.
+
+## How to obtain the image
+
+The container image can be freely downloaded from the [Juniper support portal](https://support.juniper.net/support/downloads/?p=cjunos-evolved) without a Juniper account. Type cJunosEvolved in the `Find a Product` search box.
+
+## Managing Juniper cJunosEvolved nodes
+
+!!!note
+    Containers with cJunosEvolved inside will take ~5min to boot to login prompt.
+    You can monitor the progress with `docker logs -f <container-name>`.
+!!!note
+    The management port IP is assigned by containerlab and is merged into the cJunosEvolved startup config.
+    Please refer to `docs/lab-examples/srl-cjunosevolved.md` for an example.
+
+=== "CLI via SSH"
+    to connect to the cJunosEvolved CLI
+    ```bash
+    ssh admin@<container-name>
+    ```
+
+=== "Enter CLI directly"
+    cJunosEvolved has to be fully booted before this succeeds
+    ```bash
+    docker exec -ti <container-name> cli
+    ```
+
+=== "NETCONF"
+    NETCONF server is running over port 830
+    ```bash
+    ssh admin@<container-name> -p 830 -s netconf
+    ```
+
+!!!info
+    Default user credentials: `admin:admin@123`
+
+## Interface naming
+
+You can refer to the following document for details of the interface mapping (https://www.juniper.net/documentation/product/us/en/cjunosevolved/)
+
+The default unchannelized interface mode is described here. This provides 36 interfaces for BX and 12 for BT flavors of cJunosEvolved. The channelized mode provides 144 interfaces for BX and 72 for BT. The mapping for all of these is described in the Juniper deployment document referenced above.
+
+The Linux host side interfaces are mapped to the JunosEvolved CLI notation as described in the document. To summarize:
+
+Juniper cJunosEvolved Linux uses the following mapping rules:
+
+* `eth0`- management interface connected to the containerlab management network
+* `eth1 -eth3` - Reserved interfaces  **Do not use these**
+* `eth4 onwards` - WAN interfaces.
+
+* For the BX unchannelized mode:
+
+The Linux `eth4 – eth39` interfaces correspond to the `et-0/0/0 – et-0/0/35` interfaces in the JunosEvolved CLI configuration.
+
+* For the BT unchannelized mode:
+
+The Linux `eth4 – eth15` interfaces correspond to the `et-0/0/0 – et-0/0/11` interfaces in the JunosEvolved CLI configuration.
+
+When containerlab launches -{{ kind_display_name }}- it assigns an IP address to the container's `eth0` management interface.
+This interface is transparently stitched with the cJunosEvolved's `re0:mgmt-0` interface such that users can reach the management plane of the -{{ kind_display_name }}- using containerlab's assigned IP.
+
+The WAN interfaces need to be configured with IP addressing manually using CLI or other available management interfaces.
+You could also pass in a startup CLI configuration file that has the WAN interface addresses specified. For example,
+refer to `lab-examples/srlcjunosevo01/cjunosevo.cfg`.
+
+## Features and options
+
+The user can configure cJunosEvolved in either the BX or BT mode via providing an env variable in the containerlab YAML file.
+Refer to `lab-examples/srlcjunosevo/srlcjunosevo01.clab.yml`.
+
+`CPTX_COSIM: "BX"` provisions the BX mode, i.e. the `PTX10002-36QDD` platform.
+`CPTX_COSIM: "BT"` provisions the BT mode, i.e. the `PTX10001-36MR` platform.
+
+### Node configuration
+
+Juniper cJunosEvolved nodes come up with a basic configuration. Users, management interfaces, and protocols such as SSH, NETCONF and SNMP are configured.
+
+#### Startup configuration
+
+You can make cJunosEvolved nodes boot up with a user-defined startup-config instead of a built-in one. With a [`startup-config`](../nodes.md#startup-config) property of the node/kind user sets the path to the config file that will be mounted to a container and used as a startup-config:
+
+```yaml
+topology:
+  nodes:
+    node:
+      kind: juniper_cjunosevolved
+      startup-config: cjunosevo.cfg
+```
+
+With this knob containerlab is instructed to take a file `cjunosevo.cfg` from the directory that hosts the topology file, and copy it to the lab directory for that specific node under the `/config/startup-config.cfg` name. Then the directory that hosts the startup-config dir is mounted to the container. This will result in this config being applied at startup by the node.
+
+Configuration is applied after the node is started, thus it can contain partial configuration snippets that you desire to add on top of the default config that a node boots up with.
+
+## Lab examples
+
+The following labs feature the Juniper cJunosEvolved node:
+
+* [SR Linux and Juniper cJunosEvolved](../../lab-examples/srl-cjunosevolved.md)
+
+## Known issues and limitations
+
+* To check the boot log, use `docker logs -f <node-name>`
--- a/lab-examples/srlcjunosevo/srlcjunosevo.clab.yml
+++ b/lab-examples/srlcjunosevo/srlcjunosevo.clab.yml
@@ -1,0 +1,20 @@
+name: srlcjunosevo
+
+topology:
+  nodes:
+    srl:
+      kind: nokia_srlinux
+      image: ghcr.io/nokia/srlinux:23.7.1
+      startup-config: srl.cli
+
+    cevo:
+      kind: juniper_cjunosevolved
+      image: cjunosevolved:25.2R1.3-EVO
+      startup-config: cjunosevo.cfg
+      env:
+        # To start the BT flavor, change "BX" to "BT" below
+        CPTX_COSIM: "BX"
+  links:
+    - endpoints: ["srl:e1-1", "cevo:eth4"]
+    - endpoints: ["srl:e1-2", "cevo:eth5"]
+    - endpoints: ["srl:e1-3", "cevo:eth6"]
--- a/nodes/cjunosevolved/cjunosevolved.go
+++ b/nodes/cjunosevolved/cjunosevolved.go
@@ -1,0 +1,100 @@
+// Copyright 2025 Nokia
+// Licensed under the BSD 3-Clause License.
+// SPDX-License-Identifier: BSD-3-Clause
+
+package cjunosevolved
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"regexp"
+
+	"github.com/charmbracelet/log"
+	"github.com/srl-labs/containerlab/netconf"
+	"github.com/srl-labs/containerlab/nodes"
+	"github.com/srl-labs/containerlab/types"
+	"github.com/srl-labs/containerlab/utils"
+)
+
+var (
+	kindnames = []string{"cjunosevolved", "juniper_cjunosevolved"}
+	defaultCredentials = nodes.NewCredentials("admin", "admin@123")
+        InterfaceRegexp = regexp.MustCompile(`et-0/0/(?P<port>\d+)$`)
+	InterfaceOffset = 0
+	InterfaceHelp   = "(et-0/0/X (where X >= 0) or ethX (where X >= 4)"
+)
+
+const (
+	scrapliPlatformName = "juniper_junos"
+	NapalmPlatformName  = "junos"
+
+	configDirName   = "config"
+	startupCfgFName = "startup-config.cfg"
+)
+
+// Register registers the node in the NodeRegistry.
+func Register(r *nodes.NodeRegistry) {
+        platformAttrs := &nodes.PlatformAttrs{
+	        ScrapliPlatformName: scrapliPlatformName,
+		NapalmPlatformName:  NapalmPlatformName,
+	}
+
+	nrea := nodes.NewNodeRegistryEntryAttributes(defaultCredentials, nil, platformAttrs)
+
+	r.Register(kindnames, func() nodes.Node {
+               return new(cjunosevolved)
+	}, nrea)
+}
+
+type cjunosevolved struct {
+	nodes.DefaultNode
+}
+
+func (n *cjunosevolved) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
+
+	// Init DefaultNode
+	n.DefaultNode = *nodes.NewDefaultNode(n)
+
+	// cjunosevolved requires KVM support.
+	n.HostRequirements.VirtRequired = true
+	n.HostRequirements.MinVCPU = 4
+	n.HostRequirements.MinVCPUFailAction = types.FailBehaviourError
+	n.HostRequirements.MinAvailMemoryGb = 8
+	n.HostRequirements.MinAvailMemoryGbFailAction = types.FailBehaviourLog
+
+	n.Cfg = cfg
+	for _, o := range opts {
+		o(n)
+	}
+
+	// mount config dir to support startup-config functionality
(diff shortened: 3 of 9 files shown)


hellt:
thanks @kmo7 

@vista- could you give this a try when you have time given you are my Juniper buddy 👯

hellt:
It seems the support portal does not list cjunos evolved as a product available for download

docs/manual/kinds/cjunosevolved.md
```diff
@@ -0,0 +1,120 @@
...
+The above platforms are selected via these environmental variables being specified in the topology YAML file as shown in the `lab-examples` directories for cjunosevolved:
+
+ `CPTX_COSIM: "BX"`- For PTX10002-36QDD
+ `CPTX_COSIM: "BT"`- For PTX10001-36MR
+
+Juniper cJunosEvolved nodes launched with containerlab can be provisioned to enable SSH, SNMP, NETCONF and gNMI services.
+
+## How to obtain the image
+
+The container image can be freely downloaded from the [Juniper support portal](https://support.juniper.net/support/downloads/?p=cjunos-evolved) without a Juniper account. Type cJunosEvolved in the `Find a Product` search box.
```
hellt, line 23: It seems the support portal does not list cjunos evolved as a product available for download

nodes/cjunosevolved/cjunosevolved.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/srl-labs/containerlab/utils"
+)
+
+var (
+	kindnames = []string{"cjunosevolved", "juniper_cjunosevolved"}
+	defaultCredentials = nodes.NewCredentials("admin", "admin@123")
+        InterfaceRegexp = regexp.MustCompile(`et-0/0/(?P<port>\d+)$`)
+	InterfaceOffset = 0
+	InterfaceHelp   = "(et-0/0/X (where X >= 0) or ethX (where X >= 4)"
+)
```
vista-, line 26: Since data ports start at eth4 (and not eth1 like most costs), you should also set `FirstDataIfIndex` to 4, otherwise NOS-like interface names won't work in topologies: See more information [over here](https://github.com/srl-labs/containerlab/pull/2124) about the interface aliasing feature.

docs/manual/kinds/cjunosevolved.md
```diff
@@ -0,0 +1,120 @@
+---
+search:
+  boost: 4
+kind_code_name: juniper_cjunosevolved
+kind_display_name: Juniper cJunosEvolved
+---
+# Juniper cJunosEvolved
+
+[Juniper cJunosEvolved](https://www.juniper.net/documentation/product/us/en/cjunosevolved/) is a containerized Junos OS Evolved router identified with `-{{ kind_code_name }}-` kind in the [topology file](../topo-def-file.md). It is a KVM based container that can emulate either of these PTX platforms:
```
vista-, line 9: From what I gather, is cJunosEvolved essentially an analogue to a [vrnetlab-ified](https://github.com/hellt/vrnetlab) vJunosEvolved?

docs/manual/kinds/cjunosevolved.md
```diff
@@ -0,0 +1,120 @@
...
+When containerlab launches -{{ kind_display_name }}- it assigns an IP address to the container's `eth0` management interface.
+This interface is transparently stitched with the cJunosEvolved's `re0:mgmt-0` interface such that users can reach the management plane of the -{{ kind_display_name }}- using containerlab's assigned IP.
+
+The WAN interfaces need to be configured with IP addressing manually using CLI or other available management interfaces.
+You could also pass in a startup CLI configuration file that has the WAN interface addresses specified. For example,
+refer to `lab-examples/srlcjunosevo01/cjunosevo.cfg`.
+
+## Features and options
+
+The user can configure cJunosEvolved in either the BX or BT mode via providing an env variable in the containerlab YAML file.
```
vista-, line 86: Instead requiring an user to set an envvar, perhaps it would be a better choice to let users set the `type` of the node and handle envvars internally in the code?

See more info about `type` here: https://containerlab.dev/manual/nodes/#type

In the relevant parts of the code, this would be exposed as `n.Cfg.NodeType`. Here's an implementation to go off of: https://github.com/srl-labs/containerlab/blob/main/nodes/srl/srl.go#L168

docs/manual/kinds/cjunosevolved.md
vista-: That's indeed true -- would it make more sense to merge this in once this goes live (otherwise we'll have to deal with support requests for a product that does not exist yet, as far as the public is concerned)?

docs/manual/kinds/cjunosevolved.md
```diff
@@ -0,0 +1,120 @@
...
+You can refer to the following document for details of the interface mapping (https://www.juniper.net/documentation/product/us/en/cjunosevolved/)
+
+The default unchannelized interface mode is described here. This provides 36 interfaces for BX and 12 for BT flavors of cJunosEvolved. The channelized mode provides 144 interfaces for BX and 72 for BT. The mapping for all of these is described in the Juniper deployment document referenced above.
+
+The Linux host side interfaces are mapped to the JunosEvolved CLI notation as described in the document. To summarize:
+
+Juniper cJunosEvolved Linux uses the following mapping rules:
+
+* `eth0`- management interface connected to the containerlab management network
+* `eth1 -eth3` - Reserved interfaces  **Do not use these**
```
vista-, line 66: Will these interfaces be created by an internal script?

lab-examples/srlcjunosevo/srlcjunosevo.clab.yml
```diff
@@ -0,0 +1,20 @@
...
+
+    cevo:
+      kind: juniper_cjunosevolved
+      image: cjunosevolved:25.2R1.3-EVO
+      startup-config: cjunosevo.cfg
+      env:
+        # To start the BT flavor, change "BX" to "BT" below
+        CPTX_COSIM: "BX"
+  links:
+    - endpoints: ["srl:e1-1", "cevo:eth4"]
```
vista-, line 18: Perhaps consider using the "aliased" interface names here (once the fix I mentioned is in)? It's easier for users to grasp and use, instead of having to calculate offsets themselves. Same goes for SRL node (ethernet-1/X) ;)

docs/manual/kinds/cjunosevolved.md
```diff
@@ -0,0 +1,120 @@
...
+    ```
+
+!!!info
+    Default user credentials: `admin:admin@123`
+
+## Interface naming
+
+You can refer to the following document for details of the interface mapping (https://www.juniper.net/documentation/product/us/en/cjunosevolved/)
+
+The default unchannelized interface mode is described here. This provides 36 interfaces for BX and 12 for BT flavors of cJunosEvolved. The channelized mode provides 144 interfaces for BX and 72 for BT. The mapping for all of these is described in the Juniper deployment document referenced above.
```
vista-, line 59: You mention a channelized interface mode -- is this going to be supported at some later point in the container / Clab implementation?

kmo7:
cJunosEvolved will be released at the end of this month and at that time, the link mentioned in the doc will be available.

We wanted to put this up for review early, so it can be merged in to containerlab when it is released.

If it helps, I can post the results of my testing, showing cJunosEvolved coming up and pinging Srlinux fine, etc on this review (or would you like it committed to the repo in the lab-examples document at the end?

I am the developer for cJunosEvolved, but we cannot provide an image till the official release goes out.  We can however meet and answer questions if you like?

Thanks.


Juniper Business Use Only

hellt:
Thanks @kmo7 for the comments. It is great seeing contributions for new systems coming ahead of the release time, love it.

Given our reliance on you doing the testing, I'd say the path forward is to address existing comments and then re-checking that the same MVP reproduction with srl-cjunos works

docs/manual/kinds/cjunosevolved.md
kmo7: The product is being shipped in two weeks and at that time the image and general deployment manual too will be posted publicly. If you like to delay committing this till then, that is fine.

nodes/cjunosevolved/cjunosevolved.go
kmo7: The implementation works as is, it is tested in containerlab. Please see my reply to your other comment about NOS like interfaces.

docs/manual/kinds/cjunosevolved.md
kmo7: cJunosEvolved is a container for those of our customers that prefer a container. Many customers have their own deployment environment distinct from containerlab. vJunosEvolved is the VM version and both products support the same EVO feature set. We plan to deprecate vJunosEvolved in containerlab, no reason for it in vr-netlab now that we have a container.

docs/manual/kinds/cjunosevolved.md
kmo7: We use this environment setting method for our general deployment of cJunosEvolved outside containerlab and customers are used to it from our previous generation of containers as well. The only requirement is to configure the different platforms (mentioned in the containerlab  doc) is to set the appropriate env var in the topology file. The code internally knows how to interpret that. We prefer not to change this method, and keep it generic to all deployment environments.

docs/manual/kinds/cjunosevolved.md
kmo7: eth1-eth3 are reserved for internal use and the user does not need to enumerate them in the yaml file i.e. The lab-example yaml file indication of how to define the interfaces works.
eth0 is the management interface, and is provided by containerlab. eth4 and beyond are WAN interfaces. Our user documentation (not yet released) has tables that show the interface mapping.

lab-examples/srlcjunosevo/srlcjunosevo.clab.yml
kmo7: Our general Deployment Document provides the mapping between the ethX and "et-" notation. Our code also does the transformation from the linux side ethX notation to the CLI "et-" interface notation internally. Users do not need to calculate offsets, they can refer to the provided example (for containerlab) and our Deployment Doc (if they wish) once that is released by the end of the month.

docs/manual/kinds/cjunosevolved.md
kmo7: We support the channelized mode for Docker/Docker Compose deployment on Linux Host servers. For containerlab, this is TBD  at this point. The unchannelized mode already provides 36 interfaces for the BX flavor and 12 for the legacy BT mode, which is plenty for most users to do basic bringup and testing. Will keep you posted if we decide to add channelized interfaces.

docs/manual/kinds/cjunosevolved.md
hellt: I am ok with leaving this at env var setting, given that there is only one env var that drives the emulated type.

But I wanted to check that leaving the env var unset defaults the system to one of the supported types.

lab-examples/srlcjunosevo/srlcjunosevo.clab.yml
vista-: These aliases are convenience features for users who want to replicate existing infrastructure in Containerlab or perform network CI/CD -- this is so their own business logic does not have to do the conversion between the ethX and the NOS-based `et-` interface naming. Additionally, it helps quite a lot when working with EdgeShark and the VS Code extension to see the NOS interface names when trying to pick out an interface to capture traffic on.

Most other kinds in Containerlab support this and I figured it might be nice to have as part of the day 0 feature set for a NOS.

Just to clarify: the ethX interface naming keeps working the same! It's just an [interface alternative name](https://lwn.net/Articles/794289/) applied alongside the regular ethX interface naming, the generation of this happens inside the Containerlab code and applied as part of the container link deployment process. If the container internally does not use interface altnames, it is transparent to it.

docs/manual/kinds/cjunosevolved.md
kmo7: Yes, this is true. If no env var indicating BT or BX is provided, cJunosEvolved comes up in the default BT mode.

lab-examples/srlcjunosevo/srlcjunosevo.clab.yml
kmo7: We have a way to automatically convert the linux ethX ifnames to the CLI aka NOS based names automatically so the user does not have to do it. I saw that you implemented this alias feature for containerlab. After the first commit goes through could you do the same for this container type? I want to get the first commit in.

docs/manual/kinds/cjunosevolved.md
```diff
@@ -0,0 +1,120 @@
...
+* `PTX10001-36MR` - A fixed form factor 400G transport router based on Juniper's Express 4 (aka BT) ASIC.
+
+The above platforms are selected via these environmental variables being specified in the topology YAML file as shown in the `lab-examples` directories for cjunosevolved:
+
+ `CPTX_COSIM: "BX"`- For PTX10002-36QDD
+ `CPTX_COSIM: "BT"`- For PTX10001-36MR
+
+Juniper cJunosEvolved nodes launched with containerlab can be provisioned to enable SSH, SNMP, NETCONF and gNMI services.
+
+## How to obtain the image
```
hellt, line 21: @kmo7 ping us here when the image becomes available, @vista- and myself will do some hands-on testing and then we merge this one

kmo7: Great. Will do. You can both can ping me on here if you need to discuss something after you start testing.

vista-: We can also be reached via Discord if need be :)

vista-:
@kmo7 I rebased your branch onto latest `main` to add a recent commit that fixes the CI/CD pipeline, just as a heads-up

kmo7:
> @kmo7 I rebased your branch onto latest `main` to add a recent commit that fixes the CI/CD pipeline, just as a heads-up

Thanks Vista, i don't see this commit reflected in containerlab.dev yet, e.g. on the front page where vendor notes are listed. 
https://containerlab.dev/ or in the nodes directory: https://github.com/srl-labs/containerlab/tree/main/nodes

Are you waiting on the image from us or the rebase operation to be done>

nleiva:
Hi all,

Might be too early, but would it be possible to add a **pre-config hook** to the `cJunosEvolved` router kind in Containerlab to execute a couple of operational commands that generate TLS certificates directly on the container?

The commands are:

```
request security pki generate-key-pair certificate-id myname
request security pki local-certificate generate-self-signed certificate-id myname subject CN=abc domain-name clab.com ip-address 1.2.3.4 email example@clab.com
```

This would enable the configuration of TLS-based services like RESTCONF in the startup-config:

```
services {
    rest {
        https {
            port 3000;
            server-certificate myname;
        }
    }
}
```

Let me know if that’s feasible or if there’s a preferred approach.

Thanks!
Nicolas

vista-:
Hi @kmo7,

Congrats on launching cJunosEvo! Thank you again for your hard work with preparing this PR.

I just tried out the cJunosEvo image, just a few observations:

The way the startup configuration to the node is provided does not really match the way most other Containerlab kinds do it: they just append the generated config required for management connectivity to the provided config.

cJunosEvolved, on the other hand, expects the provided startup config to contain the mgmt config to template into.

This essentially means a user would have to manually edit the configuration saved from the node in order for it to be usable in the next deploy (which might not assign the same management IP address to the node).

Rather than your current approach for generating an initial startup configuration, perhaps it would be better to put this logic (rendering a default config template and merging it with the user-provided startup config) into Containerlab? Check out how it's done with cRPD, it's fairly easy to template in runtime variables: https://github.com/srl-labs/containerlab/blob/main/nodes/crpd/crpd.go#L155

Some further notes:
- The default config loaded on the node is a bit lackluster. I would suggest adding the following:
  - Hostname should be included for easy identification of nodes in the topology
  - NETCONF and gNMI should be pre-enabled in the default config
  - `system { management-instance; }` helps avoid issues down the line where you might kill your own management connectivity
- I would add "Node can take 5+ minutes to boot" to the Kind documentation, since some users won't look in the container logs to find out if it's actually booting or not.
- Perhaps define a health check command to give feedback to the user/containerlab when the node has finished booting?

kmo7:
> Hi @kmo7,
> 
> Congrats on launching cJunosEvo! Thank you again for your hard work with preparing this PR.
> 
> I just tried out the cJunosEvo image, just a few observations:
> 
> The way the startup configuration to the node is provided does not really match the way most other Containerlab kinds do it: they just append the generated config required for management connectivity to the provided config.
> 
> cJunosEvolved, on the other hand, expects the provided startup config to contain the mgmt config to template into.
> 
> This essentially means a user would have to manually edit the configuration saved from the node in order for it to be usable in the next deploy (which might not assign the same management IP address to the node).
> 
> Rather than your current approach for generating an initial startup configuration, perhaps it would be better to put this logic (rendering a default config template and merging it with the user-provided startup config) into Containerlab? Check out how it's done with cRPD, it's fairly easy to template in runtime variables: https://github.com/srl-labs/containerlab/blob/main/nodes/crpd/crpd.go#L155
> 
> Some further notes:
> 
>     * The default config loaded on the node is a bit lackluster. I would suggest adding the following:
>       
>       * Hostname should be included for easy identification of nodes in the topology
>       * NETCONF and gNMI should be pre-enabled in the default config
>       * `system { management-instance; }` helps avoid issues down the line where you might kill your own management connectivity
> 
>     * I would add "Node can take 5+ minutes to boot" to the Kind documentation, since some users won't look in the container logs to find out if it's actually booting or not.
> 
>     * Perhaps define a health check command to give feedback to the user/containerlab when the node has finished booting?

*Reponse:*
Hi Vista, thanks for the review. Have incorporated some of the comments and we will look at the rest for the next commit. Since cJunosEvolved has now been also verified by you to be working, can you please merge the pull request into containerlab so the many others that are expressing interest to use it, can do so, and will get back to you on the rest? Please see below responses to the comments. 

- With respect to the longer comment up front, this is the key point "This essentially means a user would have to manually edit the configuration saved from the node in order for it to be usable in the next deploy (which might not assign the same management IP address to the node)."
3 points on this: 
1- The current FXP0ADDR token scheme our code uses, does correctly configure the management IP according to what containerlab passes in to the container, even when the cevo instance is restarted and a new management port IP is assigned by the containerlab. 
2- As you know, the save command has not been working in general for many containers in containerlab and you are fixing that but it is not yet committed:
 https://github.com/srl-labs/containerlab/pull/2659
 Once that is committed, for vr-net lab  based containers, could you also fix it for cjunosevo?
3- With the combination of the above 2 points, even though cjunosevolved does this differently than some other containers in clab, it will still function correctly.

wrt to these comments:
o	Hostname should be included for easy identification of nodes in the topology. Yes, agreed. We will address this in a subsequent commit. This feature works in our docker compose deployements, but has one line missing to make it work in containerlab. Either by adding that line or by pushing the hostname from the go code, we will fix this, but do not want to delay the initial commit, so people can bringup cjunosevo and use it initially.
o	NETCONF and gNMI should be pre-enabled in the default config
- netconf already is enabled in the following config file: lab-examples/srlcjunosevo/cjunosevo.cfg. Users can do same for their deployments (will show the test log for this at the end).
- GNMI was not enabled for our vJunosEvolved, vJunos-switch and vJunos-router also. This needs a bit more work and we will enable it in a subsequent commit. 
o	system { management-instance;  Added this to the above config file to separate the mgmt port into its own routing instance.
"•	I would add "Node can take 5+ minutes to boot" to the Kind documentation,". This was already documented in this kind doc file: docs/manual/kinds/cjunosevolved.md, line 28. If you think it is needed elsewhere too, can you please add it? 

"Perhaps define a health check command to give feedback to the user/containerlab when the node has finished booting?" I have added this info the kind documentation file. Besides the docker log command, the "    # docker exec -ti clab-srlcjunosevo-cevo cli" indicates bootup progress periodically, so the user is not blind to that. 

Thanks for the careful review Vista. I hope above answers the immediate comments so you can commit the code into clab and I will follow up w/ the rest. 

Here is a sample netconf log:

# ssh admin@clab-srlcjunosevo-cevo -p 830 -s netconf
Warning: Permanently added '[clab-srlcjunosevo-cevo]:830' (ED25519) to the list of known hosts.
(admin@clab-srlcjunosevo-cevo) Password: 


<hello xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
  <capabilities>
    <capability>urn:ietf:params:netconf:base:1.0</capability>
    <capability>urn:ietf:params:netconf:capability:candidate:1.0</capability>
    <capability>urn:ietf:params:netconf:capability:confirmed-commit:1.0</capability>
    <capability>urn:ietf:params:netconf:capability:validate:1.0</capability>
    <capability>urn:ietf:params:netconf:capability:url:1.0?scheme=http,ftp,file</capability>
    <capability>urn:ietf:params:xml:ns:netconf:base:1.0</capability>
    <capability>urn:ietf:params:xml:ns:netconf:capability:candidate:1.0</capability>
    <capability>urn:ietf:params:xml:ns:netconf:capability:confirmed-commit:1.0</capability>
    <capability>urn:ietf:params:xml:ns:netconf:capability:validate:1.0</capability>
    <capability>urn:ietf:params:xml:ns:netconf:capability:url:1.0?scheme=http,ftp,file</capability>
    <capability>urn:ietf:params:xml:ns:yang:ietf-yang-metadata?module=ietf-yang-metadata&amp;revision=2016-08-05</capability>
    <capability>urn:ietf:params:xml:ns:yang:ietf-netconf-monitoring</capability>
    <capability>http://xml.juniper.net/netconf/junos/1.0</capability>
    <capability>http://xml.juniper.net/dmi/system/1.0</capability>
    <capability>http://yang.juniper.net/junos/jcmd?module=junos-configuration-metadata&amp;revision=2021-09-01</capability>
  </capabilities>
  <session-id>31674</session-id>
</hello>
]]>]]>

PR #2618, opened 2025-06-12
Use HostConfig.Mounts for anonymous volumes

mzagozen:
Containerlab passes all strings from the node "binds" property to Docker via the HostConfig.Binds field. This API supports anonymous volumes and the container does correctly start up with an new anonymous volume attached. But the volume itself is missing the "Name" attribute for some reason (docker inspect). This makes it impossible to then access the volume from other containers, like with `docker run --volumes-from <existing-container> ...`. An anonymous volume created with the new Mounts spec is for all intents and purposes equivalent to one created with Binds, except it has the "Name" attribute set to its random ID.

This has worked in the past, so maybe there is a regression in Docker? Anyway by fixing it here we can ensure it will work for all Docker versions while maintaining compatibility with existing containerlab topologies.

As a side note, while I was reviewing the Docker documentation, I noticed my change to add anonymous volume support to containerlab node `binds` in #1853 mixes up two distinct concepts: [Bind mounts](https://docs.docker.com/engine/storage/bind-mounts/) and [Volumes](https://docs.docker.com/engine/storage/volumes/). In hindsight this feels like a mistake and volumes would be better handled in a new `volumes` node field. This would also align the syntax with docker-compose. What is the design guideline here?

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -297,6 +297,17 @@ func (c *CLab) createNodeCfg( //nolint: funlen
 
 	nodeCfg.Binds = binds
 
+	// initialize volumes
+	volumes, err := c.Config.Topology.GetNodeVolumes(nodeName)
+	if err != nil {
+		return nil, err
+	}
+	err = c.resolveVolumePaths(volumes, nodeName)
+	if err != nil {
+		return nil, err
+	}
+	nodeCfg.Volumes = volumes
+
 	nodeCfg.PortSet, nodeCfg.PortBindings, err = c.Config.Topology.GetNodePorts(nodeName)
 	if err != nil {
 		return nil, err
@@ -614,12 +625,6 @@ func (c *CLab) resolveBindPaths(binds []string, nodeName string) error {
 		// host path is a first element in a /hostpath:/remotepath(:options) string
 		elems := strings.Split(binds[i], ":")
 
-		if len(elems) == 1 {
-			// if there is only one element, it means that we have an anonymous
-			// volume, in this case we don't need to resolve the path
-			continue
-		}
-
 		// replace special variables
 		r := c.magicVarReplacer(nodeName)
 		hp := r.Replace(elems[0])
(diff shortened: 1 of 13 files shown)


hellt:
Hi @mzagozen 

I would agree that adding `volumes` as a property is the right approach here instead of overloading `binds`.
If you want to take this one and make a PR that would be great and it would simplify the logic and separate the concerns

plajjan:
FWIW, I have encountered the very same issue. On Docker version 28.04 I get this problem. I tried downgrading to version 26 which has the old & working behavior, so somewhere between those versions there was a change in Docker behavior.

mzagozen:
I rewrote the user-facing side of the initial proposal to add a new `volumes` property and remove support for anonymous volumes from `binds`. Internally `binds` still uses the existing *HostConfig.Bind* field and `volumes` the correct *HostConfig.Mounts*. I tested this with Docker 26.1.5 for backwards compatibility.

mzagozen:
Rebased to fix a conflict.

FloSch62 (changes requested):
Review findings (highest severity first)

- `[blocker]` `types/bind.go:43` - This breaks existing topologies using one-field `binds` as anonymous volumes.
  `binds: - /var/run/somedir` was documented behavior and the PR body says compatibility is preserved, but this now errors during topology parsing. Fix by keeping old `binds` anonymous-volume syntax as a compatibility alias and converting it internally to `Volumes`, then document deprecation separately.

- `[blocker]` `runtime/podman/util.go:322` - Podman volumes are mapped into `specs.Mount{Type: "volume"}`.
  Podman specgen has a dedicated `ContainerStorageConfig.Volumes []*specgen.NamedVolume` path for named/anonymous volumes. Feeding them through OCI `Mounts` is the wrong owner/API and likely does not create managed volumes correctly. Fix by returning both bind mounts and `specgen.NamedVolume` values, then set `Mounts` and `Volumes` separately.

- `[major]` `types/topology.go:781` - `GetNodeVolumes` ignores group inheritance.
  The project inheritance model is `node -> group -> kind -> defaults`; this hand-written merge only uses defaults, kind, and node. Fix by using `mergeStringSliceFields` like `GetNodeBinds`, then de-dupe by parsed destination. Add a group-level volume override test.

- `[major]` `runtime/docker/docker.go:617` / `runtime/docker/docker.go:1457` - Volume conversion failures and unsupported options fail open.
  On conversion error, Docker container creation continues with `Mounts = nil`; unsupported options are only debug-logged and ignored. That can start a lab without requested storage or with weaker semantics than the topology says. Fix by returning the error from `CreateContainer`, and either implement or reject unsupported options explicitly.

- `[minor]` `docs/manual/nodes.md:282` - Docs overpromise and need cleanup.
  "same short syntax as Docker --volume" is not true if options are partially ignored. Also `anonymousvolume` typo, no migration note for old anonymous `binds`, and no group/default/kind/node example.

hellt:
since the code base diverted since quite a lot, I opened a new PR #3335 where the original commit is retained. Hence closing this

PR #2631, opened 2025-06-25
Fix #2621

### LINKED ISSUE #2621: `--cleanup` doesn't delete the lab directory if no containers are running
It looks like this was the intended functionality #748, #753 . But it looks like it was broken at some point

evilhamsterman:
The destroy function was just returning if it didn't find any containers. Added some logic to look for a lab directory and remove it.

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -52,6 +52,11 @@ func init() {
 }
 
 func destroyFn(_ *cobra.Command, _ []string) error {
+	// cleanup doesn't make sense with node-filter
+	if len(common.NodeFilter) != 0 && cleanup {
+		return fmt.Errorf("cleanup cannot be used with node-filter")
+	}
+
 	var err error
 	var labs []*clab.CLab
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


cmd/destroy.go
```diff
@@ -63,13 +68,34 @@ func destroyFn(_ *cobra.Command, _ []string) error {
...
 		if len(cnts) == 0 {
-			log.Info("no containerlab containers found")
+			log.Info("No containerlab containers found")
+			if cleanup {
+				// do our best to find a labdir
+				var labDirs []string
+				if common.Topo != "" {
+					topoDir := filepath.Dir(common.Topo)
+					log.Debug("Looking for lab directory next to topology file", "path", topoDir)
+					labDirs, _ = filepath.Glob(filepath.Join(topoDir, "clab-*"))
```
hellt, line 85: I think we can assume that `common.Topo` is not empty, because this is the prerequisite for being in the `case !all:`

evilhamsterman: Not if you run `containerlab destroy --name <labname> --cleanup`. Though I guess in that case you are either running in the same folder and `filepath.Dir` will return that, or you're outside the folder in which case we don't really have a way to find the lab folder anyway.

Issue #1685, opened 2023-10-30
clab config not persistent after redeploy


KTodts:
I noticed after upgrading my clab version from 43.0 to 47.1 that the config is not persistent after a redeploy of the topology.

Test would be, deploy topology for the first time --> make a config change --> destroy lab --> deploy again --> verify if change is persistent.

1 upgrade clab to version 47.1
```
root@s4qd-pod75:/home/ubuntu/gitlab/srl-dc-fabric-fundamentals-lab# clab version upgrade
A newer containerlab 0.47.1 is available. Release notes: https://containerlab.dev/rn/0.47.1
You are running containerlab 0.43.0 version
Downloading https://github.com/srl-labs/containerlab/releases/download/v0.47.1/containerlab_0.47.1_linux_amd64.deb
Preparing to install containerlab 0.47.1 from package
(Reading database ... 115507 files and directories currently installed.)
Preparing to unpack .../containerlab_0.47.1_linux_amd64.deb ...
Unpacking containerlab (0.47.1) over (0.43.0) ...
Setting up containerlab (0.47.1) ...

                           _                   _       _
                 _        (_)                 | |     | |
 ____ ___  ____ | |_  ____ _ ____   ____  ____| | ____| | _
/ ___) _ \|  _ \|  _)/ _  | |  _ \ / _  )/ ___) |/ _  | || \
( (__| |_|| | | | |_( ( | | | | | ( (/ /| |   | ( ( | | |_) )
\____)___/|_| |_|\___)_||_|_|_| |_|\____)_|   |_|\_||_|____/

    version: 0.47.1
     commit: 1da5ebd2
       date: 2023-10-24T22:09:40Z
     source: https://github.com/srl-labs/containerlab
 rel. notes: https://containerlab.dev/rn/0.47/#0471
```

2 Start with a fresh deployment by using --reconfgure
```
root@s4qd-pod75:/home/ubuntu/gitlab/srl-dc-fabric-fundamentals-lab# ll
total 72
drwxr-xr-x  6 root root 4096 Oct 26 09:31 ./
drwxr-xr-x 17 root root 4096 Oct 26 09:22 ../
-rwxr-xr-x  1 root root   37 Oct 26 09:22 00.show_lab.sh*
-rwxr-xr-x  1 root root   36 Oct 26 09:22 01.deploy_srl.sh*
-rwxr-xr-x  1 root root  335 Oct 26 09:22 03.initialize_hosts.sh*
[...]

PR #2642, opened 2025-06-28
Fix "clab config not persistent after redeploy" #1685

evilhamsterman:
I too experienced the issue in #1685. The tests in the default_node_tests were limited to only testing the suppress option.

According to the docs it's implied if `startup-config` is defined, but `enforce-startup-config!=true` then the config saved by the node should be used unless `--cleanup` or `--reconfigure` is used. 

This makes sense because you could provide a lab with predefined configurations to nodes to share with others. But you may want to stop the lab and have any changes you've been experimenting with remain for when you next start it.

I added tests to test other options like the one above, and additionally the currently undefined option where you have `enforce-startup-config=true` but don't have a `startup-config` defined, which I think should result in an error.

### DIFF
--- a/nodes/default_node.go
+++ b/nodes/default_node.go
@@ -363,12 +363,12 @@ func (d *DefaultNode) GenerateConfig(dst, t string) error {
 	}
 
 	if d.Cfg.SuppressStartupConfig {
-		log.Infof("Startup config generation for '%s' node suppressed", d.Cfg.ShortName)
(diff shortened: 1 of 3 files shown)


evilhamsterman (commented):
A bit of logic fixing and I believe this should resolve the issues. I guess there could be an argument to allow `enforce-startup-config` without setting a `startup-config` if you just want it to regenerate the default config on every start. Though in that case you could just `containerlab redeploy --cleanup`

PR #2647, opened 2025-07-02
Add 'groups'

### LINKED ISSUE #2504: Make `groups` work as they were designed to

kaelemc:
Add `groups` to the `topology` container of the YAML.

```yaml
name: group-demo
topology:
  groups:
    leafs:
      labels:
        someLabel: someValue
      type: ixrd3l

  nodes:
    node1:
      group: leafs

    node2:
```

Precedence is node config > group config > kinds config > defaults

In progress:

- [ ] Testing full functionality
- [x] Adding tests
- [x] Adding docs

Closes #2504

### DIFF
--- a/docs/manual/nodes.md
+++ b/docs/manual/nodes.md
@@ -52,9 +52,11 @@ With `type` the user sets a type of the node. Types work in combination with the
 
 Other nodes might treat `type` field differently, that will depend on the kind of the node. The `type` values and effects defined in the documentation for a specific kind.
 
-### group
+### group 
 
-`group` is a freeform string that denotes which group a node belongs to. The grouping is currently only used to sort topology elements on a [graph](../cmd/graph.md#layout-and-sorting).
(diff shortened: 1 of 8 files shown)


hellt:
great effort, thanks @kaelemc

Issue #2638, opened 2025-06-26
docker v28 and security enforcements


hellt:
Docker v28 contains a backwards-incompatible change when external access to the containers is concerned

https://www.docker.com/blog/docker-engine-28-hardening-container-networking-by-default/

In clab realm, this likely means that we can't access unpublished ports from the hosts outside of the LAN. Some users start to hit this - https://discord.com/channels/860500297297821756/1387443912146157693/1387485324128686123

Clab's installation script is pinned right now to v27 https://github.com/srl-labs/containerlab/blob/3735d83fbf7454ef4180c2d3271ca15c3219fe53/utils/quick-setup.sh#L6 thus most users won't face the fallout of this, but if the users install docker separately and then install clab on top, they might hit the issue.

A remediation issue is to add an option to the bridge that clab creates:

```
docker network create -d bridge \
  -o com.docker.network.bridge.gateway_mode_ipv4=nat-unprotected \
  my_unprotected_net
```

This also might/should be an option that is toggle-able so that people can opt-out from default clab's behavior of exposing all unpublished ports.

Issue #2633, opened 2025-06-25
Quick setup script fails to install docker on Fedora 42


kaelemc:
The quick setup script `curl -sL https://containerlab.dev/setup | sudo -E bash -s "all"` fails to install docker correctly on Fedora 42. I suspect this is an issue with dnf5.

kaelemc:
Slight syntax change of `dnf` -> `dnf-3` should fix this, while maintaing backwards compatability (for the most part). However the docker repos don't provide f42 packages for any docker version lower than v28 (see #2638). 

This means on Fedora the script is broken for a while unless we just chose to install the latest docker for fedora in the interim. @hellt ?

kaelemc:
```
[kaelem@klmdev Downloads]$ dnf-3 list docker-ce --showduplicates
Last metadata expiration check: 0:05:38 ago on Mon 30 Jun 2025 23:38:59.
Available Packages
docker-ce.x86_64                                           3:28.1.0-1.fc42                                            docker-ce-stable
docker-ce.x86_64                                           3:28.1.1-1.fc42                                            docker-ce-stable
docker-ce.x86_64                                           3:28.2.0-1.fc42                                            docker-ce-stable
docker-ce.x86_64                                           3:28.2.1-1.fc42                                            docker-ce-stable
docker-ce.x86_64                                           3:28.2.2-1.fc42                                            docker-ce-stable
docker-ce.x86_64                                           3:28.3.0-1.fc42                                            docker-ce-stable
```

hellt:
Docker provides v28 at a minimum for f42?

kaelemc:
Yeah. https://download.docker.com/linux/fedora/42/x86_64/stable/Packages/

hellt:
@kaelemc I think then we need to take 28.2.2-1.fc42 maybe for fedora 42+?

vista-:
> Slight syntax change of `dnf` -> `dnf-3` should fix this, while maintaing backwards compatability (for the most part). However the docker repos don't provide f42 packages for any docker version lower than v28 (see [#2638](https://github.com/srl-labs/containerlab/issues/2638)).
> 
> This means on Fedora the script is broken for a while unless we just chose to install the latest docker for fedora in the interim. [@hellt](https://github.com/hellt) ?

[...]

Issue #2269, opened 2024-10-29
Expose Docker network options to a user


hellt:
Currently we don't allow setting any of the custom options via a topo file. 

We need to think how to expose driver options so that a user can set them
```
driver_opts:
      com.docker.network.bridge.gateway_mode_ipv4: 'routed'
```

hellt:
done in #2648

PR #2648, opened 2025-07-03
runtime: Add network driver options, fix gateway mode in Docker >=28

vista-:
This PR fixes connectivity after a behavior change in Docker 28 and above (see #2638). This also fixes Containerlab on distros where Docker v27 packages are not available (such as Fedora 42 #2633).

Additionally, this PR adds a new feature to the mgmt section of the topology (as requested in #2269), the `driver-opts` section, where users can add custom bridge network driver options as key-value pairs.


BYT3M3:
This PR didn't include changes to /schemas/clab.schema.json, which results in a false-positive when using yaml schema validation (e.g. vscode extension redhat.vscode-yaml)

PR #2652, opened 2025-07-04
Creation of bridges within containers and wiring links into container based bridges

steiler:
Allow for the creation of bridges within containers and also the wiring into these bridges.

```
name: "bridge-ns"
topology:
  nodes: 
     br01:
       kind: bridge
       network-mode: container:bp1
     bp1:
       kind: linux
       image: alpine:latest
     c1:
       kind: linux
       image: alpine:latest
   links:
     - endpoints: ["c1:eth1", "br01:c1eth1"]
```

The bridge `br01` is being created in the `bp1` container. Then the client `c1` is attached to the bridge that resides in container `bp1`.

A limitation that exists with this approach is, that in the nodes section names need to be unique, hence `br01` can only exist once, although it would be possible to have it exist with the same name in different namespaces.

Possible solution to this:
Add a pre or postfix to the bridgenames in the topology whilst everything behind / before the delimiter would be removed on provisioning.
```
  nodes: 
     br01|bp1:
       kind: bridge
       network-mode: container:bp1
```
the part ofter the delimiter is not necessarly the base container name whatever makes it unique should work.

### DIFF
--- a/nodes/bridge/bridge.go
+++ b/nodes/bridge/bridge.go
@@ -44,9 +44,8 @@ func Register(r *nodes.NodeRegistry) {
 
 type bridge struct {
 	nodes.DefaultNode
-	// store for the nodes of the topology, required to setup bridges within container namespaces.
-	nodesMap    map[string]nodes.Node
 	containerNs string
+	nodesMap    map[string]nodes.Node
 }
 
 func (s *bridge) Init(cfg *types.NodeConfig, opts ...nodes.NodeOption) error {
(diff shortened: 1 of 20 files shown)


nodes/bridge/bridge.go
```diff
@@ -74,17 +118,43 @@ func (b *bridge) PostDeploy(_ context.Context, _ *nodes.PostDeployParams) error
...
+		return "", err
 	}
-	// check bridge exists
-	_, err = utils.BridgeByName(b.Cfg.ShortName)
+	return curns.Path(), nil
+}
+
+func (b *bridge) CheckDeploymentConditions(ctx context.Context, nodes map[string]nodes.Node) error {
+	// store nodes map for later use
+	b.nodesMap = nodes
```
hellt: instead of changing CheckDeploymentConditions by adding the `nodes` arg to it, can't we use `*nodes.DeployParams` in the `func (n *bridge) Deploy(ctx context.Context, _ *nodes.DeployParams) error` and pass the nodes map with the deploy params and store the nodes map via it?

steiler: That was my first idea. But No!
So CheckDeploymentConditions is called a lot earlier, it is part of the checkTopologyDefinition() https://github.com/srl-labs/containerlab/blob/21844a9595a2cea79e9e5bb050e5106bd4fd795c/clab/config.go#L320 call.
image

steiler: You are right... sorry!
This was an intermediary solution where I had to change it. But then I realized, that the existence check for a bridge within a container that we're trying to deploy is not thhhaattt useful... thats when this check came into play...
https://github.com/srl-labs/containerlab/pull/2652/files#diff-0ccbceb781798ef8f90f33c1d85c93e16a66e9c1969c9e5913c9035468cb6700R169
otherwise the GetNsPath() would already need to function on the CheckDeploymentConditions() call... but you're right, that change inbetween. Again Sorry!

steiler: ok fixed

PR #2659, opened 2025-07-11
nodes: Add generic implementation for VM-based node startup config saving

### LINKED ISSUE #2615: Generic Scrapli implementation does not work for saving
Needs more testing, but `containerlab save` does not work for the following Kinds:
- vJunos-router
- vJunos-switch
- vMX
- vSRX

This might also affect other Kinds that rely on the generic Scrapli-based save implementation. Will look into this in a bit.

vista-:
This PR adds a generic implementation for fetching and saving the configuration of `vrnetlab`-based nodes based on Scrapli. Note that some `vrnetlab`-based nodes were not touched, which had unique ways of saving the config already implemented; this change only affects node types which had a saving implementation based on the `netconf.SaveConfig` function.

At the moment, Scrapli only supports the following NOS types, so saving will fail for the rest (but these have not functioned properly for a long time now):

- arista_eos
- cisco_iosxe
- cisco_iosxr
- cisco_nxos
- juniper_junos

In progress of adding this feature, a generic `PreDeploy` and `SaveConfig` function was implemented for `VRNode`, removing a bunch of duplicated code across these Kinds.

Additionally, CheckPoint CloudGuard, IPInfusion ocNOS and Dell FT-OS node Kinds have been changed over to being VRNode-based, since their initial implementation was DefaultNode-based (which made little sense at this point).

Please let me know if Netconf-based startup config saving should be re-added for the affected node Kinds -- I did not see that much point in leaving it in, since destroying the VM gets rid of the startup-config within the node anyway.

I would appreciate some help in testing this functionality, so far, I have only confirmed Junos-based nodes to be working well with Scrapli, and I don't have a too big collection of Cisco VM-based nodes.

fix #2615


vista-:
I'll address the issues brought up by the static checker sometime over the weekend or on Monday.

evilhamsterman:
I've been working on making persistence work by moving the disk overlay so it can be saved to the clab folder, which is generic to any VM.

mischadiehm:
Thanks for this work - it's been sth long awaited and done so often by hand (: I really like the approach @evilhamsterman is suggesting since it seems a more natural transparent way to the problem. Just not sure how much work still needed and if really possible for all platrforms in vrnetlab - you might be able to give some more detail @evilhamsterman?

That said i also like this approach a lot and wonder how this could be extended to platforms supported in the https://scrapli.github.io/scrapli_community/ - what would be the dependencies to resolve and test for to make these platforms also available @vista- ? I can test mikrotik ros if that would be of any help?

evilhamsterman:
It's a two part job. The vrnetlab launch code already creates an "overlay" disk file for every disk. The problem is it's not persisted anywhere. So part one change the location of the overlay disk in the launcher, put it in a subfolder instead of just beside the parent image. 

This shouldn't make any difference in the current containers because it's just a path change.

Next is the bit more time consuming. In every node type in containerlab add a volume mount from the containerlab folder to the overlay folder. Then when the container creates the overlay image it's saved to the volume if you stop the lab.

It's pretty space efficient because it's a COW delta, and multiple nodes can share a base.

It could cause a problem if you updated the container with the base image. There's a couple solutions, copy the full parent, this would be slow and not as efficient. Could do the copy in the background or something. Or maybe don't embed the base disk, mount it in externally or even download it live.

I've done some work on part one. In my fork https://github.com/evilhamsterman/vrnetlab in the overlay branch, but I've been busy with some remodeling. Take a look and feel free to use it

vista-:
> Thanks for this work - it's been sth long awaited and done so often by hand (: I really like the approach @evilhamsterman is suggesting since it seems a more natural transparent way to the problem. Just not sure how much work still needed and if really possible for all platrforms in vrnetlab - you might be able to give some more detail @evilhamsterman?
> 
> That said i also like this approach a lot and wonder how this could be extended to platforms supported in the https://scrapli.github.io/scrapli_community/ - what would be the dependencies to resolve and test for to make these platforms also available @vista- ? I can test mikrotik ros if that would be of any help?

Sadly the Scrapli Community platforms are not available for the Golang-based Scrapligo (or at least I am not aware of a Golang port of that repository).

> It's a two part job. The vrnetlab launch code already creates an "overlay" disk file for every disk. The problem is it's not persisted anywhere. So part one change the location of the overlay disk in the launcher, put it in a subfolder instead of just beside the parent image.
> 
> This shouldn't make any difference in the current containers because it's just a path change.
> 
> Next is the bit more time consuming. In every node type in containerlab add a volume mount from the containerlab folder to the overlay folder. Then when the container creates the overlay image it's saved to the volume if you stop the lab.
> 
> It's pretty space efficient because it's a COW delta, and multiple nodes can share a base.
> 
> It could cause a problem if you updated the container with the base image. There's a couple solutions, copy the full parent, this would be slow and not as efficient. Could do the copy in the background or something. Or maybe don't embed the base disk, mount it in externally or even download it live.
> 
> I've done some work on part one. In my fork https://github.com/evilhamsterman/vrnetlab in the overlay branch, but I've been busy with some remodeling. Take a look and feel free to use it

Sounds like a quite creative solution for this problem! I'll keep an eye on it 👀

This could be a good workaround for platforms where we don't have an easy way of grabbing the config from the node. I would honestly prefer if this was a fallback option only (e.g. for platforms that don't have a "single configuration", such as a FreeBSD or Linux VM), since it's much harder to share VM overlay images than say, a text-based configuration file.

evilhamsterman:
I think that my method is just for general persistence, it's useful for any platform for starting and stopping devices. Saving configs is definitely helpful as well on platforms that support it.

hellt:
> Sadly the Scrapli Community platforms are not available for the Golang-based Scrapligo (or at least I am not aware of a Golang port of that repository).

I think it is not a problem to just take out the prompts and add those files to our/scrapli repo https://github.com/scrapli/scrapligo/tree/main/assets/platforms

They can be loaded live

vista-:
> > Sadly the Scrapli Community platforms are not available for the Golang-based Scrapligo (or at least I am not aware of a Golang port of that repository).
> 
> I think it is not a problem to just take out the prompts and add those files to our/scrapli repo https://github.com/scrapli/scrapligo/tree/main/assets/platforms
> 
> They can be loaded live

The config saving relies on https://github.com/scrapli/scrapligocfg, which seems to have tailor-made implementations for each NOS type it supports.

nodeutils/save.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2020 Nokia
+// Licensed under the BSD 3-Clause License.
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Package netconf contains netconf-based and functions used in containerlab.
+package nodeutils
+
+import (
```
hellt: lets bring it back to the netconf package, iso node utils

vista-: Renamed it back

netconf/save.go
```diff
@@ -0,0 +1,98 @@
...
+	_, err = d.CopyConfig("running", "startup")
+	if err != nil {
+		return fmt.Errorf("%s: Could not send save config via Netconf: %+v", addr, err)
+	}
+
+	return nil
+}
+
+// GetConfig retrieves the running configuration and returns it as a string. It automatically picks the appropriate network driver for the provided Scrapli Platform.
+func GetConfig(addr, username, password, scrapliPlatform string) (string, error) {
```
hellt: not sure `save.go` is a warranted name. Since you have GetConfig here. The common netconf.go may be a better choice after all, unless we will have way more functions in it that would warrant the split

vista-: You're right, I'll go change that

PR #2660, opened 2025-07-12
Chore/golangci updates

carlmontanari:
@hellt as we briefly discussed -- we can close/ignore this ofc but was just messing around and figured id tidy some stuff up and repin golangci. still a load of stuff disabled otherwise would be a zillion complaints, but at least gets a bit more static analysis coverage w/ this. lemme know if you wanna keep or trash this since ill start on the cmd package decoupling and would be nice to merge or ignore this to not have merge conflict hell :)

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,13 +1,42 @@
 name: golangci-lint
 on:
-    workflow_dispatch:
+  workflow_call:
+    inputs:
+      go_ver:
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      go_ver:
+        description: "Go version"
+        required: false
+        default: "1.23.7"
 jobs:
-    golangci:
-        name: lint
-        runs-on: ubuntu-20.04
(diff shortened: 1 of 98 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -1,13 +1,13 @@
...
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v4
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@v8
-        with:
-          version: v1.38
+    golangci:
+        name: lint
+        runs-on: ubuntu-20.04
```
hellt:
```suggestion
        runs-on: ubuntu-24.04
```

I think it is time :D

carlmontanari: lololol yeah I think so too 🤣  will update!

carlmontanari: set to latest for just golangci-lint -- left the rest as they are per our convo re docker versioning etc.

PR #2661, opened 2025-07-13
store backup files in temp dir

### LINKED ISSUE #2592: move bak files into global containerlab folder
bak is created to address an issue when a user removes or changes the clab file while the lab is still running
then they have no chance to revert it back without a bak file

probably having this bak outside of the lab dir in the ~/.containerlab dir is way more appropriate

hellt:
fix #2592 

The bak files are now stored in `/tmp/.clab/bak` directory and timestamped:

```
❯ ls -l /tmp/.clab/bak 
total 4
-rw-r--r-- 1 roman roman 253 Jul 13 22:05 250713_220527_topo-1353611672.clab.yml
```

### DIFF
--- a/types/topo_paths.go
+++ b/types/topo_paths.go
@@ -177,6 +177,7 @@ func (*TopoPaths) ClabTmpDir() string {
 	if !utils.DirExists(clabTmpDir) {
 		utils.CreateDirectory(clabTmpDir, 0o755)
 	}
+
 	return clabTmpDir
 }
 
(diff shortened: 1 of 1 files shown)


steiler:
@hellt /tmp is not a good pick... it is cleared on restart.
we should use /var/cache/containerlab or /var/lib/containerlab.

hellt:
I was choosing between ~/.clab/bak and `/tmp/bak` and decided to go with tmp since I did not want to think how to rollup the ~/.clab on fill up.

The initial goal for .bak to exist was when people were removing the lab dir while the lab was running, rendering the deletion with `clab destroy` not working.

With this change to end up in a bad state a user has to do both:

1. remove the lab dir
2. and reboot

Will see if that will happen often to reconsider the tmp location and deal with rolling up of the ~/.clab/bak dir

steiler:
We have seen that there are a lot of multi-user deployments... so `~` would be the wrong place in my view... it should be a system wide location... hence `/var/<whatever>` would be a lot better suited.

PR #2663, opened 2025-07-14
tests: Move S3 test to be the last to fix CI

vista-:
This is just a temp fix until we figure out if it's just GitHub Actions acting up, or if this is an actual issue.


vista-:
This is due to a Podman bug, it seems like, nothing to do with the S3 test, it's just that this one has links in it (and the others don't).

PR #2664, opened 2025-07-15
Feat/prep for deploy cmd decoupling

carlmontanari:


cmd/root.go
```diff
@@ -34,14 +34,6 @@ var RootCmd = &cobra.Command{
 	Aliases:           []string{"clab"},
 }
 
-// Execute adds all child commands to the root command and sets flags appropriately.
```
carlmontanari, line 37: appears this part was dead code after the (I think) relatively recent move to use fang for the cli

cmd/deploy.go
```diff
@@ -150,36 +139,15 @@ func deployFn(_ *cobra.Command, _ []string) error {
...
+	// TODO
 	// log new version availability info if ready
-	version.NewVerNotification(vCh)
+	// version.NewVerNotification(vCh)
 
 	// print table summary
 	return inspect.PrintContainerInspect(containers, deployFormat)
 }
-
-// setupCTRLCHandler sets-up the handler for CTRL-C
```
carlmontanari, line 165: replaced by cmd/context.go basically same thing, just copypastad from c9s for the new one

clab/clab.go
```diff
@@ -65,121 +63,79 @@ type CLab struct {
...
-		}
-
-		if isClabAdmin, err := utils.UserInUnixGroup(currentUser.Username, "clab_admins"); err == nil && isClabAdmin {
-			c.customOwner = owner
-		} else if owner != "" {
-			log.Warn("Only users in clab_admins group can set custom owner. Using current user as owner.")
-		}
-		return nil
+// NewContainerLab function defines a new container lab.
+func NewContainerLab(opts ...ClabOption) (*CLab, error) {
```
carlmontanari, line 67: everything in clab/clab was just reorganization -- options and topology things moved to their respective files, then the New funcs moved up to near the struct for locality/sanity

cmd/deploy.go
```diff
@@ -84,16 +81,11 @@ func init() {
 }
 
 // deployFn function runs deploy sub command.
-func deployFn(_ *cobra.Command, _ []string) error {
+func deployFn(cobraCmd *cobra.Command, _ []string) error {
```
carlmontanari, line 84: cobraCmd to not shadow package name. this is now where the root signal handled context lives and as such we gotta nicely pass it all along everywhere

cmd/version/check.go
```diff
@@ -3,120 +3,34 @@ package version
...
 
 // checkCmd defines a version check command.
 var checkCmd = &cobra.Command{
 	Use:   "check",
 	Short: "Check if a new version of containerlab is available",
-	RunE: func(_ *cobra.Command, _ []string) error {
+	RunE: func(cobraCmd *cobra.Command, _ []string) error {
 		// We'll use a short 5-second timeout for the remote request
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		ctx, cancel := context.WithTimeout(cobraCmd.Context(), 5*time.Second)
```
carlmontanari, line 17: same deal here, same 5s context, just rooted in our main signal handled context

cmd/version/check.go
```diff
@@ -3,120 +3,34 @@ package version
...
 		defer cancel()
 
-		// Version check function returns a channel
-		vCh := make(chan string, 1) // buffered to avoid potential goroutine leak
+		m := GetManager()
```
carlmontanari, line 20: this bit now just fetches the "version manager" which is a singleton thingy that lives behind a sync.once and an interface so its not easily accessible. then the GetLatestVersion just accepts the context (with the timeout) and ticks to either get the version from the web or the timeout whichever happens first -- basically same flow as the channel stuff from before but little less passing things around, little more version manager dude runs in the background and we just wait based on contexts

cmd/version/manager.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+var (
+	managerInstance     Manager   //nolint:gochecknoglobals
+	managerInstanceOnce sync.Once //nolint:gochecknoglobals
+)
+
+// InitManager initializes the version manager, we have an explicit init so we can capture the root
+// context to ensure appropriate cancellation during the version fetching if necessary.
+func InitManager(
```
carlmontanari, line 23: this dude gets called from the pre exec func so it should be always available to the cli -- and since it shouldn't ever matter to a future http-able clab this should be chill here (I think)

cmd/version/manager.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+}
+
+func (m *manager) run(ctx context.Context) {
+	if m.fetchDisabled {
+		return
+	}
+
+	client := &http.Client{
```
carlmontanari, line 115: all this is mostly the same, just formatting/var naming

main.go
```diff
@@ -5,15 +5,19 @@
...
 
 	"github.com/charmbracelet/fang"
 	"github.com/srl-labs/containerlab/cmd"
 )
 
 func main() {
-	if err := fang.Execute(context.TODO(), cmd.RootCmd); err != nil {
+	ctx, _ := cmd.SignalHandledContext()
+
+	cmd.RootCmd.SetContext(ctx)
```
carlmontanari, line 17: actually should check this -- we may not need to do the SetContext here since I would assume (but havent checked yet!) that fang would do that for us. in any case, here is where the context is rooted for all the cmd pkg

cmd/context.go
```diff
@@ -0,0 +1,39 @@
...
+		sig := <-sigs
+		log.Infof("received signal '%s', canceling context", sig)
+
+		cancel()
+
+		sig = <-sigs
+		log.Infof("received signal '%s', exiting program", sig)
+
+		os.Exit(1)
+	}()
```
hellt: it seems that the logic is not the same with the new ctrlc handler

previously, when the context was cancelled, the destroy op was kicked in to remove the partially deployed lab and leave no traces:

```go
		// when interrupted, destroy the interrupted lab deployment
		cleanup = false

		if err := destroyFn(destroyCmd, []string{}); err != nil {
			log.Errorf("Failed to destroy lab: %v", err)
		}

		os.Exit(1) // skipcq: RVV-A0003
```

carlmontanari: ack, fixing!

PR #2665, opened 2025-07-15
fixes #2523 support for --name flag with clab destroy command

### LINKED ISSUE #2523: Support --name in destroy
I’m using Containerlab to teach networking. Students' solutions are tested via GitLab CI, and for each submission, a new lab is created with a unique name.

The `--name` argument in deploy is very convenient for setting this unique name. But there’s no equivalent option in destroy. As a result, I have to patch the topology file just to adjust the name, which feels a bit awkward.

The change should be quite small. If you're open for it I am happy to contribute.

benjoe1126:
fix #2523 
Added --name flag support for destroy subcommand

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -139,6 +139,9 @@ func destroyFn(_ *cobra.Command, _ []string) error {
 		if keepMgmtNet {
 			opts = append(opts, clab.WithKeepMgmtNet())
 		}
+		if common.Name != "" {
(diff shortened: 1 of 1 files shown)


hellt:
thanks @benjoe1126 
I have added a test in https://github.com/srl-labs/containerlab/commit/233a4d7a789606b55b8e42da6d3b565a9d9133b3 and it passes.

PR #2670, opened 2025-07-16
include generic VM kind in kind overview

JensErat:
`generic_vm` was missing in the kind table, just found it by accident scrolling through the details.


hellt:
thank @JensErat

PR #2683, opened 2025-07-19
fixed destroy command --name bug

benjoe1126:
fixed bug where destroy with --name would try to delete even if no match is found

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -75,7 +75,7 @@ func destroyFn(_ *cobra.Command, _ []string) error {
 	switch {
 	case !all:
 		log.Debug("topology file", "file", common.Topo)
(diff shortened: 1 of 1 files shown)


cmd/destroy.go
```diff
@@ -159,6 +162,10 @@ func destroyFn(_ *cobra.Command, _ []string) error {
 			opts = append(opts, clab.WithLabName(common.Name))
 		}
 
+		if common.Name != "" {
```
vista-: Duplicated code (see right above)?

vista- (changes requested):

cmd/destroy.go
benjoe1126: Oops, you're right, fixed it

vista- (approved):
lgtm

PR #2691, opened 2025-07-21
Fix panic if per-stage command execution fails

giorio94:
Currently, containerlab panics if one of the per-stage commands exits with a non-zero code, due to the subsequent access to the uninitialized exec result. Let's get this fixed by always propagating the exec result regardless of the exit code, and avoiding accessing it in case of other errors.


hellt:
Thanks @giorio94

PR #2692, opened 2025-07-21
Introduced IXR specific configs in template

sacckth:
- Adding const vars specific to IXR
- Some Cleanup of unused vars


nodes/sros/configs/grpc.cfg
```diff
@@ -0,0 +1,5 @@
+configure system grpc admin-state enable
+configure system security user-params local-user user "admin" access grpc true
+configure system grpc allow-unsecure-connection
+configure system grpc gnmi auto-config-save true
+configure system grpc rib-api admin-state enable
```
hellt, line 5: @sacckth give this branch a go please... until we have a test bed to have it auto tested

sacckth: tested on a few IXR-nodes. Looks good to me.

PR #2695, opened 2025-07-22
Feat/decouple cmd inspect

carlmontanari:


.golangci.yml
```diff
@@ -67,8 +68,16 @@ linters:
             - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
             - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
             - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
+    importas:
+      alias:
+        # we can add more layers here (containerlab$1$2 etc.) as we tidy/get more consistent
```
carlmontanari, line 73: obv up to the project how crazy we go here, for *me* personally... pedantic though it may be (and maybe annoyingly verbose) I think being really explicit w/ import aliases and having things be fully qualified is really nice for readability

clab/deploy.go
```diff
@@ -0,0 +1,281 @@
+package clab
```
carlmontanari: deploy file is almost entirely unchanged things, just moved from clab.go

clab/destroy.go
```diff
@@ -0,0 +1,77 @@
+package clab
```
carlmontanari: destroy also almost entirely unchanged (if not entirely) and just moved from clab.go

clab/exec.go
```diff
@@ -0,0 +1,59 @@
+package clab
```
carlmontanari: annnnnnd also exec thing just moved from clab.go

clab/inventory.go
```diff
@@ -18,7 +18,7 @@ import (
 
 const NornirPlatformNameSchemaEnvVar = "CLAB_NORNIR_PLATFORM_NAME_SCHEMA"
 
-//go:embed inventory_ansible.go.tpl
+//go:embed assets/inventory_ansible.go.tpl
```
carlmontanari, line 21: all packages that had assets I just stuffed into an "assets" directory in their package to make it stupid obvious and to not have non ".go" files breaking up the flow of things in the main package

clab/list.go
```diff
@@ -0,0 +1,180 @@
+package clab
```
carlmontanari: moved all listing related things from the clab struct here

clab/list.go
```diff
@@ -0,0 +1,180 @@
...
+
+		containerInterfaces.Interfaces = append(containerInterfaces.Interfaces, &ifaceDetails)
+	}
+	log.Debugf("Fetched %v interfaces for %v", len(interfaces), containerInterfaces.ContainerName)
+
+	return &containerInterfaces, nil
+}
+
+// ListContainersInterfaces list interfaces of all given containers.
+func (c *CLab) ListContainersInterfaces(
```
carlmontanari: the interface things were taken from inspect (mostly) and stuffed here. it makes sense to me at least to have clab struct hold these, and it helps to get cmd to be doing as little as possible (easier for future api-y things and/or clab server(s))

clab/options_clab.go
```diff
@@ -169,7 +173,7 @@ func WithTopoFromLab(labName string) ClabOption {
 		filter := []*types.GenericFilter{
 			{
 				FilterType: "label",
-				Field:      clabels.Containerlab,
+				Field:      containerlablabels.Containerlab,
```
carlmontanari: lots of this type of thing -- just the importas w/ aliases to be hella specific and explicit -- things like "labels' or "exec" or w/e could be in a bunch of packages we use -- this makes it super clear its *our* labels/exec/etc.

cmd/destroy.go
```diff
@@ -49,7 +49,8 @@ func init() {
 	destroyCmd.Flags().BoolVarP(&common.Graceful, "graceful", "", false,
 		"attempt to stop containers before removing")
 	destroyCmd.Flags().BoolVarP(&all, "all", "a", false, "destroy all containerlab labs")
-	destroyCmd.Flags().BoolVarP(&yes, "yes", "y", false, "auto-approve deletion when used with --all (skips confirmation prompt)")
+	destroyCmd.Flags().BoolVarP(&yes, "yes", "y", false,
```
carlmontanari, line 52: some of this was just gofumpt + the line fixer thing, some I just broke out args onto line per arg where I saw some long ass lines

cmd/inspect/inspect_interfaces.go
```diff
@@ -121,75 +109,13 @@ func inspectInterfacesFn(_ *cobra.Command, _ []string) error {
 		return nil
 	}
 
-	err = printContainerInterfaces(ctx, containers, interfacesFormat)
```
carlmontanari, line 124: most of these "removes" in inspect/inspect_itnerfaces moved to the clab list file

nodes/sros/sros.go
```diff
@@ -669,17 +671,17 @@ func (n *sros) createSROSFiles() error {
...
 		}
 		log.Debug("SR OS license copied", "src", n.Cfg.License, "dst", licPath)
 	}
-	utils.CreateDirectory(path.Join(n.Cfg.LabDir, n.Cfg.Env[envNokiaSrosSlot]), 0777)
-	utils.CreateDirectory(path.Join(n.Cfg.LabDir, n.Cfg.Env[envNokiaSrosSlot], "config"), 0777)
-	utils.CreateDirectory(path.Join(n.Cfg.LabDir, n.Cfg.Env[envNokiaSrosSlot], configCf1), 0777)
-	utils.CreateDirectory(path.Join(n.Cfg.LabDir, n.Cfg.Env[envNokiaSrosSlot], configCf2), 0777)
-	utils.CreateDirectory(path.Join(n.Cfg.LabDir, n.Cfg.Env[envNokiaSrosSlot], configCf3), 0777)
-	utils.CreateDirectory(path.Join(n.Cfg.LabDir, n.Cfg.Env[envNokiaSrosSlot], configStartup), 0777)
+	utils.CreateDirectory(path.Join(n.Cfg.LabDir, n.Cfg.Env[envNokiaSrosSlot]), 0o777)
```
carlmontanari, line 679: the 0777 -> 0o777 is gofumpt not me (which we should enforce!) but also its better this way :)

carlmontanari:
like the last one, unit tests seem happy locally, clab-api-server tests seem happy as well. lots of *things* happening but the majority of this is just some consistency things (importas for example) and reorganization to (hopefully!) be more clear. that one mean podman robot test seems to be quite upset (after re-running failed a few times) can check that out more tomorrow/later but I think this should be mostly good for review!

.golangci.yml
```diff
@@ -67,8 +68,16 @@ linters:
...
+        - pkg: github.com/srl-labs/containerlab/(\w+)
+          alias: containerlab$1
```
hellt, line 75: I wonder if makeing the alias `clab$1` would make it less-ish verbose?

`containerlaberrors -> claberrros`  
`containerlabruntime -> clabruntime`

carlmontanari: yep we can do that, and in fact I started w/ that, but then `clabclab` felt bad. I think in reality clab package *should* be something like `manager` or `runner` or something like that -- then all the stuttering goes away in either case and also I think its more representative of what it really is -- that package is really the "manager" (or w/e you want to call it) managing clab topologies etc.

so, yeah however you wanna do it lemme know and ill make it so. options I think are:

1. re-alias and have clabclab be a thing
2. leave as is
3. rename the whole package

hellt: lets call it `core` as discussed in the chat

PR #2707, opened 2025-07-23
SRSIM test debug

hellt:
sim tests are stalling. When enabling tmate I saw the running `clab version` hangs indefinitely.
Seems like runners are borked for this tests. Not sure why the specific runners are being selected for srsim tests, though...


hellt:
and now it seems to pass =)
gh actions are acting again

PR #2710, opened 2025-07-23
allow to use existing config to boot srsim

sacckth:
On the first release, the code to allow to use existing config wasn't present. 
Also did some minor renaming of variables to give better sense to the whole config generation functions

### DIFF
--- a/nodes/sros/sros.go
+++ b/nodes/sros/sros.go
@@ -54,7 +54,7 @@ const (
 	configCf3                 = "config/cf3"
 	configCf2                 = "config/cf2"
 	configCf1                 = "config/cf1"
-	startupCfgFName           = "config.cfg"
+	startupCfgName            = "config.cfg"
 	envNokiaSrosSlot          = "NOKIA_SROS_SLOT"
 	envNokiaSrosChassis       = "NOKIA_SROS_CHASSIS"
 	envNokiaSrosSystemBaseMac = "NOKIA_SROS_SYSTEM_BASE_MAC"
@@ -240,12 +240,12 @@ func (n *sros) PreDeploy(_ context.Context, params *nodes.PreDeployParams) error
 				envNokiaSrosSlot, n.Cfg.Env[envNokiaSrosSlot])
 		}
 		// add the config specific mounts
-		cfgPath := filepath.Join(n.Cfg.LabDir, slot, configStartup)
+		cfgStartupPath := filepath.Join(n.Cfg.LabDir, slot, configStartup)
 		cf1Path := filepath.Join(n.Cfg.LabDir, slot, configCf1)
 		cf2Path := filepath.Join(n.Cfg.LabDir, slot, configCf2)
 		cf3Path := filepath.Join(n.Cfg.LabDir, slot, configCf3)
 		n.Cfg.Binds = append(n.Cfg.Binds,
-			fmt.Sprint(cfgPath, ":", cfgDir, ":rw"),
+			fmt.Sprint(cfgStartupPath, ":", cfgDir, ":rw"),
 			fmt.Sprint(cf1Path, ":", cf1Dir, "/:rw"),
 			fmt.Sprint(cf2Path, ":", cf2Dir, "/:rw"),
 			fmt.Sprint(cf3Path, ":", cf3Dir, "/:rw"),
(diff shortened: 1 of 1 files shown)


nodes/sros/sros.go
```diff
@@ -54,7 +54,7 @@ const (
 	configCf3                 = "config/cf3"
 	configCf2                 = "config/cf2"
 	configCf1                 = "config/cf1"
-	startupCfgFName           = "config.cfg"
+	startupCfgName            = "config.cfg"
```
hellt, line 56: lets add a test? so that we know it works and won't break with subsequent change?

can be a precreated clab dir with a config file present there, or a lab that is deployed, some change is made via netconf or gnmi to emulate an ad-hoc change, then destroyed **without** the cleanup flag, then started again and it is expected that the change introduced by a user is visible

sacckth: Ok, Let's track more testing in this PR? #2711 
I tested this manually and agree we should not break what exists today. But the additional tests will require a little bit of extra time while I familiarize more with robot

nodes/sros/sros.go
```diff
@@ -699,53 +705,61 @@ func (n *sros) createSROSFilesConfig() error {
...
-			return err
+	if nodeConfigExists(cf3CfgFile) { // Found config on CF3, using that instead
+		log.Infof("Using existing config file (%s) instead of applying a new one", cf3CfgFile)
+		if err := utils.CopyFile(cf3CfgFile, cfgStartupFile, 0o644); err != nil {
+			return fmt.Errorf("startupCfg copying src %s -> dst %s failed: %v", cf3CfgFile, cfgStartupFile, err)
 		}
+	} else { // Use startup folder to generate config and use that to boot node
+		if n.Cfg.StartupConfig != "" && !isPartial { // User provides startup config
+			log.Debug("Reading startup-config", "node", n.Cfg.ShortName, "startup-config",
+				n.Cfg.StartupConfig, "isPartial", isPartial)
```
hellt: you should not need to use `nodeConfigExists` function. The config generation is done by a common function of the DefaultNode - see https://github.com/srl-labs/containerlab/blob/cf740cb8e61872bcb18b7eb4ffdef45c0c4deb10/nodes/srl/srl.go#L529

This common function checks if the config exists and takes into account all flags that influence config generation

sacckth: **TDLR** there has to be a way to check if file A exists so we can overwrite   A-->B 
A-startup config
B-running config

**Long answer**
This one has a catch, and I don't see how I can use that function here easily. Let me explain and perhaps we can think on how we can implement this better:
SR-SIM can take its config from two different directories:
- `/nokia/config/config.cfg` 
- `/home/sros/flash3/config.cfg`

The `/nokia/config` folder has precedence wrt the flash3 dir and I mount it to the host using this binding:
 `<clab-name>/<slot>/config/startup/config.cfg  <---> /nokia/config/config.cfg`

Now, every time we do `admin save` only the file under `/home/sros/flash3/config.cfg` will be updated.
The binding for this is:
`<clab-name>/<slot>/config/cf3/config.cfg <---> /home/sros/flash3/config.cfg`

So when users  update their configs, only the cf3 folder will reflect this changes, the startup folder will stay the same always.

What I try to do in this section is to check whether the file`/home/sros/flash3/config.cfg` exists and, if it does,  overwrite  `/nokia/config/config.cfg` with the former so we can use such file to boot next time.


A possible solution is to get rid of the binding to `/nokia/config`. In this case, the startup config file will be mounted directly on the `/home/sros/flash3/` and we don't use the startup folder at all. Disadvantages, the config file will always change its contents and ordering whenever we do admin save, and this can be annoying to some. I found it more convenient to have startup and running config in different folders. Specially because our startup config can be a merge of various partial configs generated prior to deployment.

hellt: I think you should get rid of one, indeed.
There is no other NOS where we deal with two different config locations, SR-SIM being the first one does not sit well with me.

We actually do want the host systems (clab host) to reflect the state of the persistent config. Every time someone does `wr mem` equivalent updates the file on the clab host that is mounted to the NOS.

The key here is that the original startup config is not modified, it is copied to the node dir inside the lab dir, and the latter is mounted to the container. So the original startup config that is used in the clab file is not overwritten whenever we do `admin save`, it is its copy that sits in `clab-<labname>/<nodename>` is.

nodes/sros/sros.go
```diff
@@ -718,30 +714,25 @@ func (n *sros) createSROSFilesConfig() error {
...
-			"startup-config", string(n.startupCliCfg))
-
-		if err := utils.CreateFile(cfgPath, string(n.startupCliCfg)); err != nil {
-			return fmt.Errorf("failed to create startup-config file %s for node %s failed: %v", cfgPath, n.Cfg.ShortName, err)
+		// Adds partial config if present
+		err = n.addPartialConfig()
+		if err != nil {
+			return err
 		}
+		cfgTemplate = string(n.startupCliCfg)
```
hellt, line 725: @sacckth this is what I meant by a single source of the startup config file - https://github.com/srl-labs/containerlab/pull/2710/commits/ffc11330297076c1455de57ebfe59cd8f93abe51

tests pass, feel free to check some manual scenarios. If all good, ping me and I will merge it in

sacckth:
tested, Ok to merge. Bedankt!

hellt (approved):

PR #2715, opened 2025-07-24
Fix SRSIM save command

steiler:


hellt:
fyi `make clint` will run a golangci lint locally

hellt (approved):

PR #2720, opened 2025-07-25
fix: use context w/ timeout rather than root for final version check post deploy

carlmontanari:
title. my bad. :)


carlmontanari:
resolves #2721

hellt (approved):

PR #2729, opened 2025-07-30
Feat/decouple cmd destroy

carlmontanari:
very draft still, lotta cleaning to do. also dont think it fully works yet (mostly dir cleanup and such lacking rn)


carlmontanari:
cancelled ci since I know shit is busted :P will revisit soon!

cmd/version/upgrade.go
```diff
@@ -12,7 +12,7 @@ import (
 	"os/exec"
 
 	"github.com/spf13/cobra"
-	"github.com/srl-labs/containerlab/cmd/common"
```
carlmontanari, line 15: totally nuked cmd/common here. it felt unnecessary and w/ some of the detangling of things really became unnecessary

cmd/version/upgrade.go
```diff
@@ -21,7 +21,7 @@ const downloadURL = "https://github.com/srl-labs/containerlab/raw/main/get.sh"
 var upgradeCmd = &cobra.Command{
 	Use:     "upgrade",
 	Short:   "upgrade containerlab to latest available version",
-	PreRunE: common.CheckAndGetRootPrivs,
+	PreRunE: utils.CheckAndGetRootPrivs,
```
carlmontanari, line 24: the cmd/common priv related stuff moved to utils since... felt util-like

cmd/deploy.go
```diff
@@ -158,5 +157,5 @@ func deployFn(cobraCmd *cobra.Command, _ []string) error {
 	m.DisplayNewVersionAvailable(versionCheckContext)
 
 	// print table summary
-	return inspect.PrintContainerInspect(containers, deployFormat)
+	return PrintContainerInspect(containers, deployFormat)
```
carlmontanari, line 160: guess this could be private but prolly doesn't matter

cmd/destroy.go
```diff
@@ -38,279 +26,104 @@ var destroyCmd = &cobra.Command{
...
+	PreRunE: utils.CheckAndGetRootPrivs,
 	RunE:    destroyFn,
 }
 
 func init() {
 	RootCmd.AddCommand(destroyCmd)
 	destroyCmd.Flags().BoolVarP(&cleanup, "cleanup", "c", false,
 		"delete lab directory. Cannot be used with node-filter")
-	destroyCmd.Flags().BoolVarP(&common.Graceful, "graceful", "", false,
+	destroyCmd.Flags().BoolVarP(&gracefulShutdown, "graceful", "", false,
```
carlmontanari, line 37: renamed graceful -> gracefulShutdown to be more explicit

cmd/destroy.go
```diff
@@ -38,279 +26,104 @@ var destroyCmd = &cobra.Command{
...
-	// cleanup doesn't make sense with node-filter
-	if len(common.NodeFilter) != 0 && cleanup {
+func destroyFn(cobraCmd *cobra.Command, _ []string) error {
+	if cleanup && len(nodeFilter) != 0 {
 		return fmt.Errorf("cleanup cannot be used with node-filter")
 	}
 
-	var err error
-	var labs []*core.CLab
-	ctx, cancel := context.WithCancel(context.Background())
```
carlmontanari, line 69: rejoice in how many lines of things here are not in cmd now :P -- to be fair majority of this now lives in core but... that seems like a better place for it

cmd/destroy.go
```diff
@@ -38,279 +26,104 @@ var destroyCmd = &cobra.Command{
...
+	destroyOptions := []core.DestroyOption{
+		core.WithDestroyMaxWorkers(maxWorkers),
+		core.WithDestroyNodeFilter(nodeFilter),
 	}
-	log.Warn("The following labs will be removed:", "labs", sb.String())
 
-	warningStyle := lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("1")) // red color (ansi code 1)
-	prompt := "Are you sure you want to remove all labs listed above? Enter 'y', to confirm or ENTER to abort: "
-	fmt.Print(warningStyle.Render(prompt))
+	if keepMgmtNet {
```
carlmontanari, line 93: added options_destroy.go (in core) and did typical functional option pattern here. so cmd basically just deals w/ appending opts and shipping them to a core method... should be nicer/easier for core to be used in api server or similar in this way

cmd/exec.go
```diff
@@ -79,18 +77,18 @@ func execFn(_ *cobra.Command, _ []string) error {
...
-		filters = append(filters, types.FilterFromLabelStrings(labFilter)...)
+	if topoFile != "" {
+		listOptions = append(
+			listOptions,
+			core.WithListLabName(c.Config.Name),
+		)
 	}
 
-	resultCollection, err := c.Exec(ctx, execCommands, core.NewExecOptions(filters))
+	resultCollection, err := c.Exec(ctx, execCommands, listOptions...)
```
carlmontanari, line 91: got rid of options_exec (in core) entirely... was literally just for filters, so just passing the options for listing here instead

cmd/inspect.go
```diff
@@ -121,31 +120,24 @@ func inspectFn(cobraCmd *cobra.Command, _ []string) error {
...
 
-	if common.Topo != "" {
+	if topoFile != "" {
 		// List containers defined in the topology file
 		containers, err = c.ListNodesContainers(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("failed to list containers based on topology: %s", err)
 		}
 	} else {
+		var listOptions []core.ListOption
```
carlmontanari, line 131: all the filter creation stuff now lives behind list options. this centralizes all the generic filter creation and hides it behind dumb simple options

cmd/root.go
```diff
@@ -14,16 +14,23 @@ import (
...
-	"github.com/srl-labs/containerlab/cmd/inspect"
 	"github.com/srl-labs/containerlab/cmd/version"
 	"github.com/srl-labs/containerlab/git"
 	"github.com/srl-labs/containerlab/utils"
 )
 
 var (
-	debugCount int
-	logLevel   string
+	debugCount       int
```
carlmontanari, line 23: centralized all those global vars (and stuff from common)

core/deploy.go
```diff
@@ -26,7 +26,7 @@ func (c *CLab) Deploy(ctx context.Context, options *DeployOptions) ([]runtime.Ge
 
 	log.Debugf("lab Conf: %+v", c.Config)
 	if options.reconfigure {
-		_ = c.Destroy(ctx, uint(len(c.Nodes)), true)
+		_ = c.destroy(ctx, uint(len(c.Nodes)), true)
```
carlmontanari, line 29: the "old" destroy method is now private and called by the "new" one that does all the logic that cmd used to do.

core/destroy.go
```diff
@@ -1,15 +1,213 @@
...
+	"github.com/srl-labs/containerlab/nodes"
+	"github.com/srl-labs/containerlab/runtime"
 	"github.com/srl-labs/containerlab/runtime/ignite"
+	"github.com/srl-labs/containerlab/types"
+	"github.com/srl-labs/containerlab/utils"
 )
 
-// Destroy the given topology.
-func (c *CLab) Destroy(ctx context.Context, maxWorkers uint, keepMgmtNet bool) error {
+func (c *CLab) Destroy(ctx context.Context, options ...DestroyOption) (err error) {
```
carlmontanari, line 24: this should all look familiar. just the cmd/destroy stuff moved here mostly w/ some general tidying

core/destroy.go
```diff
@@ -1,15 +1,213 @@
...
+	if len(errs) != 0 {
+		return fmt.Errorf("error(s) occurred during the deletion. Check log messages")
+	}
+
+	return err
+}
+
+// creates a mostly cloned version of the current c but set to the new topology, and with the
+// necessary steps (mgmt network things) handled preparing the new CLab for destruction.
+func (c *CLab) cloneForDestroy(ctx context.Context, topo, labDir string, opts DestroyOptions) (*CLab, error) {
```
carlmontanari: "clone" is perhaps poorly named -- basically just spawning a new clab instance for each topo like before. originally I was going to just copy things but runtime initialization is complicated and I revered to the normal creation process again

core/destroy.go
```diff
@@ -1,15 +1,213 @@
...
+	for _, container := range containers {
+		topoFile, ok := container.Labels[containerlablabels.TopoFile]
+		if !ok {
+			continue
+		}
+
+		topos[topoFile] = filepath.Dir(container.Labels[containerlablabels.NodeLabDir])
+	}
+
+	defer func() {
```
carlmontanari, line 66: rather than having cleanup take place above and below this point just deferring it all in one shot here. error should still be returned because of the named return bits

core/options_destroy.go
```diff
@@ -0,0 +1,72 @@
+package core
```
carlmontanari, line 1: same functional option pattern stuff for destroy. nothing fancy

core/options_list.go
```diff
@@ -0,0 +1,151 @@
...
+	containerlabLabelExists bool
+	cliArgs                 []string
+}
+
+// NewListOptions returns a new destroy options object.
+func NewListOptions() *ListOptions {
+	return &ListOptions{}
+}
+
+func (o *ListOptions) ToFilters() []*types.GenericFilter {
```
carlmontanari, line 28: and here is the hopefully nicer way to simplify/centralize the filter creation stuff

types/topo_paths.go
```diff
@@ -33,20 +33,27 @@ var clabTmpDir = filepath.Join(os.TempDir(), ".clab")
...
+	varsFile           string
 	labDir             string
 	topoName           string
 	externalCACertFile string // if an external CA certificate is used the path to the Cert file is stored here
 	externalCAKeyFile  string // if an external CA certificate is used the path to the Key file is stored here
 }
 
 // NewTopoPaths constructs a new TopoPaths instance.
-func NewTopoPaths(topologyFile string) (*TopoPaths, error) {
+func NewTopoPaths(topologyFile, varsFile string) (*TopoPaths, error) {
```
carlmontanari, line 44: rather than taking common.VarsFile shipping it straight into and storing it w/ the TopoPaths which just made sense to me

core/destroy.go
```diff
@@ -73,5 +273,149 @@ func (c *CLab) Destroy(ctx context.Context, maxWorkers uint, keepMgmtNet bool) e
...
+			},
+			{
+				FilterType: "label",
+				Field:      containerlablabels.Containerlab,
+				Operator:   "=",
+				Match:      c.Config.Name,
+			},
+		}
+
+		containers, err := c.globalRuntime().ListContainers(ctx, toolFilter)
```
carlmontanari, line 363: I wonder why this is doing list against the runtime rather than using clab list? probably also should reswizzle the label stuff here to use the new option stuff

core/destroy.go
```diff
@@ -73,5 +273,149 @@ func (c *CLab) Destroy(ctx context.Context, maxWorkers uint, keepMgmtNet bool) e
...
+			case <-ctx.Done():
+				return
+			}
+		}
+	}
+
+	// start concurrent workers
+	wg.Add(int(workers))
+
+	for i := range workers {
```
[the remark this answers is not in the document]
carlmontanari: go 1.24 so ... ignoring

hellt:
the failing srsim test was first added here https://github.com/srl-labs/containerlab/pull/2710

But it might be more generic than just the srsim. It tests that if the lab was deployed first and then a config change was made that is persisted (the result of gnmi set in the test) then one can destroy the lab (without cleanup) and deploy it again and expect to see the change made earlier to be. 

In other words, the node should pick up the config in it's labdir if one is present at the deploy time

core/options_destroy.go
```diff
@@ -0,0 +1,72 @@
...
+	maxWorkers     uint
+	keepMgmtNet    bool
+	graceful       bool
+	all            bool
+	terminalPrompt bool
+	cleanup        bool
+	nodeFilter     []string
+}
+
+// NewDestroyOptions returns a new destroy options object.
```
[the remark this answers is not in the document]
carlmontanari: I mean... sure, but pointless nit imo (and roman knows how much I love pointless nits) and it does indeed return a destroy options object. the type is already there.

core/options_list.go
```diff
@@ -0,0 +1,151 @@
...
+// ListOptions represents the options for listing containers.
+type ListOptions struct {
+	labName                 string
+	nodeName                string
+	toolType                string
+	containerlabLabelExists bool
+	cliArgs                 []string
+}
+
+// NewListOptions returns a new destroy options object.
```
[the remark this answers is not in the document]
carlmontanari: stupid nit but it should say list not destroy, fixing :)

core/list.go
```diff
@@ -14,6 +14,29 @@ import (
...
+		ctrs, err := r.ListContainers(ctx, opts.ToFilters())
+		if err != nil {
+			return containers, fmt.Errorf("could not list containers: %v", err)
+		}
+		containers = append(containers, ctrs...)
+	}
+	return containers, nil
+}
+
+// ListContainers lists all containers using provided filter.
```
hellt: Is this a leftover?

core/list.go
```diff
@@ -14,6 +14,29 @@ import (
...
+func (c *CLab) ListContainersOLD(
```
hellt: Is this a leftover?

carlmontanari: yes good shout, fixed!

core/destroy.go
hellt: good question that sent me down some memory lane, and I _think_ it was because we might have had in clab some nodes that are not backed by actual containers. This was maybe the case for the multi-node KinD clusters where a single KinD container had several nodes in clab representation (control and worker nodes).

Buuuut, I am so not sure about this :D maybe it is worth checking this up, since we have a kind test in the suites, so if anything goes sideways we hopefully catch it

hellt: ah, wait wait

in this case the tools containers are launched ad hoc, and not defined in the topology. So when clab runs, it reads the topo file and doesn't find the tools containers. Hence we tagged these containers with labels and runtime returns them based on labels

This goes back to your idea for NewClab to actually be smart and read its state when initializing

carlmontanari: Ahhh ok yep makes sense… well maybe cmd/tools* clean up is next and we can dig in!

hellt (approved):

PR #2731, opened 2025-08-02
add viper integration to destroy.go to preserve mgmt-net with env var

jbernardini:
Added viper binding to CLAB_KEEP_MGMT_NET env variable to modify destroy cmd to always keep the management network when set to true. When set to false or not present, current behavior remains. 

[Added the variable to /etc/security/pam_env.conf for system wide inheritance for all users, regardless of shell - on most modern linux distributions.] 

Deployed on my remote ssh host and tested, with true/false/na under multiple users. Tested in cli and vscode integration.

Also willing to take a crack at the documentation to explain how and why to use it - if you've setup a manual docker network for macvlan support on the management network.


jbernardini:
This PR is only meant to include commit fd2f4eb6178895be38932ca66ee42311a1c42e13. will close and new.

PR #2736, opened 2025-08-04
add viper integration to destroy.go to set env var keep mgmt-net

jbernardini:
Added viper binding to CLAB_KEEP_MGMT_NET env variable to modify destroy cmd to always keep the management network when set to true. When set to false or not present, current behavior prevails.

[Added the variable to /etc/security/pam_env.conf for system wide inheritance for all users, regardless of shell - on most modern linux distributions.]

Deployed on my remote ssh host and tested, with true/false/na under multiple users. Tested in cli and vscode integration.

Also willing to take a crack at the documentation to explain how and why to use it - if you've setup a manual docker network for macvlan support on the management network.


cmd/destroy.go
```diff
@@ -54,6 +55,10 @@ func init() {
 	destroyCmd.Flags().UintVarP(&maxWorkers, "max-workers", "", 0,
 		"limit the maximum number of workers deleting nodes")
 	destroyCmd.Flags().BoolVarP(&keepMgmtNet, "keep-mgmt-net", "", false, "do not remove the management network")
+	// Bind Cobra flag to Viper key
+	viper.BindPFlag("keep-mgmt-net", destroyCmd.Flags().Lookup("keep-mgmt-net"))
+	// Bind environment variable CLAB_KEEP_MGMT_NET to same key
+	viper.BindEnv("keep-mgmt-net", "CLAB_KEEP_MGMT_NET")
```
hellt, line 61: In order for this to not be a one-off feature, I think a better way is to adopt viper in general and use AutoEnv viper's feature to get a consistent loading of vars from envs across the full configuration scope of clab

```
envPrefix := "CLAB"
SetEnvPrefix(envPrefix)
SetEnvKeyReplacer(strings.NewReplacer("/", "_", "-", "_"))
AutomaticEnv()
```

hellt:
The options have been reworked by @carlmontanari and I think now the road is clear to add viper support for all of them using AutomaticEnv

hellt:
added full viper support in #2868

PR #2737, opened 2025-08-04
Reset the extra hosts on slots that contain networkmode: "container:..."

steiler:


nodes/sros/sros.go
```diff
@@ -1033,11 +1037,12 @@ func (n *sros) SaveConfig(ctx context.Context) error {
 		}
 		fqdn = n.Cfg.LongName
 	}
-	return n.saveConfig(ctx, fqdn)
+	return n.saveConfigWithAddr(ctx, fqdn)
 }
 
-func (n *sros) saveConfig(_ context.Context, addr string) error {
-	err := netconf.SaveRunningConfig(addr,
```
sacckth, line 1040: Had to rename this because Lint was complaining this is confusing and I agree

sacckth (approved):

nodes/sros/sros.go
```diff
@@ -1092,11 +1097,11 @@ func CheckPortWithRetry(host string, port int, timeout time.Duration, maxRetries
 // MgmtIPAddr returns ipv4 or ipv6 management IP address of the node.
 // It returns an error if neither is set in the node config.
 func (n *sros) MgmtIPAddr() (string, error) {
-	if n.Cfg.MgmtIPv4Address != "" {
-		return n.Cfg.MgmtIPv4Address, nil
-	} else if n.Cfg.MgmtIPv6Address != "" {
+	if n.Cfg.MgmtIPv6Address != "" {
```
hellt: This would benefit from the switch I think

nodes/sros/sros.go
```diff
@@ -229,6 +229,10 @@ func (n *sros) PreDeploy(_ context.Context, params *nodes.PreDeployParams) error
 
 	n.topologyName = params.TopologyName
 
+	if strings.HasPrefix(n.Cfg.NetworkMode, "container:") {
+		n.Cfg.ExtraHosts = nil
```
hellt, line 233: Can we add a test for the number of host entries to the suite with srsim? 
This would cover this feature

hellt (changes requested):

nodes/sros/sros.go
steiler: done

nodes/sros/sros.go
sacckth: just added a new test with mgmt network options, and the check you ask.

PR #2743, opened 2025-08-06
Fix Distributed SR-SIM port mapping when using Grouped topology

michelredondo:
In Distributed SR-SIM scenarios involving grouped topologies, port bindings should not be configured for line cards.


sacckth:
Closing this one and adding your commits on #2745 for some reason I wasn't able to commit to this PR.

PR #2756, opened 2025-08-12
WIP: Add startup-config For Linux Kind

hawkinsw:


hawkinsw:
Thank you, again, for building such a robust tool! As I was working on some labs, I found that having startup-config support would be useful for nodes with the Linux kind. This PR contains the necessary functionality, but it (obviously) lacks documentation and testing. I would _love_ to add those as long as you think that the feature is something that you want. I don't want to waste your time reviewing something annoying. So, if you think that having this feature would be useful, I will continue the work! 

No matter what, I appreciate your willingness to engage with contributors!
Will

nodes/linux/linux.go
```diff
@@ -96,6 +99,47 @@ func (n *linux) PostDeploy(ctx context.Context, _ *nodes.PostDeployParams) error
...
+	if n.Cfg.StartupConfig != "" {
+		log.Debugf("Running user-specified startup config for Linux '%s' node", n.Cfg.ShortName)
+
+		startupConfigFilename := path.Base(n.Cfg.StartupConfig)
+
+		// read the startup config file from the host disk.
+		c, err := os.ReadFile(n.Cfg.StartupConfig)
+		if err != nil {
+			return err
+		}
```
hellt, line 111: hi @hawkinsw 

I feel like this feature is not that relevant for the Linux kind. The linux kind is for a generic container, therefore it can not rely on `bash` for instance, to be available.

What you achieve here is realised via generic concepts of `binds` and `exec`. For example, if you want to execute with `bash` a file `my-commands.sh` that you have on disk locally, you would define your node as:

```yaml
mylinuxnode:
  kind: linux
  binds:
    - my-commands.sh:/tmp/my-commands.sh
  exec:
    - bash /tmp/my-commands.sh
```

hawkinsw: I tried to accomplish the same functionality using the `cmd` feature, but ... I was hamstrung because what I _really_ wanted to do was give an ip address to the interface automatically configured by the topology. What I specified at `cmd` was executed before that interface was available. I didn't try `exec`. I will do that!

Of course, as you said, I realize that the Linux container kind is generic and, obviously, `bash` is not a pre-requisite. That was the reason I couched it in terms of a more speculative PR. 

If the `exec` will work to accomplish my goal, then that's great. I'm not sure how I missed it!

PR #2760, opened 2025-08-15
Refactor/cmd opts struct

carlmontanari:


carlmontanari:
damnit. my jj foo is bad. deleting till I figure out how to not check in things I dont mean to w/ jj :P

PR #2761, opened 2025-08-15
Refactor/cmd opts struct

carlmontanari:


carlmontanari:
ok well I clearly still dont know how to use jj entirely correctly but.... other than the weird empty commits I think this is good to review...

obviously another horrendous pr to review and im not sure commenting in the pr will help so I will just say the *general* vibe of this was removing all the global vars in cmd package and replacing them w/ structs holding the options. 

this also involved finishing flattening out that version bits up into the parent cmd package.

there were a few linter complaints addressed as well (switch vs if/else stuff).

.golangci.yml
```diff
@@ -33,8 +33,11 @@ linters:
     - lll
     - mnd
     - nestif
+    - perfsprint
     - prealloc
     - revive
+    - wrapcheck
+    - wsl # deprecated, eventually we'll enable wsl_v5
```
hellt: maybe a typo?

Image

carlmontanari: not a typo but probably we can remove. I was just comparing to other config file I have and because I use default: all I had to disable "old" wsl. will address when I go look at the deep source stuff!

cmd/root.go
```diff
@@ -7,70 +7,343 @@ package cmd
...
-	PersistentPreRunE: preRunFn,
-	Aliases:           []string{"clab"},
+var optionsInstance *Options //nolint:gochecknoglobals
+
+func GetOptions() *Options {
+	if optionsInstance == nil {
+		optionsInstance = &Options{
+			Global: &GlobalOptions{
+				Timeout:  120 * time.Second,
+				LogLevel: "info",
```
hellt: maybe a nit, but would it make sense to move all options-related stuff to the `cmd/options.go` file to keep them cozy together in an isolated space?

cmd/root.go
```diff
@@ -7,70 +7,343 @@ package cmd
...
+		saveCmd,
+		toolsCmd,
+	}
 }
 
-func init() {
-	RootCmd.SilenceUsage = true
-	RootCmd.PersistentFlags().CountVarP(&debugCount, "debug", "d", "enable debug mode")
-	RootCmd.PersistentFlags().StringVarP(&topoFile, "topo", "t", "",
+func Entrypoint() (*cobra.Command, error) {
```
hellt, line 40: how would you describe in the function comment what the Entrypoint() does?

utils/file.go
```diff
@@ -95,7 +95,33 @@ func CopyFile(ctx context.Context, src, dst string, mode os.FileMode) (err error
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		// should only err on repeated calls to close anyway
+		_ = out.Close()
+	}()
+
+	return CopyFileContents(ctx, src, out)
```
hellt, line 124: looks like you removed this logic from the CopyFileContents func presumably to make it do less while being more closer to the function name?

I am fine with that, but I wonder if all the mechanics of ensuring that dst exists and uid/gid are set correctly should be factored out in a func?

Because right now this code piece is copied here in the `CopyFile` func as well as in the `core/config.go`

cmd/root.go
carlmontanari: ill add a comment when I push some stuff but basically this "Entrypoint creates/initializes the root cobra command -- the "entrypoint" for containerlab" if that makes sense. at least thats how I got to that name hah

cmd/root.go
carlmontanari: yep, doing!

cmd/tools_cert.go
```diff
@@ -166,40 +167,40 @@ func signCert(_ *cobra.Command, _ []string) error {
...
-	clabutils.CreateDirectory(path, 0o777) // skipcq: GSC-G302
+	clabutils.CreateDirectory(o.ToolsCert.Path, 0o777) // skipcq: GSC-G302
 
 	err = nodeCert.Write(
-		filepath.Join(path, certNamePrefix+clabtypes.CertFileSuffix),
-		filepath.Join(path, certNamePrefix+clabtypes.KeyFileSuffix),
-		filepath.Join(path, certNamePrefix+clabtypes.CSRFileSuffix))
+		filepath.Join(o.ToolsCert.Path, o.ToolsCert.CANamePrefix+clabtypes.CertFileSuffix),
+		filepath.Join(o.ToolsCert.Path, o.ToolsCert.CANamePrefix+clabtypes.KeyFileSuffix),
+		filepath.Join(o.ToolsCert.Path, o.ToolsCert.CANamePrefix+clabtypes.CSRFileSuffix))
```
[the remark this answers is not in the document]
carlmontanari: good job bot :D

cmd/version_manager.go
```diff
@@ -40,19 +40,19 @@ func InitManager(
...
+func getVersionManager() Manager {
+	if versionManagerInstance == nil {
 		// be defensive, should be initialized during first cmd spin up though
 		panic(
 			"version manager instance is nil, this should not happen",
 		)
 	}
 
-	return managerInstance
+	return versionManagerInstance
```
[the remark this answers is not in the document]
carlmontanari: reasonable comment but all private funcs that should only be used in cmd and will be caught in every single test case so I think this is nicer from an ergonomics perspective

utils/file.go
carlmontanari: yeah... there is a lot of duplication here and also some small funcs that just literally do what std lib does ex: 

```
func ReadFileContent(file string) ([]byte, error) {
	// try to read and return file content, or return an error
	b, err := os.ReadFile(file)
	return b, err
}
```

:) 

so yes, probably all this deserves more of a tune up but. was just trying to do as little as possible since theres already too much crap going on here haha. happy to keep tuning this one up however you want in this pr or a future one though, just lemme know what ya wanna do!

hellt: yeah, I am fine revisiting this. Will create an issue so that it keeps bothering us coming back to it some time later https://github.com/srl-labs/containerlab/issues/2764

hellt (approved):

carlmontanari:
woohoo! now my jj test drive can be over 🤣

PR #2763, opened 2025-08-17
Add n9kv to schema

### LINKED ISSUE #2762: Cisco N9Kv (cisco_n9kv) needs to be added as a valid kind type
The Containerlab schema does not include cisco_n9kv as a valid kind type even though this is a valid node type. Screenshot below (from vscode) confirms this:

Image

kaelemc:
Closes #2762


hellt:
Thanks. Can you move it next to existing Cisco kind definitions in the schema so it is all grouped nicely?

PR #2765, opened 2025-08-17
Add tailscale tools command

kaelemc:
heavily inspired from sshx/gotty


kaelemc:
Usage:

`clab tools tailscale attach --auth-key xyz`

Unfortunate part is you must still approve the submet route in the tailscale web UI.. and it seems the ephemeral setting takes a random amount of time for the node to actually disappear from your tailnet.

PR #2791, opened 2025-09-04
Link subnets

kaelemc:
Allow for IP addressing on an interface to be defined inside the topology file

```yaml
name: test
topology:
  nodes:
    n1:
      kind: nokia_srlinux
      image: ghcr.io/nokia/srlinux:24.10
    n2:
      kind: nokia_srlinux
      image: ghcr.io/nokia/srlinux:24.10
  links:
    - endpoints: [n1:e1-1, n2:e1-1]
      vars:
        ipv4: [n1:192.168.0.1/24, n2:192.168.0.2/24]
    - endpoints: [n1:e1-2, n2:e1-2]
      vars:
        ipv4:
          n1: 192.168.2.1/24
          n2: 192.168.2.2/24
    - type: veth
      endpoints:
        - node: n1
          interface: e1-3
          vars:
            ipv4: 192.168.3.1/24
        - node: n2
          interface: e1-3
          vars:
            ipv4: 192.168.3.2/24
```


kaelemc:
@hellt Please give your thoughts before I continue and write some docs etc.

nodes/srl/srl_default_config.go.tpl
```diff
@@ -59,6 +59,18 @@ set / interface ethernet-{{ $ep.Slot }}/{{ $ep.Port }} breakout-mode num-channel
 set / interface ethernet-{{ $ep.Slot }}/{{ $ep.Port }}/{{ $ep.BreakoutNo }} admin-state enable
   {{- end }}
 
+  {{- /* If per-endpoint IPv4/IPv6 are provided, configure them on subinterface 0 */}}
+  {{- if or $ep.IPv4 $ep.IPv6 }}
+set / interface ethernet-{{ $ep.Slot }}/{{ $ep.Port }}{{ if ne $ep.BreakoutNo "" }}/{{ $ep.BreakoutNo }}{{ end }} subinterface 0 admin-state enable
```
hellt: how about we figure out the v4 and v6 full port names in the Go code and leave template to just check if it is not empty and set per family
Less template - more sanity

kaelemc: thanks. I am not fully clear on what you mean - do you mean to effectively handle the building of the command in the go? ie. the template is just passed the full `set / interface ...` command as a string?

hellt: the buildout of the port structure can be done in go
so instead of 

```
set / interface ethernet-{{ $ep.Slot }}/{{ $ep.Port }}{{ if ne $ep.BreakoutNo "" }}/{{ $ep.BreakoutNo }}{{ end }} subinterface 0 admin-state enable
```

it can be something like

```
set / interface {{ $ep.InterfaceName }} subinterface 0 admin-state enable
```

where InterfaceName is calculated in Go taking into account if that's a breakout or not. Maybe it would make things clearer and more testable

steiler:
would this not be better suited in the extended link format?
this is again another slice that needs quite some parsing.
In the extended format you have the association to the nodes already...
```
links:
  - type: veth
    endpoints:
      - node: <NodeA-Name>                  # mandatory
        interface: <NodeA-Interface-Name>   # mandatory
        mac: <NodeA-Interface-Mac>          # optional
        ipv4: <ip>                                         # -> NEW FIELD
        ipv6: <ip>                                         # -> NEW FIELD
      - node: <NodeB-Name>                  # mandatory
        interface: <NodeB-Interface-Name>   # mandatory
        mac: <NodeB-Interface-Mac>          # optional
        ipv4: <ip>                                         # -> NEW FIELD
        ipv6: <ip>                                         # -> NEW FIELD
    mtu: <link-mtu>                         # optional
    vars: <link-variables>                  # optional (used in templating)
    labels: <link-labels>                   # optional (used in templating)
```
The endpoint in all the linktypes is a common struct, so you would only need to define these explicit fields once.

I see only just now that you've also defined them on the extended struct but also as the vars... I would not do that.
The shorthand for is just to create quickly some topology, but defining ips I'd argue is extended stuff so add it as poper fields to the endpoint struct. Thats nice and clean. my 2 cents.

kaelemc:
There is no issue from me to remove the shorthand, but I would argue it's easier for people to pickup with the shorthand as it's likely familiar to them with how they write their topos already. 

The subject of the vars, I leave to @hellt to decide, my original proposal was as you have in your sample yaml.

steiler:
Are we fine with a single IPv4 and single IPv6 or would that even be slices?
would you also need options for unnumbered or so?

links/endpoint.go
```diff
@@ -51,6 +53,12 @@ type EndpointGeneric struct {
 	Link     Link
 	MAC      net.HardwareAddr
 	randName string
+	Vars     *EndpointVars
+}
+
+type EndpointVars struct {
+	IPv4 string `yaml:"ipv4,omitempty"`
+	IPv6 string `yaml:"ipv6,omitempty"`
```
hellt, line 61: @kaelemc @steiler 

I think we will have to have two workflows. One is using the raw links (short version) as this is the one that most people use when they compose the clab topology by hand.

We will also have to make it work with the extended links format. And it would make sense to have this mode used by vscode extension where a user creates a topology via UI.

I agree with @steiler we should have it support ipv6-unnumbered and it would be great if we could set the link subnet in the top `.config` struct and have the `auto` mode where links get IP from it and auto-assign.



wdyt?

kaelemc: Sure, we still keep under `vars`? I agree with Markus it's better to have it as a top level thing like the MAC address.

Unnumbered & auto assign is fine by me. 

Regarding multiple IPs on the interface, this one is maybe in terms of how different NOSes are handling IPv4.. ie: if some vendor is allowing only primary and secondary address - and many addreses are supplied, how are we picking what is what? (IPv6 is probably ok on this)

I also want to add this for Linux kind so we can stop with the `exec:` for ip link commands, and with the `auto` you propose, this is probably awesome.. except I know some vrnetlab kinds are using `linux`.. so how do we want to handle this? maybe something under the nodeconfig to disable the IP configuration?

steiler: @hellt I do think this is a good way to promote the extended format. The shorthand format is fine for defining really quickly two endpoints, but this is already an extended usecase. In my view it makes sense to make people use the extended format for this.

kaelemc: @hellt thoughts on this?

hellt: @kaelemc keep it as is. Maybe just add the ipv6 option as well? Then we can merge and experiment with this to figure out what works and what needs improvement

hellt:
the failing srsim test has been fixed in `main`

PR #2795, opened 2025-09-06
Chore/more linters and housekeeping

carlmontanari:


carlmontanari:
more pedantic stuff but wanted to start enforcing more linters on core before getting feisty in there for maybe future works... 

that last commit may be more controversial? I dunno. the global debug count and stuff just felt clunky, and didnt want to see that transport package imported into cmd package as that felt dirty -- I also think it may be nicer to have a Send / Configure method on the clab struct rather than just creating the clab instance then kinda mostly ignoring it then just reaching into that transport package to send stuff. but... ive not thought further than that, so just sharing for comment/consideration!

PR #2796, opened 2025-09-07
Chore/centralize constants

carlmontanari:


carlmontanari:
more housekeeping. starting to try to centralize costs. if/when we enable revive linter this will get even more clear imo -- for example there are loads of exported consts in links that could maybe be unexported since they are not used anywhere else, whereas there are others that *are* used in other packages and those latter ones are the good candidates to move to the constants package imo. I think I got a handful of them found and moved but for sure more will fall out w/ additional linters complaining about things :)

PR #2802, opened 2025-09-08
SR-SIM components enhancement

kaelemc:
We want to add some nice SR-SIM integration into the vscode extension. 

The current env var based configuration of the SR-SIM nodes aren't so clean, and schema validation isn't easy either.

I propose some new fields under the `components` for `xiom`, `sfm` and `mda`. These still map to env vars, but being dedicated fields we can make some nice enhancements to the schema (which are also added in this PR; I have pulled all the types from the public 7x50 YANG models).

I have also updated the license policy to prevent a case where if no license is provided, the containers exit and the deploy cmd hangs.

Please give your thoughts. Docs & tests will come soon.

https://github.com/user-attachments/assets/77454f8b-64b5-47e8-a153-1f378290ec5d

cc: @FloSch62 @asadarafat

### DIFF
--- a/docs/manual/kinds/sros.md
+++ b/docs/manual/kinds/sros.md
@@ -158,7 +158,9 @@ Data interfaces need to be configured with IP addressing manually using the SR O
 
 ## SR-SIM variants
 
-The SR-SIM can emulate different hardware platforms as explained in the [SR-SIM Installation, deployment and setup guide](https://documentation.nokia.com/sr/25-7/7750-sr/titles/sr-sim-installation-setup.html). These variants can be set using the `type` directive in the clab topology file or by overriding the different available environment variables such as the ones for the chassis (`NOKIA_SROS_CHASSIS`) or card (`NOKIA_SROS_CARD`).  
+The SR-SIM can emulate different hardware platforms as explained in the [SR-SIM Installation, deployment and setup guide](https://documentation.nokia.com/sr/25-7/7750-sr/titles/sr-sim-installation-setup.html). These variants can be set using the `type` directive in the clab topology file, or by overriding the different available environment variables such as the ones for the chassis (`NOKIA_SROS_CHASSIS`) or card (`NOKIA_SROS_CARD`).  
+
+
 Users can then use environment variables to change the default behavior of a given container. If there is a conflict between the `type` field in the topology file and an environment variable in the topology file, the environment variable will take precedence.
 
 > When `type` is not provided in the topology file, SR-SIM will start as `SR-1` platform.
@@ -169,7 +171,9 @@ If the chosen platform is chassis-based, the SR-SIM deployment needs to be done 
 
 We call non-chassis-based systems like SR-1, SR-1s integrated variants. As these systems have a fixed form factor, they run as a single container and are represented as a single node in the topology file.
 
-Besides setting the `type` to drive the platform selection, users can then modify some of the default settings on a per-node basis using the environment variables. For example, to change the default MDA, like shown in the example below.
+Besides setting the `type` to drive the platform selection, users can then modify some of the default settings on a per-node basis using the `components` configuration on the node, or the environment variables. 
+
+For example, to change the slot 1 MDA, refer to the example below:
 /// tab | Integrated SR-SIM
 
 ```yaml
@@ -183,7 +187,24 @@ topology:
 ```
 
 ///
-/// tab | Integrated SR-SIM with override
+/// tab | Integrated SR-SIM with components based override
+
+```yaml
+topology:
+  nodes:
+    sr-sim1:
+      kind: nokia_srsim
+      image: nokia_srsim:25.7.R1
+      type: SR-1
+      license: /opt/nokia/sros/license.txt
+      components:
+        - mdas:
+          - slot: 1
+            type: me12-100gb-qsfp28 # override default MDA type in slot 1
+```
+
+///
+/// tab | Integrated SR-SIM with env-var based override
 
 ```yaml
 topology:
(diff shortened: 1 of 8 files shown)


steiler:
Did we reach the point where it makes sense to add some custome unmarshalling like done with the links?
The concept there is to unmarshall a partial struct, that mainly contains the type, this can then be used to unmarshall into the correct node struct. Overall that would be some work to do it for all the kinds that we have but it would be the strongest types solution.

kaelemc:
> how will the card types be mapped to the right node? Can you elaborate a bit more on how did you generate this list?

@sacckth Thanks for taking a look. I am not clear by what you mean on "how will the card types be mapped to the right node?".

The card type handling hasn't changed from the intial implementation, simply the env vars for sfm, iom, and mda become their own field now. There is no checking in the schema against compatability of the different components.

I pulled all the types from [here](https://github.com/nokia/7x50_YangModels/blob/master/latest_sros_25.7/nokia-types-card.yang). I hope it is correct.

kaelemc:
> Did we reach the point where it makes sense to add some custome unmarshalling like done with the links? The concept there is to unmarshall a partial struct, that mainly contains the type, this can then be used to unmarshall into the correct node struct. Overall that would be some work to do it for all the kinds that we have but it would be the strongest types solution.

I think only SR-SIM would benefit from this currently. I would prefer not to make that kind of change yet, I intended for this to be a simple enabler so we can make the SR-SIM experience in the VS Code extension better :)

hellt:
I like this change. @sacckth do you have any concerns after the provided clarifications

sacckth:
> I like this change. @sacckth do you have any concerns after the provided clarifications

What would be the mapping for an integrated node with no components?
e.g.
```yaml
topology:
  nodes:
    sr-sim1:
      kind: nokia_srsim
      image: nokia_srsim:25.7.R1
      type: SR-1
      license: /opt/nokia/sros/license.txt
      env:
        NOKIA_SROS_MDA_1: me12-100gb-qsfp28 # override default MDA type in slot 1
  ```
  
  I tried this and it didn't work:
  ```yaml
  nodes:
    srsim1:
      kind: nokia_srsim
      type: SR-1
      components:
        - slot: A
          mda: 
            - slot: 1
              type: me12-100gb-qsfp28
```

The rest, looks good. We were planning to make some changes on the docs, and add support to some new node types for the upcoming SROS release. Shall I make a new PR for this?

hellt:
@sacckth 

```yaml
nodes:
  srsim1:
    kind: nokia_srsim
    type: SR-1
    components:
      - slot: A
      - slot: 1
        mda: 
          - slot: 1
            type: me12-100gb-qsfp28
```

What about this? Maybe `slot: A` is not even required

sacckth:
thats two containers as slot has two elements in array. SR-1 is just one container, same as other integrated sr-sims. In addition, the default slot for these containers is always "A"


________________________________

docs/manual/kinds/sros.md
```diff
@@ -379,21 +379,21 @@ topology:
...
-          env:
-            NOKIA_SROS_SFM: m-sfm6-7/12
-            NOKIA_SROS_CARD: iom5-e # (1)!
-            NOKIA_SROS_MDA_1: me6-100gb-qsfp28
-            NOKIA_SROS_MDA_2: me16-25gb-sfp28+2-100gb-qsfp28
+          type: iom5-e
+          sfm: m-sfm6-7/12
+          mda:
+            - slot: 2
+              type: me6-100gb-qsfp28
```
hellt: @kaelemc can we also add support for MDAs for the non-component based (single slot) routers like SR-1?

See comment from @sacckth https://github.com/srl-labs/containerlab/pull/2802#issuecomment-3292988894

kaelemc: yep, embrassing oversight from my end.

I am thinking either:

```yaml
components:
  - slot: A
    mda:
      - slot: 1
        type: xyz
      - slot: 2
        type: ...
```

OR.. the following where if defined in the below manner. we imply integrated if this format is detected and apply the env vars directly to the 'root' container.

```yaml
components:
  mda:
    - slot: 1
      type: xyz
    - slot: 2
      type: ...
```

sacckth: I feel that the second option would make the schema a little messy(map, instead of array). So can we use the first example with only array elements? This would make the schema more uniform. We can maybe default the linecard slot to A in case we want to save a couple of bytes:
```yaml
components:
  - slot: A #default can be omitted only when array has one element
    mda:
      - slot: 1
        type: xyz
      - slot: 2
        type: ...
```

kaelemc: @hellt @sacckth Thanks for catching this one guys.

The below topology sample is now valid. Please review and leave thoughts. I shall update docs once I have your blessings.

```yaml
    sr1-01:
      kind: nokia_srsim
      type: sr-1
      components: 
        - slot: A # override integrated MDA explicit slot
          mda:
            - slot: 1
              type: me12-100gb-qsfp28

    sr1-02:
      kind: nokia_srsim
      type: sr-1
      components: 
         - mda: # override integrated MDA implicit slot
            - slot: 1
              type: me12-100gb-qsfp28
```

hellt: then I think the only thing that is left is a little doc addition?

sacckth: > @hellt @sacckth Thanks for catching this one guys.
> 
> The below topology sample is now valid. Please review and leave thoughts. I shall update docs once I have your blessings.
> 
> ```yaml
>     sr1-01:
>       kind: nokia_srsim
>       type: sr-1
>       components: 
>         - slot: A # override integrated MDA explicit slot
>           mda:
>             - slot: 1
>               type: me12-100gb-qsfp28
> 
>     sr1-02:
>       kind: nokia_srsim
>       type: sr-1
>       components: 
>          - mda: # override integrated MDA implicit slot
>             - slot: 1
>               type: me12-100gb-qsfp28
> ```

Almost there, as discussed just need to make sure that we honor the env variables above the `components` vars.

kaelemc: @hellt @sacckth fixed.. in terms of the docs. Shall we promote the components format more.. ie. move it up the page now that it applies for both integrated & distributed, remove the experimental status etc?

hellt: I am good with making components the first choice. If @sacckth feels the same we can do that

sacckth: Yes, I was actually planning to do this in a separate PR. But if you wanna go for it, just ping me once done so I can review as well.

kaelemc:
@sacckth @hellt please take a look at the updated docs, let me know your thoughts.

I am unsure on whether we still should refer to it as a "grouped topology" and "standard topology".. but I left it as is.

Made some other minor formatting tweaks and also changed the license part as with the new license policy behaviour the deploy can never happen unless license is provided.

fwiw here is what I am cooking on the vscode side with the updated schema :) 

https://github.com/user-attachments/assets/fad99f88-a090-41be-879d-ba96601cc536

sacckth:
> @sacckth @hellt please take a look at the updated docs, let me know your thoughts.
> 
> I am unsure on whether we still should refer to it as a "grouped topology" and "standard topology".. but I left it as is.
> 
> Made some other minor formatting tweaks and also changed the license part as with the new license policy behaviour the deploy can never happen unless license is provided.
> 
> fwiw here is what I am cooking on the vscode side with the updated schema :)
> 
>  2025-09-17_23-46-38.mp4

Full blessing

hellt (approved):

kaelemc:
Thanks @hellt @sacckth

PR #2811, opened 2025-09-16
Add short and json version output options

evilhamsterman:
To make version checks easier in scripting, this adds an option to output just the version or a json with the same data as in the default output.

Otherwise it just outputs the same format as before.

Example
```
$ bin/containerlab version --short
0.0.0
$ bin/containerlab version --json
{"version":"0.0.0","commit":"none","date":"unknown","repository":"https://github.com/srl-labs/containerlab","releaseNotes":"https://containerlab.dev/rn/0.0/"}
```

### DIFF
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -99,6 +99,10 @@ func GetOptions() *Options {
 				Port:           defaultVxlanPort,
 				DeletionPrefix: "vx-",
 			},
+			Version: &VersionOptions{
+				Short: false,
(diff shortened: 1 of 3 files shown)


evilhamsterman:
I know it's possible to grep and parse out the version from the existing format, but this makes it a bit easier, especially with something like a configuration system like Ansible.

hellt:
Thanks @evilhamsterman

PR #2813, opened 2025-09-18
tools/vxlan: Add source port option to VXLAN tool command

vista-:
This PR adds a 'source port' option for the VXLAN tool's create command. This is useful when going through NAT, limiting the VXLAN tunnel to a single port allows the other side to connect back even if port 4789 is not explicitly port-forwarded, and helps maintain the NAT session for the VXLAN tunnel. 

The port (destination port) option has been renamed to be consistent with the other option in both the CLI flags and the topology description.

The --port option/udp-port field is still functional, but has been marked as deprecated.

### DIFF
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -96,7 +96,8 @@ func GetOptions() *Options {
 			},
 			ToolsVxlan: &ToolsVxlanOptions{
 				ID:             defaultVxlanID,
-				Port:           defaultVxlanPort,
+				DstPort:        defaultVxlanPort,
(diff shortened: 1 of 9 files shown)


vista-:
Note: I looked around GitHub and I think I saw a single repository that used the vxlan endpoint type... so the change in the topology format isn't going to have a large blast radius. https://github.com/srl-labs/clab-io-draw will have to be updated.

kaelemc:
Looks good.. I will give it a try

PR #2815, opened 2025-09-19
SR-SIM XIOM components fix

kaelemc:


kaelemc:
I misunderstood XIOMs.. the topology now looks like this for XIOMS.. and it seems there is one case where you can have both normal MDA and XIOM (and it's nested MDA) on a card.. so I keep it flexible this way.

XIOM is always prefixed with X, so I feel safe to leave MDA and XIOM slot fields as integer.. Would you guys agree?

```yaml
nodes:
  srsim:
      kind: nokia_srsim
      type: sr-2s
      components: 
        - slot: A
          sfm: sfm-2s
        - slot: 1
          sfm: sfm-2s
          type: xcm-2s
          xiom:
            - slot: 1 # xiom 1/x1 -> maps to NOKIA_SROS_XIOM_X1 on the card 1 container
              type: iom-s-3.0t
              mda:
                - slot: 1 # mda 1/x1/1 -> maps to NOKIA_SROS_MDA_X1_1 on the card 1 container
                  type: ms18-100gb-qsfp28
```

kaelemc:
According to my new best friend no integrated model is supporting XIOM... should I leave this in as a 'just incase'.. or omit XIOM (and SFM) env var setting from integrated?

CleanShot 2025-09-19 at 23 37 13@2x

sacckth:
> According to my new best friend no integrated model is supporting XIOM... should I leave this in as a 'just incase'.. or omit XIOM (and SFM) env var setting from integrated?
> 
> CleanShot 2025-09-19 at 23 37 13@2x

Yeah, please avoid setting vars on nodes that don't require them. Similary, IXR nodes don't have XIOMS

kaelemc:
Happy from my side. Please give your thoughts @sacckth @hellt

PR #2818, opened 2025-09-20
Chore/types package cleanup

carlmontanari:


core/config.go
```diff
@@ -229,6 +229,11 @@ func (c *CLab) createNodeCfg( //nolint: funlen
 		Components:      c.Config.Topology.GetComponents(nodeName),
 	}
 
+	if nodeDef != nil {
```
carlmontanari, line 232: these parts (basically most of the changes in core/config.go) are lame. I feel like we have a bit of a silly pattern here where we have the config node then try to access the node def from the topo and it's just a bit wonky. I think the safety checks put in in this pr are fine for now but just reswizzling node instantiation to be safer in general in the future prolly would be good

types/node_definition.go
```diff
@@ -111,306 +114,20 @@ func (n *NodeDefinition) UnmarshalYAML(unmarshal func(any) error) error {
 	return nil
 }
 
-func (n *NodeDefinition) GetKind() string {
```
carlmontanari, line 114: all the "getters" is what started this adventure. just was looking to tidy up the types package for some maybe future refactoring (esp for making things nicer for c9s!) and felt this was a lil non-idiomatic and clunky. so basically all this has been removed/moved to topology which now has some generic funcs to handle safely doing the little cascading field check (i.e. node -> group -> kind -> defaults priority thing).

.golangci.yml
```diff
@@ -115,7 +115,6 @@ linters:
       - nodes
       - runtime
       - schemas
-      - types
```
carlmontanari, line 118: rejoice! more packages under linter authoritarianism!

utils/env.go
```diff
@@ -23,22 +23,22 @@ func ConvertEnvs(m map[string]string) []string {
 	return s
 }
 
-func mapify(i interface{}) (map[string]interface{}, bool) {
+func mapify(i any) (map[string]any, bool) {
```
carlmontanari, line 26: just using `any` since I saw older interface{} in a few places

types/types.go
```diff
@@ -69,9 +73,10 @@ func (m *MgmtNet) UnmarshalYAML(unmarshal func(any) error) error {
 
 	type MgmtNetWithDeprecatedFields struct {
 		MgmtNetAlias         `yaml:",inline"`
-		DeprecatedIPv4Subnet string `yaml:"ipv4_subnet,omitempty" json:"ipv4_subnet,omitempty"`
-		DeprecatedIPv6Subnet string `yaml:"ipv6_subnet,omitempty" json:"ipv6_subnet,omitempty"`
+		DeprecatedIPv4Subnet string `json:"ipv4_subnet,omitempty" yaml:"ipv4_subnet,omitempty"`
```
carlmontanari, line 76: this came from one of the linters. obviously superficial, but now all the things are json tags first then yaml so thats nice for consistency I guess

types/topology.go
```diff
@@ -53,74 +227,95 @@ func (t *Topology) GetKinds() map[string]*NodeDefinition {
...
 
 func (t *Topology) GetGroups() map[string]*NodeDefinition {
 	if t.Groups == nil {
 		return make(map[string]*NodeDefinition)
 	}
+
 	return t.Groups
 }
 
 func (t *Topology) GetNodeKind(name string) string {
```
hellt: I think this one does not use `getField` because it has some custom logic?

same for GetNodeType?

types/topology.go
```diff
@@ -691,77 +872,38 @@ func (t *Topology) ImportEnvs() {
...
-		// so that we don't have to check for nil in the caller
-		defaultDNS = &DNSConfig{}
-	}
-
-	return defaultDNS
-}
-
 // GetCertificateConfig returns the certificate configuration for the given node.
-func (t *Topology) GetCertificateConfig(name string) *CertificateConfig {
+func (t *Topology) GetCertificateConfig(nodeName string) *CertificateConfig {
```
hellt, line 856: can this benefit from `getField`? Probably not because again some custom logic?

carlmontanari: this one is/was because the merging. I think we could getField if we just wanted the first set value but this one merges them all how the string slice/map ones do and since it was the only one of this flavor I left it alone

types/topology.go
carlmontanari: soo... yeah, totally doable I think but not worth the squeeze because at the moment if we did it we would recurse and hit max frames :P obviously that can be fixed but that would make the magic generics getter thing less friendly than it already is which imo is not worth it. so I just pushed a tiny lil commit to do a bit more tidying but I think this feels reasonable imo

hellt (approved):

PR #2827, opened 2025-09-23
feat: add SR-OS component config generation to SR-SIM

### LINKED ISSUE #2823: Configure installed components for SR-SIM
If we define an SR-SIM in the topology using the components structure, we should also generate the configuration to set card/sfm/xiom/mda type, as well as admin enable these components.

kaelemc:
This is currently only for distributed types, why?

1. Most integrated nodes have preprovisioned cards/mdas
2. Some DO NOT, which makes it hard when a user wants to only override MDA, they should also defined card type.. but hold on a sec, overriden MDAs are defined under slot A as when only slot A is defined we hop into a special case where the env vars apply to the root node.. but in the config we need to provision card 1, and a card-type for the card 1 IOM.

So that is a bit messy for integrated right now.. I will check what types do/don't have a default config and work from there in another PR.

Have to test fully/add some tests for this.

closes #2823


sacckth:
I think this can get quite ugly quite fast.  That's why I avoided from the beginning to preconfigure cards. If you want to continue down this path, I suggest to make this feature optional by  adding a flag do enable this behavior.  Moreover, there are other configurations that you would have to include related to the power-shelf modules.

kaelemc:
Agreed that it can (will) get ugly.. I will check on the power shelf configs for each type, thanks for mentioning that. and yup, a flag is always good to add.

I still feel it is ok to proceed as even if the power shelf config would be omitted there is some net benefit provided to the UX.

nodes/sros/sros.go
```diff
@@ -1204,3 +1221,68 @@ func (n *sros) MgmtIPAddr() (string, error) {
...
+	for _, component := range components {
+		slot := strings.ToUpper(component.Slot)
+
+		if slot == "A" || slot == "B" {
+			// cpm -> skip
+			continue
+		}
+
+		if component.Type != "" {
+			config.WriteString(fmt.Sprintf("/configure card %s card-type %s admin-state enable\n", slot, component.Type))
```
sacckth: Can we put commands without "/" and see if it still works?

kaelemc: sure, I will try. Any reason as to why full context with the `/` is not preferred?

nodes/sros/sros_default_config.go.tpl
```diff
@@ -16,4 +16,6 @@ configure system security user-params local-user user "admin" public-keys rsa rs
 
 {{ range $index, $key := .SSHPubKeysECDSA }}
 configure system security user-params local-user user "admin" public-keys ecdsa ecdsa-key {{ subtract 32 $index }} key-value {{ $key }}
-{{ end }}
\ No newline at end of file
+{{ end }}
+{{/* Component configuration for cards, xioms, xiom-mdas and mdas */}}
+{{ .ComponentConfig }}
```
sacckth, line 21: If we are adding card configuration, we need to make sure power related configs are added. This only applies to some 7750 nodes, so you will need to add tests for pretty much all the distributed nodes.

kaelemc: I do not mind, as long as it's not every possible permutation right?

ie. we can just set a single power supply type and it should suffice for all MDA/XIOM/Card combinations?

nodes/sros/sros.go
sacckth: all the other configs are w/o full context, so in case we need to add more config later, and someone forgets the "/", this will fail misserably

sacckth:
> Agreed that it can (will) get ugly.. I will check on the power shelf configs for each type, thanks for mentioning that. and yup, a flag is always good to add.
> 
> I still feel it is ok to proceed as even if the power shelf config would be omitted there is some net benefit provided to the UX.

yeah, UX is good, but IMHO if we are going to solve a problem (i.e. help with the initial card config), we ought to make sure the node is functional. If  no power config is provided, cards won't come up"

nodes/sros/sros_default_config.go.tpl
sacckth: Not sure about this one, but I think PSU are per chassis type, e.g. 2s and 2se, you have to issue show commands to see which one  is it. Some examples were included [here](https://github.com/srl-labs/containerlab/tree/main/lab-examples/sr-sim)

nodes/sros/sros.go
hellt: imho, instead, all other commands should use `/` 
non root-based commands are a ticking bomb

nodes/sros/sros_default_config.go.tpl
hellt: the PSU types should be documented in the vrnetlab types as well https://github.com/srl-labs/vrnetlab/blob/master/nokia/sros/docker/launch.py#L125

nodes/sros/sros.go
kaelemc: I converted all other commands to be root context based. Let me know if you want this reverted.

nodes/sros/sros_default_config.go.tpl
hellt: @kaelemc I think we need to document what is configured for what types in the sros.md so that people know what they are gonna get when not configuring the nodes/shelves/etc

kaelemc: > @kaelemc I think we need to document what is configured for what types in the sros.md so that people know what they are gonna get when not configuring the nodes/shelves/etc

Things are only generated when `components:` are defined.. so I guess pretty much only the power supply config is generated if you configure the slots, but don't specify any card type, mda type etc.. like seen in the below topo.

```yaml
r1:
  kind: nokia_srsim
  image: ...
  components:
    - slot: A
    - slot: B
    - slot: 1
```

but yea, I will document this, as well as the overall config behaviour... just need to add a switch to disable it.. as well as adding some tests once I figure out all the correct configs.

kaelemc: @hellt kindly take a look. I think I'm ok from my side.

docs/manual/kinds/sros.md
```diff
@@ -267,6 +267,28 @@ Users can simplify the topology file with distributed SR-SIM nodes by using the
...
+You can disable this config generation behaviour by setting the `SROS_DISABLE_COMPONENT_CONFIG` env var on the base node.
+
+```yaml hl_lines="6-7"
+topology:
+  nodes:
+    sr-sim1:
+      kind: nokia_srsim
+      type: SR-7
+      env:
+        SROS_DISABLE_COMPONENT_CONFIG: "xyz"
```
hellt: @kaelemc please rename the env var to 

`CLAB_SROS_DISABLE_COMPONENT_CONFIG` as this is a clab scoped env var

docs/manual/kinds/sros.md
```diff
@@ -267,6 +267,28 @@ Users can simplify the topology file with distributed SR-SIM nodes by using the
...
+/// details | Disabling generated SR-OS configuration for `components`
+    type: tip
+You can disable this config generation behaviour by setting the `SROS_DISABLE_COMPONENT_CONFIG` env var on the base node.
+
+```yaml hl_lines="6-7"
+topology:
+  nodes:
+    sr-sim1:
+      kind: nokia_srsim
+      type: SR-7
```
hellt: I think our schema only works with lower cap type names

docs/manual/kinds/sros.md
```diff
@@ -267,6 +267,28 @@ Users can simplify the topology file with distributed SR-SIM nodes by using the
 
 Each card in the chassis can have it's card type, SFM, XIOM and MDA configured either using the relevant directives, or via environment variables.
 
+When using the `components` structure in the node configuration for a distributed node, containerlab will also generate the  configuration for the installed components in the chassis, as well as relevant power supply configuration[^5] to ensure the installed components come up.
```
hellt: we need an example here. Like have the same topo as in the details block below and show what will be injected in the config based on the specified components

kaelemc: @hellt done, please check and give feedback.

docs/manual/kinds/sros.md
```diff
@@ -784,7 +872,8 @@ The following labs feature Nokia SR OS (SR-SIM) node:
...
 [^3]: If needed, switches can be created using the clab kind `bridge` or using `iproute2` commands. MTU needs to be set to 9000 at least.
 [^4]: The word SHOULD is interpreted as [RFC2129](https://datatracker.ietf.org/doc/html/rfc2119) and [RFC8174](https://datatracker.ietf.org/doc/html/rfc8174). Links will come up as long as they are attached to the same Linux namespace.
-[^5]: This is a change from the [Vrnetlab](../vrnetlab.md) based vSIM where line cards and MDAs were pre-provisioned for some cases.
-[^6]: `~` is the home directory of the user that runs containerlab.
-[^7]: If a user wishes to provide a custom startup-config with public keys defined, then they should use key IDs from 1 onwards. This will minimize chances of key ID collision causing containerlab to overwrite user-defined keys.
-[^8]: See Github issue [#2741](https://github.com/srl-labs/containerlab/issues/2741)
+[^5]: Power configuration is only applied for sr-1s, 1se, 2s, 2se, 7s and 14s nodes. See [sros.go](https://github.com/srl-labs/containerlab/pull/2827/files#diff-ae71218e629cf2763a2702c67297cb2ade467276acff8f39973caf1a09731d94R142-R175) for more info.
```
hellt, line 878: for my bad memory, why power supplies are only required for these chassis, and not, say, for sr-7?

kaelemc: The other chassis have different configuration paths.. and to me it doesn't seem like you actually need to configure them for the compatible cards/mdas to come up, that is not to say I have tested every single permutation though.

```
*[ex:/configure chassis router chassis-number 1]
A:admin@sr-sim1-a# /show chassis power-supply 
===============================================================================
Chassis 1 Detail
===============================================================================
Power Supply Information
  Number of power supplies          : 2
  
  Power supply number               : 1
    Defaulted power supply type     : dc
    Power supply model              : pem
    Status                          : up
  
  Power supply number               : 2
    Defaulted power supply type     : dc
    Power supply model              : pem
    Status                          : up
===============================================================================
```

```
*[ex:/configure chassis router chassis-number 1]
A:admin@sr-sim1-a# power-supply 1 power-supply-type ?

 power-supply-type <keyword>
 <keyword> - (none|dc-single|ac-single|ac-multiple|auto)
 Default   - auto

    Power supply type

*[ex:/configure chassis router chassis-number 1]
```

hellt: I see; so some chassis require psu for cards to come up while others don't

kaelemc: essentially yes, that is my understanding.. for those SR-s/se chassis, they are not by-default provisioned with the 'maximum' power configuration, so quite a few cards/mdas will give you this insufficient power error and never come up.

I think for this config excerpt above, you can see the default setting for power-supply-type is `auto`.. presumably on a HW emulation level, the installed PSU is already 'maxxed out'.. or adjusting itself when certain mda/card is installed which require more power.. because we don't actually specify the HW installed PSU anywhere for any node type. but Mr @sacckth probably knows better than me.

tiago-amado: The vSIM/SR-SIM emulates exactly the real HW chassis and configs. The SR-7/12 chassis have integrated PSUs and we don't need to configure them. The SR-7s/14s have an external power-shelf, decoupled from the chassis, where we plug the PSUs, and we need to explicitly configure these modules. The same is required for vsim/SR-SIM.

kaelemc: So the configuration paths on the 7 & 12 don't do anything? There is no card/MDA combo which needs them provisioned? 

I might be hallucinating but I recall those PSUs weren't provisioned when spinning up a 7 with the default config.

tiago-amado: > So the configuration paths on the 7 & 12 don't do anything? There is no card/MDA combo which needs them provisioned?
> 
> I might be hallucinating but I recall those PSUs weren't provisioned when spinning up a 7 with the default config.

@kaelemc , my previous comment was about power related components only. All modular chassis require Card provisioning (IOM, MDA, XMA and SFMs. CPMs are never provisioned). In summary:
- SR7/12 - requires config for IOM, MDA, IMM and SFM (for SFM5 and above. Up to v4 it was integrated into the CPM and was not provisioned).
- SR-s family - requires config for XCM, xIOM-s, MDA-s, XMA-s, SFM-s, **_power-shelf and PSUs_**

kaelemc: @tiago-amado Sorry I wasn't clear.. what I meant is that, is there no possible card/mda combo on the 7/12 chassis that requires us to provision those power supplies (to prevent that 'Lo Power' error. etc)?

tiago-amado: @kaelemc , No. 
These chassis have the PSUs/APEQs directly plugged in the chassis, they don't have a power-shelf and will never have. These chassis support only up to FP4, they will not support further FP generations. We are discontinuing FP3 now, and  we will discontinue FP4 in ~5y from now. So within ~10y to ~15y most of these chassis will disappear. 
The SR-s architecture is different, it has a power-shelf module with a control card, and that's why we need to provision the power  related modules.

PR #2831, opened 2025-09-24
fix: SR-SIM don't lowercase the network node

kaelemc:
If I use a distributed SR-SIM topology with uppercase node names, the other components cannot join the netns of slot A because the name is lowered, and the container cannot be found.

rel https://github.com/srl-labs/containerlab/pull/2821

### DIFF
--- a/tests/13-srsim/07-srsim-mix.clab.yml
+++ b/tests/13-srsim/07-srsim-mix.clab.yml
@@ -72,7 +72,7 @@ topology:
             - slot: 2
               type: me3-400gb-qsfpdd
         - slot: 2
-    srsim11-a:
+    SRSIM11-a:
       kind: nokia_srsim
       type: sr-7
       startup-config: |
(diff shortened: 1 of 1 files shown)


nodes/sros/sros.go
```diff
@@ -486,8 +486,8 @@ func (n *sros) setupComponentNodes() error {
 		// the first node will create the namespace, so NetworkMode remains unchanged.
 		// all consecutive need to be attached to specifically that Namespace via NetworkMode
 		if idx > 0 {
-			componentConfig.NetworkMode = fmt.Sprintf("container:%s", strings.ToLower(
-				n.componentNodes[0].GetShortName()))
+			componentConfig.NetworkMode = fmt.Sprintf("container:%s",
+				n.componentNodes[0].GetShortName())
```
hellt, line 490: can you add or modify one of the tests to feature the mixed case name so that this is tested?

kaelemc: @hellt Done

hellt (approved):

PR #2834, opened 2025-09-24
fix: add component sorting to SR-SIM

kaelemc:
If I have a topology like below. It will never complete deploying, because the IP address is getting handed to the slot 1 , which is not CPM. Since the CPM doens't have an IP, the deploy keeps waiting for the CPM to get an IP (which never happens).

This PR is sorting the components so that slot A should always get the IP/host config and all component nodes attach to it's namespace.

```yaml
    sr1:
        kind: nokia_srsim
        image: nokia_srsim:25.7.r1
        license: srsim.txt
        type: sr-2s
        components:
            - slot: 1
              type: xcm-2s
              xiom:
              - mda:
                - slot: 1
                  type: ms18-100gb-qsfp28
                - slot: 2
                  type: ms18-100gb-qsfp28
                slot: 1
                type: iom-s-3.0t
            - slot: 2
              type: xcm-2s
              xiom:
              - slot: 2
                type: iom-s-3.0t
            - slot: A
              type: cpm-2s
            - slot: B
              type: cpm-2s
```
@hellt I would appreciate the 0.70.2 patch rel on this as well.

### DIFF
--- a/nodes/sros/sros.go
+++ b/nodes/sros/sros.go
@@ -16,10 +16,12 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"text/template"
 	"time"
+	"unicode"
 
 	"github.com/beevik/etree"
 	"github.com/brunoga/deep"
@@ -460,10 +462,36 @@ func (n *sros) DeleteNetnsSymlink() error {
 	return n.DefaultNode.DeleteNetnsSymlink()
 }
 
+// sortComponents ensure components are in order of
+// CPM then regular IOM/XCMs
+func (n *sros) sortComponents() {
+	slices.SortFunc(n.Cfg.Components, func(a, b *clabtypes.Component) int {
+		s1 := strings.ToUpper(strings.TrimSpace(a.Slot))
+		s2 := strings.ToUpper(strings.TrimSpace(b.Slot))
+
+		p1 := n.getSortOrder(s1)
+		p2 := n.getSortOrder(s2)
+
+		return p1 - p2
+	})
+}
+
+func (n *sros) getSortOrder(slot string) int {
+	r := rune(slot[0])
+	if unicode.IsLetter(r) {
+		return int(r) // A=65, B=66
+	} else {
+		num, _ := strconv.Atoi(slot)
+		return 1000 + num // 1001, 1002... always bigger than slot A/B
+	}
+}
+
 func (n *sros) setupComponentNodes() error {
 	if !n.isDistributedBaseNode() {
 		return nil
 	}
+
+	n.sortComponents()
 
 	// Registry, because it is not a package Var
 	nr := clabnodes.NewNodeRegistry()
--- a/nodes/sros/sros_test.go
+++ b/nodes/sros/sros_test.go
@@ -1,3 +1,67 @@
 package sros
 
-/* TO DO */
+import (
+	"testing"
+
+	clabtypes "github.com/srl-labs/containerlab/types"
+)
+
+func TestComponentSorting(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    []*clabtypes.Component
+		expected []string // expected slot order
+	}{
+		{
+			name: "random mix",
+			input: []*clabtypes.Component{
+				{Slot: "3"},
+				{Slot: "b"},
+				{Slot: "1"},
+				{Slot: "A"},
+				{Slot: "2"},
+			},
+			expected: []string{"A", "b", "1", "2", "3"},
+		},
+		{
+			name: "iom/xcm only",
+			input: []*clabtypes.Component{
+				{Slot: "3"},
+				{Slot: "1"},
+				{Slot: "2"},
+				{Slot: "10"},
+			},
+			expected: []string{"1", "2", "3", "10"},
+		},
+		{
+			name: "cpm only",
+			input: []*clabtypes.Component{
+				{Slot: "b"},
+				{Slot: "A"},
+			},
+			expected: []string{"A", "b"},
+		},
+	}
+
+	for _, c := range tests {
+		t.Run(c.name, func(t *testing.T) {
+			n := &sros{}
+			n.Cfg = &clabtypes.NodeConfig{
+				Components: c.input,
+			}
+
+			n.sortComponents()
+
+			if len(n.Cfg.Components) != len(c.expected) {
+				t.Fatalf("expected %d components, got %d", len(c.expected), len(n.Cfg.Components))
+			}
+
+			for i, expectedSlot := range c.expected {
+				actualSlot := n.Cfg.Components[i].Slot
+				if actualSlot != expectedSlot {
+					t.Errorf("expected slot %q, got %q", expectedSlot, actualSlot)
+				}
+			}
+		})
+	}
+}


sacckth:
From the docs:
When a distributed SR-SIM node is defined using components, we need to take into account the following:

Individual containers will be attached to the namespace of the 1st element of the components list: CPM-A in the above examples.

So, with this sorting you are always forcing the attachment to CPM-A... Which might not be ideal. If we are going to sort, then a variable or key needs to be introduced to control the container in which we are going to attach all the other namespaces.

nodes/sros/sros.go
sacckth, line 1: see my comment

kaelemc:
Alright, what is the possible usecase for attachment to non CPM containers instead ? 

The issue would still persist that the CPM doesn't get the IP address. I guess that logic would need to be adjusted too.

sacckth (changes requested):

sacckth:
> Alright, what is the possible usecase for attachment to non CPM containers instead ?
> 
> The issue would still persist that the CPM doesn't get the IP address. I guess that logic would need to be adjusted too.

Let's say you want to do a cpm-failover scenario. You can then attach all nodes to a linecard namespace or to CPM-B and do the failover usecase safelly.

kaelemc:
> > Alright, what is the possible usecase for attachment to non CPM containers instead ?
> > The issue would still persist that the CPM doesn't get the IP address. I guess that logic would need to be adjusted too.
> 
> Let's say you want to do a cpm-failover scenario. You can then attach all nodes to a linecard namespace or to CPM-B and do the failover usecase safelly.

Ok, understood. Thanks for clarifying.

kaelemc:
Please correct me if wrong but my understanding is the IP is going to the container who owns the NS.

This means the only valid nodes are A or B.. Otherwise if slot 1 for example would get the IP, the deploy flow fails as there is no way for CPM to have an IP on the mgmt interface right?

sacckth:
> Please correct me if wrong but my understanding is the IP is going to the container who owns the NS.
> 
> This means the only valid nodes are A or B.. Otherwise if slot 1 for example would get the IP, the deploy flow fails as there is no way for CPM to have an IP on the mgmt interface right?

This is something that was added with the `deployFabric()` function. A non-components install would  just not assign an IP. Unfortunately it seems that the changes in  the dependency_manager.go now break this on the also when not using components... This used to work:
```
    SRSIM11-a:
      kind: nokia_srsim
      type: sr-7
      startup-config: |
        /configure  sfm 1 sfm-type m-sfm6-7/12
        /configure  sfm 2 sfm-type m-sfm6-7/12
        /configure card 1 card-type iom5-e
        /configure card 2 card-type iom5-e
        /configure card 1 mda 1 mda-type me6-100gb-qsfp28
        /configure card 2 mda 1 mda-type me6-100gb-qsfp28
        /configure card 2 mda 2 mda-type me16-25gb-sfp28+2-100gb-qsfp28
        /configure port 1/1/c1 admin-state enable
        /configure port 1/1/c1 connector breakout c4-10g
        /configure port 2/1/c1 admin-state enable
        /configure port 2/1/c1 connector breakout c4-10g
        /configure port 1/1/c1/1 admin-state enable
        /configure port 1/1/c1/2 admin-state enable
        /configure port 2/1/c1/1 admin-state enable
        /configure port 2/1/c1/2 admin-state enable
        /configure router "Base" interface "system" admin-state enable
        /configure router "Base" interface "system" ipv4 primary address 100.0.0.11
        /configure router "Base" interface "system" ipv4 primary prefix-length 32
        /configure router "Base" interface "ptp1" admin-state enable
        /configure router "Base" interface "ptp1" port 1/1/c1/1
        /configure router "Base" interface "ptp1" ipv4 primary address 10.0.0.11
        /configure router "Base" interface "ptp1" ipv4 primary prefix-length 24
        /configure router "Base" interface "ptp2" admin-state enable
        /configure router "Base" interface "ptp2" port 2/1/c1/1
        /configure router "Base" interface "ptp2" ipv4 primary address 10.0.1.11
        /configure router "Base" interface "ptp2" ipv4 primary prefix-length 24
        /configure router "Base" interface "l2" admin-state enable
        /configure router "Base" interface "l2" port 2/1/c1/2
        /configure router "Base" interface "l2" ipv4 primary address 10.111.0.100
        /configure router "Base" interface "l2" ipv4 primary prefix-length 24
        /configure router "Base" isis 0 admin-state enable
        /configure router "Base" isis 0 area-address [49.0001]
        /configure router "Base" isis 0 advertise-passive-only true
        /configure router "Base" isis 0 interface "ptp1" { }
        /configure router "Base" isis 0 interface "ptp2" { }
        /configure router "Base" isis 0 interface "l2" passive true
        /configure router "Base" isis 0 interface "system" passive true
      mgmt-ipv4: 10.78.140.3
      mgmt-ipv6: fd00:10:78:140::3
      network-mode: container:SRSIM11-1
      env:
        NOKIA_SROS_SYSTEM_BASE_MAC: 02:f3:11:00:00:00
    srsim11-b:
      kind: nokia_srsim
      type: sr-7
      network-mode: container:SRSIM11-1
      env:
        NOKIA_SROS_SYSTEM_BASE_MAC: 02:f3:11:00:00:00
        NOKIA_SROS_SLOT: B
    SRSIM11-1:
      kind: nokia_srsim
      type: sr-7
      env:
        NOKIA_SROS_SLOT: 1
    srsim11-2:
      kind: nokia_srsim
      type: sr-7
      network-mode: container:SRSIM11-1
      env:
        NOKIA_SROS_SLOT: 2
        NOKIA_SROS_CARD: iom5-e
        NOKIA_SROS_MDA_1: me6-100gb-qsfp28
        NOKIA_SROS_MDA_2: me16-25gb-sfp28+2-100gb-qsfp28
        NOKIA_SROS_SFM: m-sfm6-7/12
        ```

**Comment by kaelemc** [comment]
I rebased this one on my other PR in the meantime as the other PR implements some function that this relies on... I can make a fresh one later.

**Comment by kaelemc** [comment]
So, the original PR comment is not so relevant. The changes I have made:

Components are sorted, but in the order that the LC slots are coming first, and then the CPM slots last.. so the order would be something like 1, 2, 3, 4... A, B.

This means that the other components attach to LC containers, and then CPM container if no LC is defined, allowing for failover tests.

The management IP getter has also been updated, for a distributed node it will now return the IP from the container which owns the netns (since we sort the containers, it should be 0th index of the components array). Thus meaning any time a component node is trying to get the overall nodes mgmt IP, it should get something back as long as there is actually a v4/v6 address assigned to the netns 'owner' container.

I have also made it so the base component node does not have its name changed with the CPM suffix, meaning the hostname and SSh config will register for node name *without* the slot suffix.. Since a node defined with components is a single netns with only one set of mgmt IP, I think this is a logical change to make. 

Have also updated the tests & cert generation to account for this.

I will open this PR for review once the config generation one is merged as this one is based on that PR.

**Review by sacckth — COMMENTED** [review]
Ok on my end... Needs just docs update and review tests

**Comment by kaelemc** [comment]
@hellt @sacckth updated docs, and tests. Once #2827 is merged. I will update this branch and resolve any conflicts. Otherwise it is good from me I think.

File: docs/manual/kinds/sros.md [hunk 1]
```diff
@@ -382,14 +382,14 @@ topology:
 
 When a distributed SR-SIM node is defined using `components`, we need to take into account the following:
 
-1. Individual containers will be attached to the namespace of the 1st element of the `components` list: CPM-A in the above examples.
+1. The component order gets sorted[^5] upon deployment of the lab. Individual containers will be attached to the namespace of the 1st element in the sorted `components` list: CPM-1 in the above examples.
```
hellt, line 385: likely we need to explain the sorting order, right?

hellt: also it seems the `-a|b` suffix is no more for the containers? This needs to be explained in the docs maybe?

kaelemc: The linked footnote is not sufficient enough?

https://github.com/srl-labs/containerlab/pull/2834/files#diff-c9558ec08bafe8588ba257a8a5da12d4e097ed46285060fb7dc97c5332ab031fR878

kaelemc: re: suffix, ok yes I will adjust that

kaelemc: @hellt Actually it looks like there is not really any mention of the CPM suffix in the docs to begin with?

hellt: yeah, looks like it =)

hellt:
lgtm
@sacckth maybe give it a try to check that some existing labs are running fine?

@tiago-amado maybe you also want to try this load, because now you don't get `-a` suffix in the node name, which will impact the anysec lab I guess

the binary to download (linux/amd64) - https://github.com/srl-labs/containerlab/actions/runs/18219861048/artifacts/4174087800

tiago-amado:
> lgtm @sacckth maybe give it a try to check that some existing labs are running fine?
> 
> @tiago-amado maybe you also want to try this load, because now you don't get `-a` suffix in the node name, which will impact the anysec lab I guess
> 
> the binary to download (linux/amd64) - https://github.com/srl-labs/containerlab/actions/runs/18219861048/artifacts/4174087800

Yes, this will impact the Automation Panel (backend main.go and frontend svelte), that are referencing `PE1-a` and `PE2-a`.
I'll then raise a new PR to update this.

sacckth:
unleash the kraken! checked on my end. it seems ok. Side note: 

```
###### CLAB-sros-START ######
172.20.20.2	clab-sros-sr-14s-a	# Kind: nokia_srsim
172.20.20.2	clab-sros-sr-14s	# Kind: nokia_srsim
3fff:172:20:20::2	clab-sros-sr-14s-a	# Kind: nokia_srsim
3fff:172:20:20::2	clab-sros-sr-14s	# Kind: nokia_srsim
```
I think is good to have aliases for the `node` name and the `node-a` In theory the other labs should work still, except of course for TLS stuff, unless we add another SAN for `node` & `node-a`

docs/manual/kinds/sros.md
```diff
@@ -382,14 +382,14 @@ topology:
...
 2. When changing a MDA or card type from its default value, the configuration for card, SFM and MDA must be also defined.
 3. Links can be added referring to the node name. The same [interface naming](#interface-naming) convention holds for all SR-SIM nodes.
 4. Environment variable based configuration on per-component, or node-level will override the configuration set in `type`, `xiom`, `sfm` and `mda` fields.
 
 ##### Configuration for components
 
-When using the `components` structure in the node definition for a distributed node, containerlab will also generate the SR OS configuration for the installed components in the chassis, as well as relevant power supply configuration[^5] to ensure the installed components come up without requiring a user to manually provide the configuration.
+When using the `components` structure in the node definition for a distributed node, containerlab will also generate the SR OS configuration for the installed components in the chassis, as well as relevant power supply configuration[^6] to ensure the installed components come up without requiring a user to manually provide the configuration.
```
sacckth, line 392: maybe mention this can be disabled as well?

kaelemc: It's right below those lines as an admonition :D

kaelemc:
> unleash the kraken! checked on my end. it seems ok. Side note:
> 
> ```
> ###### CLAB-sros-START ######
> 172.20.20.2	clab-sros-sr-14s-a	# Kind: nokia_srsim
> 172.20.20.2	clab-sros-sr-14s	# Kind: nokia_srsim
> 3fff:172:20:20::2	clab-sros-sr-14s-a	# Kind: nokia_srsim
> 3fff:172:20:20::2	clab-sros-sr-14s	# Kind: nokia_srsim
> ```
> 
> I think is good to have aliases for the `node` name and the `node-a` In theory the other labs should work still, except of course for TLS stuff, unless we add another SAN for `node` & `node-a`

SANs are already there for `node-a` by default, In this PR I already manually add the SAN for the short & long name of the base node (node without any slot suffix). So this is already the behaviour. 

I think you don't need to change your lab @tiago-amado

kaelemc:
Thanks all :)

tiago-amado:
> > I think is good to have aliases for the `node` name and the `node-a` In theory the other labs should work still, except of course for TLS stuff, unless we add another SAN for `node` & `node-a`
> 
> SANs are already there for `node-a` by default, In this PR I already manually add the SAN for the short & long name of the base node (node without any slot suffix). So this is already the behaviour.
> 
> I think you don't need to change your lab @tiago-amado

@kaelemc , I've tested with v0.71.0 and I'm getting the warnings below, but seems these are cosmetic and don't have impact.
Think these should not be shown as they appear as warn in yellow.

```
18:17:57 WARN SR-SIM node has no type set for component in slot, skipping component SR OS config generation. node=pe2-a slot=1 

18:17:58 WARN SR-SIM node has no type set for component in slot, skipping component SR OS config generation. node=pe1-a slot=1 

18:17:59 WARN SR-SIM node has no type set for component in slot, skipping component SR OS config generation. node=ce5-a slot=1 

18:17:59 WARN SR-SIM node has no type set for component in slot, skipping component SR OS config generation. node=ce6-a slot=1
```
image

kaelemc:
Thanks for the feedback, yes I agree, we should change to info.

PR #2855, opened 2025-09-27
chore: refactor gotty/sshx tools

carlmontanari:


cmd/graph.go
```diff
@@ -122,7 +122,7 @@ func graphFn(o *Options) error {
 	}
 
 	if o.Graph.GenerateDotFile {
-		return c.GenerateDotGraph()
+		return c.GenerateDotGraph(ctx)
```
carlmontanari, line 125: updated golancli locally and it had some more complaints about not using contexts in places, so fixed those while at this

cmd/tools_gotty.go
```diff
@@ -39,11 +38,6 @@ type GoTTYListItem struct {
 	Owner       string `json:"owner"`
 }
 
-// GoTTYNode implements runtime.Node interface for GoTTY containers.
-type GoTTYNode struct {
```
carlmontanari, line 43: removed this and added a endpointless node in runtime rather than having this thin shim for gotty/sshx

cmd/tools_gotty.go
```diff
@@ -250,50 +244,33 @@ func NewGoTTYNode(
...
-
 	// Create port bindings for the GoTTY web interface
 	portStr := fmt.Sprintf("%d/tcp", port)
-	portBindings := nat.PortMap{
-		nat.Port(portStr): []nat.PortBinding{
-			{
-				HostIP:   "0.0.0.0",
-				HostPort: fmt.Sprintf("%d", port),
+
+	return clabruntime.NewEndpointlessNode(
```
carlmontanari, line 250: heres that endpointless node, nothing special

cmd/tools_gotty.go
```diff
@@ -347,53 +321,25 @@ func gottyAttach(cobraCmd *cobra.Command, o *Options) error { //nolint: funlen
...
-	clabInstance, err := clabcore.NewClabFromTopologyFileOrLabName(
-		o.Global.TopologyFile,
-		o.Global.TopologyName,
-		o.Global.VarsFile,
-		o.Global.Runtime,
-		o.Global.DebugCount > 0,
-		o.Global.Timeout,
-		o.Global.GracefulShutdown,
-	)
+	c, err := clabcore.NewContainerLab(o.ToClabOptions()...)
```
carlmontanari, line 324: probably the most important part of this patch -- removed the NewClabFromTopologyFileOrLabName and now just use the normal o.ToClabOptions and NewContainerlab bits

cmd/tools_gotty.go
```diff
@@ -347,53 +321,25 @@ func gottyAttach(cobraCmd *cobra.Command, o *Options) error { //nolint: funlen
...
 	if err != nil {
 		return err
 	}
 
-	labName := clabInstance.Config.Name
-	networkName := clabInstance.Config.Mgmt.Network
-
-	if networkName == "" {
-		networkName = "clab-" + labName
+	rt, ok := c.Runtimes[o.Global.Runtime]
```
carlmontanari, line 329: removed all the runtime initializer stuff in here because we already do that during NewContainerlab basically

cmd/tools_gotty.go
```diff
@@ -347,53 +321,25 @@ func gottyAttach(cobraCmd *cobra.Command, o *Options) error { //nolint: funlen
...
-	if err != nil {
-		return fmt.Errorf("failed to get runtime initializer for '%s': %w", o.Global.Runtime, err)
-	}
-
-	rt := rinit()
-
-	err = rt.Init(
-		clabruntime.WithConfig(&clabruntime.RuntimeConfig{Timeout: o.Global.Timeout}),
-		clabruntime.WithMgmtNet(&clabtypes.MgmtNet{Network: networkName}),
+	containers, err := c.ListContainers(
```
carlmontanari, line 339: replaced all the listing using the runtime directly w/ the c.ListContainers

cmd/tools_gotty.go
```diff
@@ -347,53 +321,25 @@ func gottyAttach(cobraCmd *cobra.Command, o *Options) error { //nolint: funlen
...
+		ctx,
+		clabcore.WithListContainerName(o.ToolsGoTTY.ContainerName),
```
carlmontanari, line 341: add WithListContainerName which I guess only matters for tools but still seems nicer

core/clab.go
```diff
@@ -105,50 +104,6 @@ func NewContainerLab(opts ...ClabOption) (*CLab, error) {
 	return c, err
 }
 
-// NewClabFromTopologyFileOrLabName creates a containerlab instance using either a topology file
-// path or a lab name. It returns the initialized CLab structure with the topology loaded.
-func NewClabFromTopologyFileOrLabName(
```
carlmontanari, line 110: removing this is prolly the highlight -- was confusing me why it was needed for a while :)

core/options_list.go
```diff
@@ -150,3 +173,10 @@ func WithListFromCliArgs(ss []string) ListOption {
 		o.cliArgs = ss
 	}
 }
+
+// WithListFilter adds user created normal filters to the filter set.
+func WithListFilter(fs []*clabtypes.GenericFilter) ListOption {
```
carlmontanari, line 178: add this list option too -- used in sshx/gotty

runtime/runtime.go
```diff
@@ -120,49 +120,45 @@ func WaitForContainerRunning(
...
 
-	resultErr := fmt.Errorf(
-		"node %q waited %s for external dependency container %q to come up, which did not happen. Giving up now",
-		nodeName,
-		time.Since(startTime),
-		contName,
-	)
+	startTime := time.Now()
 
-TIMEOUT_LOOP:
```
carlmontanari, line 142: just happened to see this and was like wtf loop label! cleaned up thing is functionally the same but no label, yay!

tests/01-smoke/02-destroy-all.robot
```diff
@@ -51,8 +51,8 @@ Inspect ${lab2-name} lab using its name
 
     ${num_lines} =    Run    bash -c "echo '${output}' | wc -l"
     # lab2 only has 1 nodes and therefore inspect output should contain only 1 node with two lines
-    # (+4 lines for the table header and footer)
-    Should Be Equal As Integers    ${num_lines}    6
+    # (+4 lines for the table header and footer, +1 for the log line about parsing topo)
```
carlmontanari, line 54: because we now have an extra log line in the output due to cmd options -> ToClabOptions added the extra parse topology bits

PR #2870, opened 2025-10-05
Add generic interface struct in NodeCfg & IP link var support for XRd

kaelemc:
I added a generic template interface struct to make integration for new NOSes easier for the IPv4/v6 link vars.

XRd is now also the first NOS using this, and by extension now supports the IPv4/v6 link vars, as well as setting the MTU to match the link.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -182,6 +182,8 @@ type NodeConfig struct {
 	TLSCert              string `json:"tls-cert,omitempty"`
 	TLSKey               string `json:"-"` // Do not marshal into JSON - highly sensitive data
 	TLSAnchor            string `json:"tls-anchor,omitempty"`
+	// TemplateInterfaces contains interface data for use in configuration templates
+	TemplateInterfaces []TemplateInterface `json:"template-interfaces,omitempty"`
 	// TLS Certificate configuration
 	Certificate *CertificateConfig
 	// Healthcheck configuration parameters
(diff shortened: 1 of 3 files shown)


types/types.go
```diff
@@ -182,6 +182,8 @@ type NodeConfig struct {
 	TLSCert              string `json:"tls-cert,omitempty"`
 	TLSKey               string `json:"-"` // Do not marshal into JSON - highly sensitive data
 	TLSAnchor            string `json:"tls-anchor,omitempty"`
+	// TemplateInterfaces contains interface data for use in configuration templates
+	TemplateInterfaces []TemplateInterface `json:"template-interfaces,omitempty"`
```
hellt, line 186: I don't think we need or want to extend the NodeConfig struct.

The idea we may follow for config templation is as follows:


1. a user can provide vars for the links and nodes
2. the nodes already have the `Config` property that embeds the Vars - https://github.com/srl-labs/containerlab/blob/e6579300487afe85c294d38d6b0a09fb3eafd60b/types/types.go#L139
3. we can remove the Config block and promote the Vars to be the top level struct in the NodeConfig as we don't have the config dispatcher concept anymore
4. The vars should contain free from vars that users can fill in. As they are propagated to topology json, and ansible/nornir inventories they were meant for users to write their own config templates to provision the lab later on
5. these node and link vars can be used by clab itself in the config generation.

The free form nature of vars struct allows users to put whatever in it. Thats why I was not super hyped when we removed the `map[string]any` from the link vars when we added the ipv4/6 vars under the links.

So I want to make sure we keep the vars structure flexible, so that people can put stuff in it and it in generation.

Hope this makes sense :D

kaelemc: Yes, makes sense

Although the templae config uses the nodeconfig to render right, so we need to store the interfaces of the node somewhere right?

Am I missing something here?

hellt:
This one is not needed now I think?

kaelemc:
Correct I'll add for XRd in another PR but will leave the generic stuff for later

PR #2876, opened 2025-10-06
Adds support for Cisco SD-WAN controllers and C8000v controller mode.

torbbang:
Changes:
  - vr_cisco_sdwan node kind: Supports manager, controller, and validator components with zCloud XML and cloud-init configs
  - C8000v modes: Added autonomous/controller mode support via type parameter

Related to the following vrnetlab PR: https://github.com/srl-labs/vrnetlab/pull/396

PR #2877, opened 2025-10-07
Support for Cisco SD-WAN controllers and C8000v controller mode.

hellt:
move of https://github.com/srl-labs/containerlab/pull/2876

### DIFF
--- a/docs/manual/kinds/cisco_sdwan.md
+++ b/docs/manual/kinds/cisco_sdwan.md
@@ -6,7 +6,7 @@ kind_display_name: Cisco SD-WAN Controllers
 ---
 # Cisco SD-WAN
 
-Cisco SD-WAN controller components are identified with `-{{ kind_code_name }}-` kind in the [topology file](../topo-def-file.md).
+Cisco SD-WAN controller components are identified with `-{{ kind_code_name }}-` kind in the [topology file](../topo-def-file.md). It is built using [srl-labs/vrnetlab](https://github.com/srl-labs/vrnetlab/tree/master/cisco/sdwan-components) project and essentially is a Qemu VM packaged in a container format.
 
 This kind supports all three Cisco SD-WAN controller components:
 
--- a/docs/manual/kinds/cisco_sdwan.md
+++ b/docs/manual/kinds/cisco_sdwan.md
@@ -1,12 +1,12 @@
 ---
 search:
   boost: 4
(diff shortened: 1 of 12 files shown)


docs/manual/kinds/cisco_sdwan.md
```diff
@@ -0,0 +1,256 @@
...
+
+    sdwan-validator:
+      kind: cisco_sdwan
+      type: validator
+      image: vrnetlab/cisco_sdwan-validator:20.16.1
+
+    edge1:
+      kind: cisco_c8000v
+      type: controller  # SD-WAN managed mode
+      image: vrnetlab/vr-c8000v:17.11.01a
```
hellt, line 227: @torbbang I am curious, why the mode is called "controller", whereas this node is not controller but a managed device?

Maybe we should call it "sdwan-managed" if there is no official name for this mode of operation in the cisco docs?

nodes/cisco_sdwan/cisco_sdwan.go
```diff
@@ -0,0 +1,232 @@
...
+
+	"github.com/charmbracelet/log"
+	clabconstants "github.com/srl-labs/containerlab/constants"
+	clabnodes "github.com/srl-labs/containerlab/nodes"
+	clabtypes "github.com/srl-labs/containerlab/types"
+	clabutils "github.com/srl-labs/containerlab/utils"
+)
+
+var (
+	kindNames          = []string{"cisco_sdwan"}
```
hellt, line 24: fyi, renamed the kind to `cisco_sdwan` without the vr prefix

docs/manual/kinds/cisco_sdwan.md
torbbang: I agree that it is not very intuitive, I chose to use autonomous and controller as these are the names Cisco uses for the two boot modes. It is fine by me if we go with regular and sdwan instead as this probably is clearer to most users.

nodes/cisco_sdwan/cisco_sdwan.go
torbbang: Good. I must have got a bit fixated on the "vr" prefix used elsewhere in the codebase

docs/manual/kinds/cisco_sdwan.md
hellt: if this comes from the docs - all good by me.

PR #2878, opened 2025-10-07
SR-SIM startup configuration handling tweaks

kaelemc:
If a user provides a full startup config, I think we should skip any NETCONF operations as the config may have it disabled.

Also in the PR is a slight change to the component config gen.. if the partial config is being provided by the user, I think it's safe to generate the component config considering we already generate a fair chunk of stuff for the user already.

Let me know your thoughts.

### DIFF
--- a/nodes/sros/sros.go
+++ b/nodes/sros/sros.go
@@ -444,7 +444,7 @@ func (n *sros) PostDeploy(ctx context.Context, params *clabnodes.PostDeployParam
 				)
 			}
 			// don't save config if full config is sent.. it might not have NETCONF
-			if !isFullConfgFile(n.Cfg.StartupConfig) {
+			if !isFullConfigFile(n.Cfg.StartupConfig) {
 				err = n.saveConfigWithAddr(ctx, addr)
 				if err != nil {
 					return fmt.Errorf("save config to node %q, failed: %w", n.Cfg.LongName, err)
@@ -1065,7 +1065,7 @@ func (n *sros) addDefaultConfig() error {
 	componentConfig := ""
 
 	// Generate component configuration IF no startup-config, or partial is defined.
-	if !isFullConfgFile(n.Cfg.StartupConfig) {
+	if !isFullConfigFile(n.Cfg.StartupConfig) {
 		componentConfig = n.generateComponentConfig()
 	} else {
 		log.Debugf("SR-SIM node %q has non-partial startup-config defined, skipping component config gen", n.Cfg.LongName)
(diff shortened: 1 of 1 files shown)


nodes/sros/sros.go
```diff
@@ -1503,10 +1507,23 @@ func isPartialConfigFile(c string) bool {
 	return strings.Contains(strings.ToUpper(c), ".PARTIAL")
 }
 
+// isNonPartialConfigFile returns true if the config file doesn't contain .partial substring
+// and the config file is NOT nil (ie. startup config IS defined).
+// it is intended that the 'c' arg is n.Cfg.StartupConfig.
+func isNonPartialConfigFile(c string) bool {
+	return c != "" && !isPartialConfigFile(c)
+}
+
```
hellt, line 1516: `isNon` is not a good func name, as it leads to double negation - `!isNon`

Instead do `isPartialConfigFile` and the `!isPartialConfigFile` will give you the negative case

kaelemc: I changed it to `isFullConfigFille`

The reason I want this specific check is to make sure that there is some config file defined, and it is NOT partial.. ie both cases need to be true.

whereas the current `isPartialConfigFile` doesn't care if startupConfig is defined or not.

makes sense? kinda?

nodes/sros/sros.go
```diff
@@ -1503,10 +1507,29 @@ func isPartialConfigFile(c string) bool {
...
+	if isFullConfigFile(n.Cfg.StartupConfig) {
+		log.Debug(
+			"node has full startup config, skipping NETCONF check",
+			"kind",
+			n.Cfg.Kind,
+			"node",
+			n.Cfg.ShortName,
+		)
+		return true, nil
+	}
```
hellt, line 1532: @kaelemc hmm, can't we maybe look into the provided startup config, find the netconf config piece there and do everything that requires netconf instead of bailing out?

Maybe some dumb check for a substring that has `netconf` + some predicate to identify this is the server that is configured

kaelemc: Yes, but why exactly?. The only NETCONF operation is just a save config. If a user gives a full startup cfg, what is there to save?

kaelemc: I guess the cert provisioning is also there.. but debatable if we wanna do that when a full cfg instead of partial is provided.

hellt: for `func (n *sros) IsHealthy`?

kaelemc: Sergio may be better to comment but my understanding was this is to just confirm the port is open (because the server doesnt start instantly) and then start doing the NETCONF shenanigans.. There is another healthcheck before this which confirms that the node has booted successfully.

hellt: I see, then I am fine with going the way it is defined right now.

lets see if @sacckth has comments

sacckth: yeah, I just check for port 830 open... One caveat we had with providing full config file is that all netconf operations will anyway fail if the password doesn't match the `defaultCredentials` var. I found it impossible to render initial config without having to change the admin pwd, so there's that. So TDLR it makes sense to skip the entire netconf piece if we do this... Maybe throw a warning log if we provide full config?

Also, docs...

kaelemc: The current NETCONF config save/healthcheck behaviour is not even mentioned in the docs,  unsure on what (and where) I should document.

PR #2880, opened 2025-10-09
Move endpoint IP addresses out of vars + restore vars functionality

kaelemc:
We added IP link vars, however in the process I broke the existing vars functionality. We should move the IPv4/v6 addresses of an endpoint *outside* of vars to the top level (like MTU).. and also restore the old functionality of vars (whereby it's an aribitrary set of values provdied by user).

### DIFF
--- a/core/export_templates/auto.tmpl
+++ b/core/export_templates/auto.tmpl
@@ -45,6 +45,18 @@
         "node": "{{ $ep.GetNode.GetShortName }}",
         "interface": "{{ $ep.GetIfaceName }}",
         "mac": "{{ $ep.GetMac }}",
+        {{- $ipv4 := $ep.GetIPv4Addr -}}
+        {{- if $ipv4.IsValid }}
+        "ipv4": "{{ $ipv4.String }}",
+        {{- end -}}
+        {{- $ipv6 := $ep.GetIPv6Addr -}}
+        {{- if $ipv6.IsValid }}
+        "ipv6": "{{ $ipv6.String }}",
+        {{- end }}
+        {{- $vars := $ep.Vars -}}
+        {{- if and $vars (gt (len $vars) 0) }}
+        "vars": {{ ToJSONPretty $vars "        " "  " }},
+        {{- end }}
         "peer": "{{if eq (len $eps) 2}}z{{else}}dummy{{end}}"
       }
       {{- if eq (len $eps) 2 }}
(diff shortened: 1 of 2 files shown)


docs/manual/topo-def-file.md
```diff
@@ -387,18 +391,20 @@ Such interfaces are useful for testing and debugging purposes where we want to m
 
 ##### Variables
 
-Link variables are a way to supply additional link-related information that can be passed to the configuration templates.
+Link variables are a way to supply additional link-related information that can be passed to the configuration templates and will be rendered in the topology json file.
```
hellt, line 394: @kaelemc can you please check if `vars` end up being in the topology-data.json?

kaelemc: cheers for catching that.. how is it supposed to work for link vars with brief format? are they are inherited by both endpoints?

hellt: good question. The short notation has only one way to specify the vars:

```
  links:
    - endpoints: ["srl:ethernet-1/1", "ceos:Ethernet1/1"] 
      vars:
        var1: val1
```

whereas the extended link notation is more flexible, as it allows to put the vars in the link object and under the endpoints:


```
links:
  - type: veth
    endpoints:
      - node: <NodeA-Name>                  
        interface: <NodeA-Interface-Name>   
        mac: <NodeA-Interface-Mac>          
        vars:
          var2: value2
      - node: <NodeB-Name>                  
        interface: <NodeB-Interface-Name>   
        mac: <NodeB-Interface-Mac>          
        vars:
          var3: value3
    mtu: <link-mtu>                         
    vars:
      var1: value1
    labels: <link-labels>                   
```

Now, the default export template needs to be adjusted, because right now it is modelled like this:

```json
  "links": [
    {
      "a": {
        "node": "srl1",
        "interface": "e1-1",
        "mac": "aa:c1:ab:75:9f:e6",
        "peer": "z"
      },
      "z": {
        "node": "srl2",
        "interface": "e1-1",
        "mac": "aa:c1:ab:93:43:2f",
        "peer": "a"
      }
    }
  ]
}
```

and I think it needs to be extended to represent the extended link format, e.g.:

```json
{
  "links": [
    {
      "endpoints": [
        {
          "a": {
            "node": "srl1",
            "interface": "e1-1",
            "mac": "aa:c1:ab:75:9f:e6",
            "peer": "z",
            "vars": {
              "var2": "value2"
            }
          },
          "z": {
            "node": "srl2",
            "interface": "e1-1",
            "mac": "aa:c1:ab:93:43:2f",
            "peer": "a",
            "vars": {
              "var3": "value3"
            }
          }
        }
      ],
      "vars": {
        "var1": "value1"
      }
    }
  ]
}
```

I think this can be a different PR, though. Let me know if that makes things clear?

kaelemc: yep, crystal clear chief  👌.. and yea should be another PR IMO

hellt: ok, this one lgtm with the recent commit adding IP to the links.

let's tackle the link vars in a subsequent PR

PR #2896, opened 2025-10-17
[docs] cEOS: Add documentation about the forwarding agent and interface mode

robinchrist:
This PR adds some documentation about cEOS forwarding agent and interface mode.

### DIFF
--- a/docs/manual/kinds/ceos.md
+++ b/docs/manual/kinds/ceos.md
@@ -5,18 +5,18 @@ kind_code_name: arista_ceos
 kind_display_name: Arista cEOS
 ---
 # -{{ kind_display_name }}-
+
 -{{ kind_display_name }}- is identified with `-{{ kind_code_name }}-` kind in the [topology file](../topo-def-file.md).
 The `-{{ kind_code_name }}-` kind defines a supported feature set and a startup procedure of a `ceos` node.
 
-
-cEOS nodes launched with containerlab comes up with
+cEOS nodes launched with containerlab come up with
 
 * their management interface `eth0` configured with IPv4/6 addresses as assigned by docker
 * hostname assigned to the node name
 * gNMI, Netconf and eAPI services enabled
 * `admin` user created with password `admin`
 
-## Getting cEOS image
+## Getting -{{ kind_display_name }}- image
 <!-- --8<-- [start:ceos-get-image] -->
 Arista requires its users to register with arista.com before downloading any images. Once you created an account and logged in, go to the [software downloads](https://www.arista.com/en/support/software-download) section and download ceos64 tar archive for a given release.
 
@@ -31,32 +31,45 @@ docker import cEOS64-lab-4.32.0F.tar.xz ceos:4.32.0F
 
 Arista cEOS node launched with containerlab can be managed via the following interfaces:
 
-=== "bash"
-    to connect to a `bash` shell of a running ceos container:
-    ```bash
-    docker exec -it <container-name/id> bash
-    ```
-=== "CLI"
-    to connect to the ceos CLI
-    ```bash
-    docker exec -it <container-name/id> Cli
-    ```
-=== "NETCONF"
-    NETCONF server is running over port 830
-    ```bash
-    ssh root@<container-name> -p 830 -s netconf
-    ```
-=== "gNMI"
-    gNMI server is running over port 6030 in non-secure mode
-    using the best in class [gnmic](https://gnmic.openconfig.net/) gNMI client as an example:
-    ```bash
-    gnmic -a <container-name/node-mgmt-address>:6030 --insecure \
-    -u admin -p admin \
-    capabilities
-    ```
+/// tab | bash
+to connect to a `bash` shell of a running ceos container:
 
-!!!info
-    Default user credentials: `admin:admin`
+```bash
+docker exec -it <container-name/id> bash
+```
+
+///
+/// tab | CLI
+to connect to the ceos CLI
+
+```bash
+docker exec -it <container-name/id> Cli
+```
+
+///
+/// tab | NETCONF
+NETCONF server is running over port 830
+
+```bash
+ssh root@<container-name> -p 830 -s netconf
+```
+
+///
+/// tab | gNMI
+gNMI server is running over port 6030 in non-secure mode
+using the best in class [gnmic](https://gnmic.openconfig.net/) gNMI client as an example:
+
+```bash
+gnmic -a <container-name/node-mgmt-address>:6030 --insecure \
+-u admin -p admin \
+capabilities
+```
+
+///
+
+### Credentials
+
+Default user credentials: `admin:admin`
 
 ## Interfaces mapping
 
@@ -105,8 +118,9 @@ When containerlab launches ceos node, it will set IPv4/6 addresses as assigned b
 
 ### User-defined interface mapping
 
-!!!note
-    Supported in cEOS >= 4.28.0F
+/// note
+Supported in cEOS >= 4.28.0F
+///
 
 It is possible to make ceos nodes boot up with a user-defined interface layout. With the [`binds`](../nodes.md#binds) property, a user sets the path to the interface mapping file that will be mounted to a container and used during bootup. The underlying linux `eth` interfaces (used in the containerlab topology file) are mapped to cEOS interfaces in this file. The following shows an example of how this mapping file is structured:
 
@@ -210,8 +224,9 @@ topology:
 
 This topology will be equivalent to `ceos1:Ethernet1/1` connected to `ceos2:Ethernet2/1/1`.
 
-!!!note
-    This feature can not be used together with interface mapping. If the interface mapping is in use, all names must be redefined in the map and the underscore naming option will not work. Also, it's only possible to rename Ethernet interfaces this way, not management ports.
+/// note
+This feature can not be used together with interface mapping. If the interface mapping is in use, all names must be redefined in the map and the underscore naming option will not work. Also, it's only possible to rename Ethernet interfaces this way, not management ports.
+///
 
 ## Features and options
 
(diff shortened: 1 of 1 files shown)


hellt:
thanks!

Hi @ankudinov, would someone from eos team vet the doc changes? Ty!

ankudinov:
@hellt ack I'll review it early next week

docs/manual/kinds/ceos.md
```diff
@@ -326,6 +326,68 @@ To start an Arista cEOS node containerlab uses the following configuration:
...
+
+Starting with `EOS-4.30.1F`, cEOS-lab and vEOS-lab use `Arfa` by default if not configured otherwise.
+
+You can manually switch between the two forwarding agents using
+```
+platform tfa
+
+personality python
+personality arfa
+```
```
ankudinov: I'd suggest to mention that python personality exists, but avoid adding configuration commands, etc. arfa is the default and I'd expect it to cover most of lab scenarios by now. If it's not the case - please bring specific examples where ARFA doesn't work and I can follow-up.

Add ident to config snippet:

```text
platform tfa
   personality arfa
``` 

We can simply mention what is that and why users see these configuration lines in recent releases.

docs/manual/kinds/ceos.md
```diff
@@ -326,6 +326,68 @@ To start an Arista cEOS node containerlab uses the following configuration:
...
+or via the magic file `/mnt/flash/ceos-config`.
+
+For `python`, use
+```title="/mnt/flash/ceos-config"
+TFA_VERSION=1
+``` 
+For `Arfa`, use
+```title="/mnt/flash/ceos-config"
+TFA_VERSION=2
+```
```
ankudinov: I'd avoid adding that at all as this can be confusing. Setting the required lines in the intial configuration is the preferred way.
If we want to add this to cLab docs - I'd look at that under a specific use case angle. It's hard for me to imagine, why a user would set TFA_VERSION. However there are [other valid cases](https://github.com/aristanetworks/aclabs/blob/d9d91fff8dd2734027bd84d1cecb4ee890322d30/labs/cvaas-cvaas-and-avd-demo--evpn-mlag/clab/topology.clab.yml#L27) for `/mnt/flash/ceos-config` in general. For example, setting [serial number and system mac address](https://github.com/aristanetworks/aclabs/blob/d9d91fff8dd2734027bd84d1cecb4ee890322d30/labs/cvaas-cvaas-and-avd-demo--evpn-mlag/clab/sn/l01.txt#L1) for certain CVP/CVaaS lab cases.
It's fine to mention TFA_VERSION as well while talking about `/mnt/flash/ceos-config`, but let's make it a story.

docs/manual/kinds/ceos.md
```diff
@@ -326,6 +326,68 @@ To start an Arista cEOS node containerlab uses the following configuration:
...
+
+
+The environment variable `ETBA` is not properly documented by Arista, but apparently it is necessary for correct operation.
```
ankudinov: Remove this. Yes, ETBA env variable is required for `docker run` or similar cases. Containerlab is setting all required variables and it works great. Also it's not clear what is the conclusion from this sentence.

docs/manual/kinds/ceos.md
```diff
@@ -326,6 +326,68 @@ To start an Arista cEOS node containerlab uses the following configuration:
...
+- Multicast
+- FlexEncap
+- Q-in-Q subinterface encapsulation
+- Pseudowire dual tagging
+
+The following features are not supported in `python` mode:
+
+- PBR
+- VRF Selection Policy
+- Sflow Egress interface
```
ankudinov: Remove. Python is "legacy", let's talk about ARFA only.

docs/manual/kinds/ceos.md
```diff
@@ -326,6 +326,68 @@ To start an Arista cEOS node containerlab uses the following configuration:
...
+
+Interface mode can be configured using
+```
+platform tfa
+
+interface mode relaying
+interface mode default
+```
+
+It is not documented which other advantages or disadvantages relay mode has compared to default mode.
```
ankudinov: What is the purpose of documenting that in cEOS-lab docs? I have a feeling this must be documented in cEOS TOIs, even if not all of this exists in our docs right now. =)
I'm a bit concerned about documenting it here as:
1. There is no explanation of what relaying mode is.
2. Currently there is a single valid use case I'm aware of for `interface mode relaying`, which is [PTP support](https://www.arista.com/en/support/toi/eos-4-35-0f/22485-precision-time-protocol-support-on-ceos-lab).
3. I have to follow-up on other features, but not certain if they are even implemented right now. And a bit confused as this information may have leaked from some NDA document, roadmap or something. No worries, just flagging that I don't think we (Arista) published that anywhere. Correct me if I'm wrong.

docs/manual/kinds/ceos.md
```diff
@@ -326,6 +326,68 @@ To start an Arista cEOS node containerlab uses the following configuration:
...
+
+
+The environment variable `ETBA` is not properly documented by Arista, but apparently it is necessary for correct operation.
+
+The following features are not supported in `Arfa` mode:
+
+- Multicast
+- FlexEncap
+- Q-in-Q subinterface encapsulation
+- Pseudowire dual tagging
```
ankudinov: It's will extremely hard to maintain this section and keep it up to date, especially on cLab docs side. Trust me, even internally it's a bit of challenge. For ex.:
- multicast is HUGE. What multicast features are we talking about?
- I see that some features from your list are already in fixed status or being worked on. However this is constantly changing and there will be many other features that customers are asking for.
Personally I believe that the best way to know if feature is supported - test the latest release and ask your SE in case of concerns.

Not to mention all possible kernel limitations that have nothing to do with cEOS but can break things for a specific lab. And talking a lot about kernel can scare potential cLab and cEOS-lab users. =)

docs/manual/kinds/ceos.md
robinchrist: I added that because a document mentioned that `Arfa` does not support Multicast (just "Multicast", no explanation). 

But sure, we can remove the `personality python` part

docs/manual/kinds/ceos.md
robinchrist: This goes hand-in-hand with your first comment regarding `personality python`

But sure, let's expand the section to serial number and mac address

docs/manual/kinds/ceos.md
robinchrist: Die `ETBA` Env Variable was the **main reason** why I even made this PR.

Entirely undocumented $magic env variables in an open source project are a big no, imo.

docs/manual/kinds/ceos.md
robinchrist: > There is no explanation of what relaying mode is

Yeah, I'd like to know that too. I've poked around inside cEOS, but haven't come to a conclusion yet ;)

> Currently there is a single valid use case I'm aware of for interface mode relaying, which is [PTP support]

What about the 4 other features listed?

> roadmap or something. No worries, just flagging that I don't think we (Arista) published that anywhere. Correct me if I'm wrong.

Just poking around inside cEOS while trying to understand where the magic `ETBA` variables comes from...

docs/manual/kinds/ceos.md
robinchrist: > multicast is HUGE. What multicast features are we talking about?

I'd like to know that too ;)

Multicast was the main reason why I even added this point. The other ones are pretty niche, but Multicast is a pretty important topic.

There is also zero documentation about

```
router multicast
    ipv4
        software-forwarding kernel
```

vs
```
router multicast
    ipv4
        software-forwarding tfa
```

docs/manual/kinds/ceos.md
ankudinov: They are kind of documented on cLab side: https://github.com/srl-labs/containerlab/blob/7f4b5d9a1226049bf68f8e2bc208109dda25f7b3/nodes/ceos/ceos.go#L44

So, why exactly do you want to expose these env vars to the cLab enduser?
One big benefit I see when using cLab is that there is no need to think about these env vars and set them manually. Otherwise I'd do everything in Linux shell.

In general I'd expect everyone using these env to be comfortable reading cLab source code. But a valid case can change my opinion.

@hellt what do you think?

robinchrist: That's not documentation imo, that's just setting the env variable.

They are shown directly above, and while 
```
CEOS:1
EOS_PLATFORM:ceoslab
container:docker
SKIP_ZEROTOUCH_BARRIER_IN_SYSDBINIT:1
INTFTYPE:eth
MAPETH0:1
MGMT_INTF:eth0
```

Are pretty self-explanatory, `ETBA:1` is not.

What is this environment variable? Why is it necessary? What does it do there? Questions that should be answered in an open source project, even if it's just `This environment variable is related to the forwarding plane and must be set for cEOS to work`

robinchrist: @ankudinov I have moved the `ETBA` env variable documentation into the already existing `Environment Variables` tab under `Container Configuration` and converted it into a code annotation.

Does that work better for you?

docs/manual/kinds/ceos.md
robinchrist: Please check again in the updated version, I've reworked the entire section.

docs/manual/kinds/ceos.md
robinchrist: Please check again in the updated version, I've reworked the entire section.

docs/manual/kinds/ceos.md
robinchrist: Please check again in the updated version, I've reworked the entire section.

docs/manual/kinds/ceos.md
robinchrist: Please check again in the updated version, I've reworked the entire section.

docs/manual/kinds/ceos.md
ankudinov: @robinchrist just to be clear - we are not talking about any open source project here. cLab is an open source project. cEOS-lab is Arista image an is not open source. ETBA is cEOS-lab env variable. This is not changing the fact that we all deserve better docs. But it's not as easy as we all would love to see. ETBA is Ethernet Bridging Test Agent. Arfa is specific implementation of ETBA. By setting ETBA=1 you enable it where TFA_VERSION (or config) allows you to pick between Python and Arfa.

docs/manual/kinds/ceos.md
ankudinov: My point is - it will be impossible to maintain on Containerlab docs side. Let's remove it completely.
In fact I found [the document](https://www.arista.com/en/support/toi/eos-4-30-1f/17517-arfa-a-faster-forwarding-agent-for-ceos-lab-and-veos-lab) that seems to be the inspiration for this PR. Looks like we recently updated it. Let me follow-up internally and see what can be improved in this doc. But it must stay on arista.com as it has nothing to do with cLab as such.

Meanwhile please reach to you SE if you have any specific concerns or ping me on LinkedIn and we can discuss your use case.

docs/manual/kinds/ceos.md
ankudinov: Leave that to cEOS-lab docs please and the TOI mentioned below. No reason to add that to cLab docs.

docs/manual/kinds/ceos.md
robinchrist: It‘s already removed in the newest revision :)

docs/manual/kinds/ceos.md
robinchrist: Also already removed in the newest revision :)

docs/manual/kinds/ceos.md
robinchrist: Thank you for the explanation!

Are you okay with the annotation in the Environment Variables I moved this into in the newest revision?

I‘m not saying that there should be exhaustive documentation about this flag, but some words of explanation (needed for the dataplane) should definitely be there, because the env variable is a magic acronym and no explanation at all is not good.

PR #2902, opened 2025-10-23
Cisco vIOS & vIOS-L2

torbbang:
Adds support for Cisco vIOS and vIOSL2 nodes


core/register.go
```diff
@@ -38,6 +38,8 @@ import (
 	clabnodesvr_cat9kv "github.com/srl-labs/containerlab/nodes/vr_cat9kv"
 	clabnodesvr_csr "github.com/srl-labs/containerlab/nodes/vr_csr"
 	clabnodesvr_freebsd "github.com/srl-labs/containerlab/nodes/vr_freebsd"
+	clabnodesvr_vios "github.com/srl-labs/containerlab/nodes/vr_vios"
+	clabnodesvr_viosl2 "github.com/srl-labs/containerlab/nodes/vr_viosl2"
```
hellt: thansk @torbbang 
would you mind changing the kinds to `cisco_vios` and `cisco_viosl2` accordingly.

This would mean `clabnodescisco_vios` and `clabnodescisco_viosl2` import paths accordingly

torbbang: No problem!

tvarohohlavy:
I see that the default credentials are cisco/cisco instead of admin/admin as for other nodes.
Not sure how the VS Code CLAB plugin is done, but at the moment for me its using admin/admin by default.
Either it needs update as well or using admin/admin here would be better.
What do you think @torbbang ?

tvarohohlavy:
Also it sems all vrnetlab node kinds should probably have the file names starting with "vr" which you have for docs, but not for Go code.

torbbang:
> I see that the default credentials are cisco/cisco instead of admin/admin as for other nodes. Not sure how the VS Code CLAB plugin is done, but at the moment for me its using admin/admin by default. Either it needs update as well or using admin/admin here would be better. What do you think @torbbang ?

Good catch @tvarohohlavy! Will push a fix this evening

kaelemc:
VR prefix is not necessary anymore imo, artifact of the past...

we migrated to this unified format of `<vendor>_<model>` to match the kinds

torbbang:
Credentials have been corrected in the latest commit

kaelemc:
Thanks, will take a look shortly.

kaelemc:
@torbbang Can we keep this as just a single `cisco_vios` kind, if anything in future is needed for L2 specific, lets follow IOL and use the `type:` to differentiate.

nodes/cisco_vios/cisco-vios.go
```diff
@@ -0,0 +1,91 @@
...
+	"path"
+	"regexp"
+
+	clabnodes "github.com/srl-labs/containerlab/nodes"
+	clabtypes "github.com/srl-labs/containerlab/types"
+	clabutils "github.com/srl-labs/containerlab/utils"
+)
+
+var (
+	kindnames          = []string{"cisco_vios", "vr-vios", "vr-cisco_vios"}
```
kaelemc: Just keep `cisco_vios` for this one.

torbbang:
> @torbbang Can we keep this as just a single `cisco_vios` kind, if anything in future is needed for L2 specific, lets follow IOL and use the `type:` to differentiate.

I want to keep the L2 version as it has a more robust feature-set than the equivalent IOS based IOL L2 images.

You are correct that it would be better to use a `Type` field for this. I made them separate kinds mostly to match the existing structure in vrnetlab. I will consolidate vios and viosl2 in vrnetlab and update this PR accordingly. Thank you for the feedback!

tvarohohlavy:
I personally find it inconvenient to have Types instead of different images as you cannot define both in defaults section afaik. I trust clab founders there is a good reason for that distinction, but as its discussed here, wanted to give my 5 cents.

kaelemc:
It's more a matter of consistency and reducing duplicated code. In the current PR you can see that the L2 and non-L2 only have diffs in the naming otherwise they are exactly the same. 

From a maintainability perspective it makes more sense to have some logic for what L2 needs done differently; and for clab it should be a very minimal difference since we only care about bringing the box up and giving mgmt connectivity (see IOL for what we do different for L2 image).

On the kind defaults front, yes that is true you can't differentiate between type BUT [groups]( https://containerlab.dev/manual/topo-def-file/#groups) is best suited for that kind of usecase IMO. What specific params are you looking to have different between L2/non-L2 node on kind-default basis?

tvarohohlavy:
Yep, you are right you can do a lot with groups and it deduplicates code. I am fine with that : ]

torbbang:
The latest commit consolidates the vios node kinds into one and implements the `Type` field.  
I have also created a matching vrnetlab PR to consolidate vios & viosl2: https://github.com/srl-labs/vrnetlab/pull/409

tvarohohlavy:
I have tested it before the vrnetlab changes in mentioned PR and it works for me fine for both types with or without configs.


I see inconsistency between for example iol and vios in how configs are handled as with vios whatever you supply is basically partial, not full config, but thats outside of this PR scope I believe.
I guess if desired we could in another PR move default config generation into clab from vrnetlab and then we would also have actual startup config in the clab lab dconfig directories, not empty or the partial files like now.

This would be altered to only apply config file if present: https://github.com/srl-labs/vrnetlab/blob/master/cisco/vios/docker/launch.py#L127
Logic would be moved into CLAB

kaelemc:
Thanks, I will give it a test.

It is preferred to have the config generation on the clab side so we can do exactly what you mention regarding the partial or full startup configs.

I think it should be fairly straightforward. Just ensure on the clab side that we set `CLAB_MGMT_PASSTHROUGH` env var as `"true"` else the mgmt ip will mismatch. If you want, I can assist with that.. only for vIOS though, doesn't have to be every vrnetlab kind.

torbbang:
That would be greatly appreaciated @kaelemc!

tvarohohlavy:
> ...only for vIOS though, doesn't have to be every vrnetlab kind.

@kaelemc What exactly do you mean by that?

kaelemc:
We just shift the config gen to clab for vIOS only. Not the other NOSes like Cat8k, CSR, SROS etc.

tvarohohlavy:
> We just shift the config gen to clab for vIOS only. Not the other NOSes like Cat8k, CSR, SROS etc.

@kaelemc  Sorry for my dumb questions, but I still am not sure what's the meaning..
A) You want to help with this, but only for vIOS. We can do the same for other platforms without your help.
B) For some reason this should not be done for other platforms, but for vIOS its ok.

And then, should it be done in this PR or better it another?

Thanks for your patience : )

kaelemc:
No problem, it is not a dumb question.

What I mean is we can move the config generation onto the clab side for vIOS only in the current opened PRs, of which I myself am happy to assist you guys on this.

The other vrnetlab platforms can be done as needed, and in other PRs.

DanPartelly:
@torbbang, what exactly do you miss from the IOL L2 image as features? If you don't mind, please write a list of features that do not work for you on the IOS XE image, but work on old  IOS classic image. 

> I want to keep the L2 version as it has a more robust feature-set than the equivalent IOS based IOL L2 images.
>

kaelemc:
apologies for the delay to getting around to this one. I made the relevant changes (on the vrnetlab side too) so that bootstrap config is handled in clab, the partials and override configs should work too, I will later find an L2 image to test.

Pelase make any tweaks to the config if you want, and let me know how this works for you @torbbang

hellt:
anything else to be done here?

kaelemc:
Don't think so, just need someone to sanity check the config generation changes I made (also applies to the vrnetlab PR since the generation was moved from vrnetlab into clab).

@torbbang @tvarohohlavy please take a look when you get some time :)

core/register.go
```diff
@@ -38,6 +38,7 @@ import (
 	clabnodesvr_cat9kv "github.com/srl-labs/containerlab/nodes/vr_cat9kv"
 	clabnodesvr_csr "github.com/srl-labs/containerlab/nodes/vr_csr"
 	clabnodesvr_freebsd "github.com/srl-labs/containerlab/nodes/vr_freebsd"
+	clabnodescisco_vios "github.com/srl-labs/containerlab/nodes/cisco_vios"
```
hellt, line 42: there is also `index.md`, `readme.md` and `vrnetlab.md` files where the links to the new systems should be added

torbbang:
Config works great @kaelemc! Thank you for moving the config generation

core/register.go
torbbang: Added in [3717af1](https://github.com/srl-labs/containerlab/pull/2902/commits/3717af1edf57c989b4a69d9dfffbc5464f9cd50f)

hellt:
thanks all!

sbng:
I tested cisco_vios on 0.72. Works beautifully. Thank you very much.

PR #2914, opened 2025-11-04
VPP version pinning

hellt:
@pimvanpelt it seems the datapath tests started to fail with `latest` tag in the e2e test

This PR pins the version to something available as a tag in your repo, and if these fails I will temporarily remove the vpp test to unblock the pipeline


hellt:
ok, v25.02 worked, but the most recent 25.10 did not. Pinning to 25.02 for the time being

Issue #2910, opened 2025-10-30
[sros] improved support for nokia_srsim nodes running in classic mode


wisotzky:
There are some challenges around the current integration of nokia_srsim running in classic mode:

1. When providing a partial startup config, the mode cannot be changed to “classic”. It’s only possible to keep it in “md mode” or changing it to “mixed mode”. In consequence, one must always provide a complete config, which may resolve in issues, when the same config/pipeline is used for multiple SROS releases.
2. The mapping to Ansible inventory is static. For SROS nodes it always assumes model model, even if the node is running in mixed or classic.

Business impact:

While for any new integration we recommend the usage of model mode (default), in some situations it can be desired to run classic mode. This includes the validation of the SROS Ansible collection, development/regression for traditional EMS/NMS, and DigitalTwin scenarios.

Proposed solution:

- Create a new nokia_srsim specific option, to enable classic mode after the node has successfully started
- Ensure, to populate Ansible inventory in accordance with the mode the SROS nodes is running
- There is no expectation for other advanced methods (like configuration save, …)

Implementation details:

When running in classic mode, the configuration that is automatically added by clab should be adjusted. We may not need to enable the netconf and gRPC servers. We may want to automatically add snmp-community private (RWA) used by NFMP. We may consider to automatically configuring snapshot/rollback used for Ansible. We may consider enabling SNMP persistency.

wisotzky:
You can assign the issue to me. Will have a quick look on how to implement.

PR #2916, opened 2025-11-04
SR-SIM classic CLI support

sacckth:
To Fix issue #2910

### DIFF
--- a/core/inventory.go
+++ b/core/inventory.go
@@ -90,7 +90,7 @@ func (c *CLab) generateAnsibleInventory(w io.Writer) error {
 	}
 
 	for _, n := range c.Nodes {
-		kind, props := c.getAnsibleKindAndProps(n.Config())
+		kind, props := c.ansibleKindAndProps(n.Config())
 		inv.Kinds[kind] = props
 
 		ansibleNode := &AnsibleInventoryNode{
@@ -132,8 +132,8 @@ func (c *CLab) generateAnsibleInventory(w io.Writer) error {
 	return err
 }
 
-// Determine Ansible kind and properties for containerlab node.
-func (c *CLab) getAnsibleKindAndProps(cfg *clabtypes.NodeConfig) (string, *AnsibleKindProps) {
+// ansibleKindAndProps returns kind and properties for a provided containerlab node config.
+func (c *CLab) ansibleKindAndProps(cfg *clabtypes.NodeConfig) (string, *AnsibleKindProps) {
 	ansibleGroup := cfg.Kind
 	ansibleProps := &AnsibleKindProps{}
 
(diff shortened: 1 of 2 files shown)


kaelemc:
Could we instead keep a minimal classic template config and boot from that? This way the node can boot straight into classic instead of provisioning the MD config then connecting to perform the switch.

hellt:
> Could we instead keep a minimal classic template config and boot from that? This way the node can boot straight into classic instead of provisioning the MD config then connecting to perform the switch.

sounds like a good approach

wisotzky:
> > Could we instead keep a minimal classic template config and boot from that? This way the node can boot straight into classic instead of provisioning the MD config then connecting to perform the switch.
> 
> sounds like a good approach

Switching from MD to classic is a legitimate SROS operation, so that should not be a bigger concern.
 
**Caveats (my view):**

- Switching configuration mode "post deploy" requires CLI/SSH access. It increases the chance of errors and impacts the deploy time by 2-3 seconds.
-  Defining the configuration in MD-CLI format, while the node runs in classic mode feels a bit awkward.

**Advantages (my view):**

- We leverage all existing features and configuration generation. That's true for today and for future enhancements. That is DRY (don't repeat yourself) principle at excellence.
- It's very easy to flip the same lab between classic and md-mode, without the need to rebuild the config.

wisotzky:
Some general concerns:
- Building a **Classic CLI** _config.cfg_ pre-deploy seems more complex, as ordering dependencies must be considered
- We still want to enable SNMP persistency, which would require to render bof.cfg
- TLS certificate installation (using NETCONF) would need to be revisited
- Keeping the Classic SR template separate from MD template impacts future evolution (duplication impacts DRY)

kaelemc:
absolutely fair points, I agree with you on the DRY principles. On the flip side I think this enables us for the classic partials and the ability to cleanly separate the classic and MD templates in cases where we want certain things enabled/disabled differingly per config mode.

Regarding the BOF rendering, is there any caveat or issue regarding that @sacckth?. I guess we don't know the mgmt IP until  after deploy stage, which will increase the complexity since we can't start the container unless the BOF is ready.

sacckth:
> absolutely fair points, I agree with you on the DRY principles. On the flip side I think this enables us for the classic partials and the ability to cleanly separate the classic and MD templates in cases where we want certain things enabled/disabled differingly per config mode.
> 
> Regarding the BOF rendering, is there any caveat or issue regarding that @sacckth?. I guess we don't know the mgmt IP until after deploy stage, which will increase the complexity since we can't start the container unless the BOF is ready.

My understanding is that we can change some settings in the BOF using env vars, I haven't RTFM properly. Other option is to use the new SSH wrapper and send the bof configs in PostDeploy. But this would need to add more logic to PostDeploy that is not yet there

kaelemc (approved):

PR #2945, opened 2025-11-16
feat(asav): add support for asav

jmreicha:
Adds support for Cisco ASAv firewall. This change includes registering the new node type and implementing its initialization logic, allowing users to deploy Cisco ASAv nodes within their container-based network labs. Adds basic documentation and an example lab for usage.

I tested a few things locally with the sample lab provided, not sure what else needs to be checked/verified in order to be accepted.

### DIFF
--- a/docs/manual/kinds/cisco_asav.md
+++ b/docs/manual/kinds/cisco_asav.md
@@ -38,7 +38,7 @@ telnet <container-name> 5000
 You can also connect to the container and use `telnet localhost 5000` if telnet is not available on your container host.
 
 /// note
-Default user credentials: `admin:CiscoAsa1!`
+Default user credentials (non-standard due to complexity length requirement): `admin:CiscoAsa1!`
 ///
 
 ## Interface naming


torbbang:
I have both the image available and some time if there is a need for testers

kaelemc:
that would be helpful. I will get to these PRs in the weekend :)

kaelemc (dismissed):

jmreicha:
I forgot about unit tests. Do you want me to add any extras?

kaelemc (approved):

kaelemc:
@jmreicha It's all good from my side, what else did you want to add tests for?

jmreicha:
Oh I just saw you added a test file, which I had totally forgotten about. Seems there wouldn't be that many other things to test anyway so if you're good with it I'm good.

docs/manual/kinds/cisco_asav.md
```diff
@@ -0,0 +1,95 @@
...
+
+```bash
+# from container host
+telnet <container-name> 5000
+```
+
+You can also connect to the container and use `telnet localhost 5000` if telnet is not available on your container host.
+
+/// note
+Default user credentials: `admin:CiscaoAsa1!`
```
hellt: is this a default password set by Cisco? If not, and we are forced to set it we should use the common password `clab@123`

docs/manual/kinds/cisco_asav.md
```diff
@@ -0,0 +1,95 @@
+---
+search:
+  boost: 4
+kind_code_name: cisco_asav
+kind_display_name: Cisco ASAv
+---
+
+# Cisco ASAv
+
+[Cisco ASAv](https://www.cisco.com/c/en/us/products/security/adaptive-security-virtual-appliance-asav/index.html) is identified with `cisco_asav` kind in the [topology file](../topo-def-file.md). It is built using [vrnetlab](../vrnetlab.md) project and essentially is a Qemu VM packaged in a docker container format.
```
hellt: this seems to be a 404 link

docs/manual/kinds/cisco_asav.md
kaelemc: Not a default by Cisco, but if I am not hallucinating in the version I tested there is a complexity length requirement where clab@123 wasn't cutting it.

Anyway I notice a typo in there anyway - should be `CiscoAsa1!`

jmreicha: Updated!

docs/manual/kinds/cisco_asav.md
jmreicha: Should be fixed.

PR #2951, opened 2025-11-21
core/config: Add Git variables to lab topology names

### LINKED ISSUE #2949: Feature suggestion: magic variable for git branch name
Dear maintainers

I'm currently preparing an interop test and I find myself switching branches fairly regularly. And every time before I do such a switch, I must remember to destroy my lab first, because the topology may have changed and this can prevent me from destroying my lab. 

It would be very helpful if I could use a magic variable like __clabGitBranch__ in the lab name, so I can easily check out the branch from which I deployed my lab. I leave it up to your wisdom whether I have suggested a feature or a bug.

Thank you for the amazing tool.

vista-:
A quick proof-of-concept code I whipped up on a long train ride based on #2949.

Works similarly as the magic node name variables that can get substituted in, except for the topology name. I have done some cursory testing, but didn't want to over-commit on that before I get some feedback on the overall idea.

This could help with use-cases involving Containerlab in multi-tenant labs and CI/CD environments, where the runner isn't a temporary environment.5

fix #2949

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -13,7 +13,7 @@ import (
 	"strings"
 
 	gogit "github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing"
+	gogitplumbing "github.com/go-git/go-git/v5/plumbing"
 
 	"github.com/charmbracelet/log"
 	"github.com/pmorjan/kmod"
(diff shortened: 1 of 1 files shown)


core/config.go
```diff
@@ -37,6 +40,10 @@ const (
 	clabDirVar  = "__clabDir__"
 	nodeDirVar  = "__clabNodeDir__"
 	nodeNameVar = "__clabNodeName__"
+
+	// clab name specific variables
+	gitBranchVar = "__gitBranch__"
+	gitHashVar   = "__gitHash__"
```
hellt, line 46: I think it would make sense to record the branch and hash as node labels as well

core/config.go
```diff
@@ -777,3 +791,47 @@ func (c *CLab) magicVarReplacer(nodeName string) *strings.Replacer {
 		nodeNameVar, nodeName,
 	)
 }
+
+// magicVarReplacer returns a string replacer that replaces all supported magic variables.
```
hellt: looks like the comment needs to be adjusted to only mention specific variables?

core/config.go
vista-: I'll add this and move the PR to a ready to review state once I have some free time - hopefully soon.

PR #2980, opened 2025-12-09
go.mod: bump trivial dependencies

tklauser:
Looks like dependabot didn't automatically update some of those in a while.

Bump the dependencies that don't have any API changes to their latest versions and don't break any functionality.


go.mod
```diff
@@ -18,13 +18,13 @@ require (
 	github.com/docker/go-connections v0.6.0
 	github.com/docker/go-units v0.5.0
 	github.com/dustin/go-humanize v1.0.1
-	github.com/florianl/go-tc v0.4.5
+	github.com/florianl/go-tc v0.4.7
```
hellt: looks like tc introduced some breaking changes, looking at the failing test

tklauser: Thanks. I dropped that dep from the update for now.

PR #2999, opened 2026-01-19
Fix SR Linux overlay config copying for large configs 

vista-:
So far, due to the implementation of overlay config copying in the SR Linux node kind, the maximum argument length (`getconf ARG_MAX`) was limited by the OS/kernel of the host running containerlab. Some users might have run into limit when loading particularly long startup configs.

To get around this limit, this PR introduces a different implementation of the overlay (and default) config loading, by using the container runtime's built in file copy to container mechanism instead.

### DIFF
--- a/runtime/podman/podman.go
+++ b/runtime/podman/podman.go
@@ -521,12 +521,12 @@ func (p *PodmanRuntime) CopyToContainer(
 		return fmt.Errorf("error creating tar stream from source file %s: %w", srcPath, err)
 	}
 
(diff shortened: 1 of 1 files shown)


hellt:
thanks @vista-

versluis-org:
Thank you @vista-, did quite some testing and can confirm this fixed the issue and all nodes now boot with expected startup-config. Greatly appreciate all the effort and quick fixing!

PR #3001, opened 2026-01-21
Docs: Enhance documentation for node exec property

robinchrist:
Explicitly document that if shell tokens are used, e.g. '>' to redirect output to a file, the entire command must be wrapped in the shell function

Thanks to @hellt  for pointing this out so quick in the Discord! Thought it's worth adding to the docs as an example.


hellt:
thanks @robinchrist

Issue #3004, opened 2026-01-22
Using a bridge in container namespace prevents startup of ceos


danfai:
When using the notation of `bridge|container` for a node in the same containerlab file, the NO_PROXY environment is filled using the `bridge|container` name.
As the name is not escaped or in quotation in e.g. https://github.com/srl-labs/containerlab/blob/main/nodes/ceos/ceos.go#L109 the commandline for executing the container includes the | as a linux pipe.

As an alternative, it would be good to either:
a) be able to disable the automatic adding of no_proxy environment variables
b) escape the no_proxy string or quote the values.
c) maybe something I didn't consider yet...

Using:
```
$ containerlab version
   version: 0.72.0
   commit: 60c9eee8a
   date: 2025-12-03T12:31:48Z
```

Containerlab file:
```
name: test
topology:
  nodes:
    br|spine:
      kind: bridge
      network-mode: container:spine
    spine:
      kind: linux
      image: alpine:latest
    ceos:
      kind: arista_ceos
      image: ceos:4.35

  links:
    - endpoints: ["ceos:eth1_1", "br|spine:ceos1"]
```

Some debug output:
```
$ docker container inspect clab-test-ceos | jq .[].Config.Cmd
[
  "bash",
  "-c",
[...]

PR #3007, opened 2026-01-23
node ceos: Add quotes for environment on startup

danfai:
This patch adds support for using characters like space ' ' or pipe '|' to the environment of ceos nodes. This is required for clabs, that have bridges in a parent container namespace, as then the no_proxy environment is added for a node with the notation of br|container and the commandline will try to execute the part after the pipe.

fix: https://github.com/srl-labs/containerlab/issues/3004


hellt:
thanks @danfai

PR #3010, opened 2026-01-26
Fix links and template references in inventory.md

ondrej-fabry:
The links were broken since they got moved in July 2025:  https://github.com/srl-labs/containerlab/commits/main/core/export_templates/auto.tmpl


hellt:
thanks @ondrej-fabry

PR #3012, opened 2026-01-26
Added ability for certificates to be automatically loaded into CEOS.

ryan-thornton-bloom:
This will allow the user to auto generate certificates for CEOS without having to add extra lines to the topology.yml file.


hellt:
Thanks @ryan-thornton-bloom ! Good stuff

PR #3018, opened 2026-01-28
Validate node and lab name with json schema

### LINKED ISSUE #3008: Underscore in name: causes resolution failures
Having an underscore character in the topology name results in un-resolvable entries in the /etc/hosts file. In linux, at least, host entries can't have an underscore. This makes commands like containerlab save fail because the hostname will never resolve. Might want to consider either making _ an unusable character in the name or at least displaying a warning.

For example:

```
name: ine_v5_base 

topology:
  defaults:
    kind: cisco_c8000v
    image: vrnetlab/cisco_c8000v:17.16.01a
    restart-policy: always
  nodes:
    R1:
      mgmt-ipv4: 172.20.20.101
```

```
> cat /etc/hosts
.
.
.
172.20.20.101   clab-ine_v5_base-R1 # Kind: cisco_c8000v
```

```
>  ping clab-ine_v5_base-R1
ping: clab-ine_v5_base-R1: Name or service not known

> containerlab save --node-filter R1
16:50:44 INFO Applying node filter: ["R1"]
16:50:44 INFO Parsing & checking topology file=base.clab.yaml
16:50:48 ERRO err: failed to open generic driver for clab-ine_v5_base-R1: dial tcp: lookup clab-ine_v5_base-R1: no such host
```

### LINKED ISSUE #3017: Schema does not validate node name
With current clab schema, there's no limitation on node name. I can name it as for example "srl1:/\=-[]" and it'll break deployment. The same goes for lab name since they are used to name the containers:

```Invalid container name (clab-SRLINUX_/:\-[]-srl1), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed```

Can we add a regex pattern for node name? I suggest to only allow for alphanumeric characters plus `-` and `_` like `^[A-Za-z0-9-_]+$`.

### PR DESCRIPTION
**Pull request by juliusliu02**
- Add pattern check for lab names and node names. (fixes #3017)
- Change minItems for non-array types.

fixes #3008

### DIFF
--- a/schemas/clab.schema.json
+++ b/schemas/clab.schema.json
@@ -1777,12 +1777,12 @@
         "name": {
             "description": "topology name",
             "type": "string",
-            "pattern": "^[a-zA-Z0-9][a-zA-Z0-9.-_]*$"
+            "pattern": "^[a-zA-Z0-9][a-zA-Z0-9-._]*$"
         },
         "prefix": {
             "description": "lab prefix",
             "type": "string",
-            "pattern": "^$|^[a-zA-Z0-9][a-zA-Z0-9.-_]*$",
+            "pattern": "^$|^[a-zA-Z0-9][a-zA-Z0-9-._]*$",
             "markdownDescription": "[lab prefix](https://containerlab.dev/manual/topo-def-file/#prefix)"
         },
         "mgmt": {
@@ -1879,7 +1879,7 @@
                     "markdownDescription": "topology [nodes](https://containerlab.dev/manual/nodes/) configuration container",
                     "type": "object",
                     "patternProperties": {
-                        "^[a-zA-Z0-9][a-zA-Z0-9.-_]*$": {
+                        "^[a-zA-Z0-9][a-zA-Z0-9-._|]*$": {
                             "oneOf": [
                                 {
                                     "type": "null"

### CODE REVIEW

**Comment by hellt** [comment]
hi
the schema won't prevent the lab from failing, it is a visual aid in the editor, unless explicitly checked with external tooling.

To prevent the lab from failing, the regex check should happen in clab code

**Comment by juliusliu02** [comment]
Hi @hellt,

That makes sense. I'll implement the regex check in clab code too then.

**Comment by hellt** [comment]
After thinking about this more, I think I don't want the lab to fail in case **node name** doesn't adhere to the dns label regex

Instead I propose to issue a log warn message with a clear message saying that DNS resolution won't work for the node names. The reason being that many people don't need the DNS resolution for their node names, but they need to have the node hostnames to look like in prod (e.g. `R1_AMS`).

So let's refactor this to be a warning and not an error. The schema bits might still have the stricter nameing rules, as it is not enforcing

**Comment by alexandrehassan** [comment]
How do we want to enforce the container name characters? Should it also just be a warning or would that one cause an error? Specifically talking about https://github.com/srl-labs/containerlab/issues/3017#issuecomment-3814598547

**Comment by hellt** [comment]
we don't want to enforce the node characters by failing the lab. We want to display the warning in the CLI log.

From the json schema perspective we should also not error if the name contains underscores

**Comment by juliusliu02** [comment]
Hey @hellt, as I understand it, we can enforce the regex for container name (fail the lab) when the container won't start (with /\ for example) but only show a warning in CLI for DNS invalid characters `._`. Is that correct?

**Comment by hellt** [comment]
correct
don't fail when the lab starts but DNS resolution wont work - only warn

and fail if the name contains illegal characters when the lab wont' even start.

The schema should be

```
"pattern": "^[a-zA-Z0-9][a-zA-Z0-9-]*$"
```

allowing non dns names, but disallowing illegal characters

File: core/config.go [hunk 1]
```diff
@@ -204,6 +212,18 @@ func (c *CLab) createNodeCfg( //nolint: funlen
 		longName = fmt.Sprintf("%s-%s", c.Config.Name, nodeName)
 	}
 
+	if !containerNamePatternRe.MatchString(longName) {
+		return nil, fmt.Errorf("Node name contains invalid characters: %s", longName)
+	}
```
hellt: the smoke test No24 is failing because we have a special notation for bridges that are created in some parent namespace

https://containerlab.dev/manual/kinds/bridge/#bridges-in-container-namespace

this means this check either needs to be done later when the name is parsed by a particular node or add `|` to allowed charset

juliusliu02: Thanks for letting me know! I moved the validation into checkDeploymentCondition in DefaultNode and updated the json schema accordingly. I couldn't get the robot test to work locally but it seems to work with manual testing. Can you run the pipeline again please?

PR #3020, opened 2026-01-30
[SR-OS]: reject SR-SIM image when kind is nokia_sros and vice-versa

sacckth:
When kind is set to nokia_sros (vrnetlab), verify the Docker image is not the SR-SIM container image by checking the label
org.opencontainers.image.title. If the label value is "srsim", return a clear error instructing the user to use kind: nokia_srsim with that image, or a vrnetlab nokia_sros image for kind: nokia_sros.

The check runs in PreDeploy after images are pulled. Runtimes that do not support image inspection (e.g. Podman, Ignite) skip the check.

### DIFF
--- a/nodes/sros/sros_test.go
+++ b/nodes/sros/sros_test.go
@@ -158,9 +158,6 @@ func Test_sros_verifyNokiaSrsimImage(t *testing.T) {
 	})
 }
 
-// boolPtr returns a pointer to b.
-func boolPtr(b bool) *bool { return &b }
-
 func Test_sros_buildStartupConfig(t *testing.T) {
 	t.Run("full_config_from_file", func(t *testing.T) {
 		dir := t.TempDir()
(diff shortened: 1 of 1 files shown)


hellt:
Do we want the same check on srsim kind?

kaelemc:
Can we also put a warn message if someone is using `components` with sros kind? as they need to use `type` and construct that lovely timos line :)

nodes/sros/sros_test.go
```diff
@@ -65,3 +75,140 @@ func TestComponentSorting(t *testing.T) {
...
+		n.Cfg = &clabtypes.NodeConfig{ShortName: "n1", Image: "img"}
+		n.WithRuntime(mockRt)
+		err := n.verifyNokiaSrsimImage(ctx)
+		assert.NoError(t, err)
+	})
+}
+
+// boolPtr returns a pointer to b.
+func boolPtr(b bool) *bool { return &b }
+
```
kaelemc: I think this is unused?

sacckth: Cursor!

PR #3021, opened 2026-01-30
Nokia SRSIM config generation simplification

sacckth:
- Fix duplicate {{ .SNMPConfig }} in sros_config_sros25.go.tpl
- Move embedded configs from version.go to configs_embed.go
- Introduce config variant (mode, family, secure gRPC) and single resolver
- Refactor applyNodeTypeSpecificConfig and selectConfigTemplate to use variant
- Rename getConfigTemplate to buildStartupConfig and document flow
- Add version placeholder in getTemplateForVariant for SR OS 26+
- Table-driven snippet resolution: getFullSnippetSet(v) in prepareConfigTemplateData
- generateComponentConfig: struct + loop (component_config.go, componentCfgLine)

Compatibility: no topology or runtime changes; same generated config.

### DIFF
--- a/.github/workflows/srsim-tests.yml
+++ b/.github/workflows/srsim-tests.yml
@@ -77,7 +77,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Pull and re-tag srsim image
-        run: docker pull ghcr.io/srl-labs/containerlab/nokia_srsim:25.10.R1 && docker tag ghcr.io/srl-labs/containerlab/nokia_srsim:25.10.R1 registry.srlinux.dev/pub/nokia_srsim:25.10.R1
+        run: docker pull ghcr.io/srl-labs/containerlab/nokia_srsim:25.10.R2 && docker tag ghcr.io/srl-labs/containerlab/nokia_srsim:25.10.R2 registry.srlinux.dev/pub/nokia_srsim:25.10.R2
 
       # - name: setup tmate session
       #   uses: mxschmitt/action-tmate@v3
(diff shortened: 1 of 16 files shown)


hellt (approved):

kaelemc (approved):
I floated to @sacckth that we should infact disable the default card config generation when `components` is defined as there are a few cases where it causes unnecessary issues.

For example some chassis layouts make you define a card type which is not actually valid in the configuration, so the generated config would cause the whole configuration to be rejected even though a valid SR-SIM chassis is there.

There is one option to have some var/file containing all the valid types which we *will* gen the configuration for. but it will something else to maintain, and maybe not worth it.

Got any thoughts @hellt ?

hellt:
if you won't auto-configure when components are defined, what will user do? How they will configure it? How large the portion of the config is?

kaelemc:
The auto configure is basically a QoL thing. just a helper to bring the cards up. the config is really minimal.

hellt:
Containerlab is about QoL, so we need to understand the burden of maintaining the special cases or moving it to the user' shoulders.

kaelemc:
Ok. I would put it this way. Without the auto-config it's like the vSIM. You would build your TiMOS line or whatever, but you still needed to go into the CLI and provision the card, xiom, mda/whatever and set it as admin up.

The auto-config works great for removing those steps, except for certain chassis where card type != cli configured type (ie. even if it's a distributed model, the cpm and iom is a single type as `cpm-xyz/iom-xyz`.

My solution is to just pull all the configurable types from the 7x50 YANG models, and basically when we go to generate the configuration, we check if the `components:` defined type is one that is seen in the YANG model, if not, let's not generate the config for it.

If we're ok with that I'm happy to PR it.

hellt:
> except for certain chassis where card type != cli configured type

How many configuration of this type we have?

kaelemc:
Mr @sacckth likely knows better than me. but from a quick skim of the docs I identify at least 4-5 chassis which have various different types like this.

kaelemc:
Will work on something re: this config gen thing with @sacckth. He has some good ideas.

PR #3023, opened 2026-01-31
Netem: Add --node and --container Options with Show All Impairements

### LINKED ISSUE #3022: Netem Not Working for Distributed SR-SIM Devices
Hey!

The containerlab tools netem set/show/reset commands fail when used with distributed SR-SIM (nokia_srsim) devices. Because the netem CLI bypasses the node abstraction and directly queries the container runtime, which doesn't work for distributed deployments where the logical node maps to multiple containers.

For example, I have this node:

```
 nodes:
   srsim10:
     kind: nokia_srsim
     type: SR-7s
     components:
       - slot: A
         type: cpm5
       - slot: B
         type: cpm5
       - slot: 1
          type: iom5-e
```

This creates separate containers per component:
- clab-lab-srsim10-a (CPM-A) - owns the network namespace
- clab-lab-srsim10-b (CPM-B) - shares namespace via NetworkMode: container:srsim10-a
- clab-lab-srsim10-1 (line card) - shares namespace

**Important:** There is no container named clab-lab-srsim10. There is only the component containers exist. 

When running netem commands:

```
containerlab tools netem set -n clab-lab-srsim10 -i e1-1-1 --delay 100ms
```

The CLI calls rt.GetNSPath(ctx, "clab-lab-srsim10") directly on the container runtime. I found this in the cmd/tools_netem.go, line 192. Since no container with that name exists, the command fails.

```                                                                                                      
nodeNsPath, err := rt.GetNSPath(ctx, o.ToolsNetem.ContainerName)
```

The netem:
1. Does not load the topology file
2. Does not use the node's GetNSPath() method which properly resolves to CPM-A's namespace
3. Treats the --node/-n flag as a direct container name, not a topology node name

## Proposed Solution

Modify the netem CLI to support two modes of operation:

1. Node mode (new): Use -t/--topo with -n/--node to look up the node from the topology and use its proper GetNSPath() method 
2. Container mode (existing, renamed): Use -c/--container for direct container name access

For example:
```
containerlab tools netem set -t topology.clab.yml -n srsim10 -i e1-1-1 --delay 100ms
containerlab tools netem set -c clab-lab-srsim10-a -i e1-1-1 --delay 100ms   
```

Let me know if you have any other proposition to  solve this issue. I am willing to create a MR to solve this issue. We need the link impairments working in the SR-SIM devices.

bayars:
# Summary                                                                                                                                                                                 
Adds topology-aware node selection and bulk impairment viewing to netem commands.                                                                                                          
                                                                                                                                                                                      
- Add `--node/-n` option to select nodes by name from the topology file (requires `--topo/-t`)                                                                                             
- Add `--container/-c` option for direct container access without topology                                                                                                                 
- Add show-all functionality: when `--topo` is provided without `--node` or `--container`, display impairments for all nodes in the topology                                               
- Improve namespace path resolution to support sr-sim nodes with different container layouts  

Closes #3022


FloSch62:
@bayars 

1. I am ok with the -c option, but it will be a breaking change with the vscode extension as it changes the default behavior of the command, so this needs to updated as well. 

2. Problem: Show-all JSON output drops nodes that share a namespace
```
Example:
  - Topology has nodes lc1 and lc2 that (for whatever reason) share the same netns (nsPath = X).
  - Loop hits lc1 first: computes impairments for X, writes JSON entry for lc1.
  - Loop hits lc2: sees X already processed, skips it.
  - JSON contains only "lc1": [...] and no "lc2": [...], even though lc2 is in the topology.
```

3. Please als update smoke-tests (08*.robot) and also please the documentation for the command.

cmd/tools_netem.go
```diff
@@ -403,36 +464,39 @@ func qdiscToJSONData(qdisc *gotc.Object) clabtypes.ImpairmentData {
...
+	return showNodeImpairments(ctx, o, nodeNsPath, displayName)
+}
+
+func showNodeImpairments(
+	ctx context.Context,
+	o *Options,
+	nodeNsPath string,
+	displayName string,
+) error {
+	_ = ctx // reserved for future use
```
FloSch62: What is the future use?

cmd/tools_netem.go
```diff
@@ -486,36 +550,112 @@ func netemShowFn(o *Options) error {
...
 
-	err = rt.Init(
-		clabruntime.WithConfig(
-			&clabruntime.RuntimeConfig{
-				Timeout: o.Global.Timeout,
-			},
-		),
-	)
+	// Track nodes we've already processed (for distributed nodes sharing namespace)
+	processedNSPaths := make(map[string]bool)
```
FloSch62: - Dedupe-by-namespace that causes a drop: cmd/tools_netem.go:562-577
      - processedNSPaths := make(map[string]bool) at :563
      - skip on already-seen ns path at :572-576 (if processedNSPaths[nsPath] { continue })
  - JSON map population happens only for non-skipped nodes: cmd/tools_netem.go:578-585
      - allImpairments[nodeName] = impairments at :584 (never executed for skipped nodes)

Example:
  - Topology has nodes lc1 and lc2 that (for whatever reason) share the same netns (nsPath = X).
  - Loop hits lc1 first: computes impairments for X, writes JSON entry for lc1.
  - Loop hits lc2: sees X already processed, skips it.
  - JSON contains only "lc1": [...] and no "lc2": [...], even though lc2 is in the topology.

FloSch62 (changes requested):

hellt:
we can park this until sr-sim supports netem rules; for now adding a bridge between the srsim interfaces and setting netems on the bridge interface is the workaround

cmd/tools_netem.go
bayars: I left it when I was developing. It's removed now.

cmd/tools_netem.go
bayars: Good catch, I fixed, and tested. I replaced the processedNSPaths skip-map with a nsPathImpairments cache-map. Now on first encounter of a namespace we query and cache the impairments; on subsequent encounters we  reuse the cached result.

Critically, allImpairments[nodeName] = cached executes for every node regardless of whether the namespace was seen before, so lc1 and lc2 both appear in the JSON output.

For table output, the processedNSPaths guard was removed entirely. Every node gets its header and table printed.

bayars:
> we can park this until sr-sim supports netem rules; for now adding a bridge between the srsim interfaces and setting netems on the bridge interface is the workaround

Yes, that works but I can't expect all people to create a bridge between sr-sim interfaces every time. 

Is there any news from the SROS team about this? Is there a timeline for when this fix will be available in srsim?

FloSch62 (dismissed):
lgtm

hellt:
> Is there any news from the SROS team about this? Is there a timeline for when this fix will be available in srsim?

reach to the sr os team for details

sacckth:
> > Is there any news from the SROS team about this? Is there a timeline for when this fix will be available in srsim?
> 
> reach to the sr os team for details

As indicated on #3022 no estimated time to fix this. I will add a note on the docs indicating netem is currently not supported directly on SR-SIM. And I guess we can close this

kaelemc:
thanks all. continuation in https://github.com/srl-labs/containerlab/pull/3409

PR #3025, opened 2026-02-03
Lowercase the slot in mgmt IP getter

kaelemc:
Fix bug whereby we wouldn't get the mgmt IP if only CPM card were defined because the mgmt IP getter was not lowercasing the slot number like the rest of the code did.


hellt:
oops, I haven't spotted the review request from @sacckth 

I wonder if we should have a test for this?

kaelemc:
> oops, I haven't spotted the review request from @sacckth
> 
> I wonder if we should have a test for this?

Something else will negate this later anyways.... but this can be our interim bugfix :)

PR #3030, opened 2026-02-03
add missing sfm types to clab schema

alexandrehassan:
SROS type was missing a couple of sfm types

https://documentation.nokia.com/sr/23-7-2/books/md-cli-command-reference/sfm-sfm_0.html#d57829

### DIFF
--- a/schemas/clab.schema.json
+++ b/schemas/clab.schema.json
@@ -1765,7 +1765,10 @@
                 "sfm-x20-b",
                 "sfm-x20s-b",
                 "sfm2-s",
-                "sfm2-x20s"
+                "sfm2-x20s",
+                "sfm-ixr-6",
+                "sfm-ixr-10",
+                "m-sfm6-12e"
             ]
         }
     },


kaelemc:
Are these types actually supported in 25.7 onwards? (ie. do they actually work with SRSIM) 

The schema was built from the 25.7 YANG model, so I just wanna make sure they actually are compatible :)

alexandrehassan:
Checking through the SROS/SRSIM doc `sfm2-x20s` appears to be in the supported sfm types (Relevant doc link)[https://documentation.nokia.com/sr/25-10/7x50-shared/srsim-installation-setup/appendices.html#tzp_gbj_xfc].

I also know that a lab with this node in it works perfectly fine (stripped out any information that isn't relevant).

```yaml
    BP1:
      kind: nokia_srsim
      image: nokia_srsim:25.10.R2
      type: sr-12e
      components:
        - slot: 3
          type: iom4-e
          mda:
            - slot: 1
              type: me6-10gb-sfp+
          sfm: m-sfm6-12e
```

I also think that if the list of sfm exists and isn't scoped to srsim version it needs to include every sfm value that would work in any version.

PR #3029, opened 2026-02-03
Bump docker to 28.5

FloSch62:

PR #3033, opened 2026-02-05
Adding support for Rocky 10 to utils/quick-setup.sh

Oliver-Mustoe:
# Summary
- To utils/quick-setup.sh adds an and statement to the RHEL check to check if Rocky 10 is the host OS, and if so then setting Docker version 28.5.2

### DIFF
--- a/utils/quick-setup.sh
+++ b/utils/quick-setup.sh
@@ -327,4 +327,3 @@ function all {
 }
 
 "$@"
-


utils/quick-setup.sh
kaelemc, line 6: @hellt as I understand as of #3029 we are on docker 28.5 now, so probably we can remove some of the logic that bumped it for specific versions and just bump this global var? (ofc unless diff distros have diff minor vers available)

hellt: I think this requires testing, if e.g. debian 12 has 28.5 in docker's pkg registry
same for ubuntu and rhel

so quite a matrix to validate

PR #3036, opened 2026-02-05
SR-SIM: write chassis_info.json to node LabDir 

sacckth:
Works from 25.10.R2 onwards.
- Add ReadFileFromImageInspect in version.go to read files from image graph driver (UpperDir/MergedDir), refactor readVersionFromImageLayers to use it
- Add chassis_info.go with DefaultChassisInfoPath and license header for the generated file
- In createSROSFiles(ctx), call writeChassisInfoToLabDir to fetch /opt/nokia/chassis_info.json from the image and write to LabDir with same filename and BSD-3-Clause license header
- Update README to document chassis_info.json in lab dir


kaelemc (approved):
go! @sacckth

PR #3038, opened 2026-02-06
fix: allow destroy by name without topology file

### LINKED ISSUE #2524: Destroy with only name specified
The setting is the same as in #2523.

I am using Docker image of containerlab and during development, when something goes wrong I sometimes find myself with lost topology file and a dozen of started containers of container lab. In this case I want to destroy all containers from a lab by the name and clenup ssh if set.

In CI I never need to destroy managment network, so destroying by name should make my CI environment simpler.

WDYT about cutting destroy dependency on topology file if --keep-mgmt-net is set and we have name from #2523.

ndtobs:
Fixes #2524

When using 'clab destroy --name <lab>' without a topology file, the command now works by using only the lab name to find and remove containers, rather than requiring the topology file to exist.

This is useful when:
- The topology file was accidentally deleted
- Running in CI where the topo file isn't persisted
- Managing labs from a different directory

The fix adds a new WithLabNameOnly() option that sets only the lab name without attempting to load or verify the topology file.

### DIFF
--- a/core/destroy.go
+++ b/core/destroy.go
@@ -118,6 +118,23 @@ func (c *CLab) makeCopyForDestroy(
 ) (*CLab, error) {
 	newOpts := []ClabOption{
 		WithTimeout(c.timeout),
+	}
+
+	// Try to load topology file if it exists, otherwise use lab name only.
+	// This allows destroying labs even when the topology file has been deleted.
+	// WithTopoPath / WithLabNameOnly must run before WithNodeFilter so filterClabNodes
+	// sees topology nodes (see filterClabNodes in clab.go).
+	if clabutils.FileOrDirExists(topo) {
+		newOpts = append(newOpts, WithTopoPath(topo, c.TopoPaths.VarsFilenameAbsPath()))
+	} else {
+		// Derive lab name from lab directory (format: clab-<labname>)
+		labName := filepath.Base(labDir)
+		labName = strings.TrimPrefix(labName, "clab-")
+		log.Debugf("topology file %q not found, using lab name %q for destroy", topo, labName)
+		newOpts = append(newOpts, WithLabNameOnly(labName))
+	}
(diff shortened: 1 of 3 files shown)


cmd/options.go
```diff
@@ -255,7 +259,12 @@ func (o *GlobalOptions) toClabOptions() []clabcore.ClabOption {
 	}
 
 	if o.TopologyFile == "" && o.TopologyName != "" {
-		options = append(options, clabcore.WithTopologyFromLab(o.TopologyName))
+		if o.LabNameOnly {
+			// Use lab name only without requiring topology file (e.g., for destroy)
```
hellt: it seems that `if o.LabNameOnly` is exactly the same condition that the one line above `if o.TopologyFile == "" && o.TopologyName != ""` which suggests that the `else` clause is never executed.

You likely don't need another option (LabNameOnly), and can just append different options based on one condition of `if o.TopologyFile == "" && o.TopologyName != ""`

hellt:
also I am not sure this PR solves anything new. The goal of this PR should

The existing 0.72.0 version already supports destroying labs by name:

```
clab des --name linux   
13:24:00 INFO Parsing & checking topology file=linux.clab.yml
13:24:00 INFO Parsing & checking topology file=linux.clab.yml
13:24:00 INFO Destroying lab name=linux
13:24:00 INFO Removed container name=clab-linux-linux_1
13:24:00 INFO Removing host entries path=/etc/hosts
13:24:00 INFO Removing SSH config path=/etc/ssh/ssh_config.d/clab-linux.conf
```

but it requires a topology file to be present.

What is needed, is for destroy to work without the topology file, which I think this PR does not solve for?

ndtobs:
Thanks for the feedback.  I simplified the approach and tested the fix.  Should work as intended now.

PR #3039, opened 2026-02-06
Fix bug where vxlan-stitch interface type was not stitched

mgreve:
Fixes https://github.com/srl-labs/containerlab/issues/3040

Containerlab vxlan-stitch interfaces when used in containerlab topology files are not actually stitched. This PR fixes this issue.

To be honest, I don't particularly like this approach. The underlying issue/bug is that a vxlan-stitch interface never gets deployed, but only its underlying vxlan and veth interfaces are deployed. This means the tc rules to stitch are never added as they are only called as part of the deploy of the vxlan-stitch interface. 

In my opinion, a better fix would be to call deploy on the vxlan-stitch interface and not its underlying interfaces. The underlying interfaces are captured in the special host links. This capturing is intrinsically tied to the code for these interface types, and it doesn't seem quite obvious how to change this without a larger refactoring.

### DIFF
--- a/core/deploy.go
+++ b/core/deploy.go
@@ -168,7 +168,7 @@ func (c *CLab) Deploy( //nolint: funlen
 		// the veth and vxlan underlying links have already been deployed.
 		err = vxlanStitchLink.DeployWithExistingVethAndVxlan(ctx)
 		if err != nil {
-			log.Warnf("failed deploying vxlan-stitch endpoint %s", vxlanStitchLink)
+			log.Warnf("failed deploying vxlan-stitch endpoint %+v", vxlanStitchLink)
 		}
   }
 


mgreve-akamai:
@hellt or anyone else: Would it be possible to get the workflow approved from a maintainer? A review or fix would also very much be appreciated since this is blocking a real use case. I'm also happy to adapt the PR if needed.

hellt:
@mgreve we have an auto-test that should have tested the vxlan stitching functionality and it is passing in `main`. https://github.com/srl-labs/containerlab/blob/main/tests/08-vxlan/02-vxlan-stitch.robot

mgreve:
Thanks, @hellt. This test looks a bit odd to me, and it's specifically *not* checking if the tc stitching rules are being setup. I can dig a bit deeper into the test to see if the stitching rules are there. I believe they won't be due to the bug reported here.

I will also dig into why existing tests are failing with my PR.

hellt:
The test should check the end to end connectivity between the nodes, which shouldn't work without the tc rules stitching the connectivity. Granted, the test may be flawed, feel free to check it out as the container images used in it are all public

mgreve-akamai:
This test fails on my machine with main. The tc ingress rules are not setup. Grepping through the log output from my machine I also can't see the expected lines, but the containers are setup correctly, the interfaces are created. Checking filters with tc, I see nothing on the veth interfaces.

These are the lines I would expect to see
```
.. INFO configuring ingress mirroring with tc in the direction of ...
```

@hellt , would it be possible to grab the output xml file from a Github actions runner to validate if the log lines for the tc stitching are there. This would be the file:
```Output:  /home/runner/work/containerlab/containerlab/tests/out/02-vxlan-stitch.robot-docker-out.xml``` from any PR.

Looking at the artifact upload, it looks like only the html report is uploaded, not the xml file, see e.g. https://github.com/srl-labs/containerlab/actions/runs/22290223244/job/64476169964
and
```
actions/upload-artifact@v6
  with:
    name: 08-vxlan-log-02.robot-docker
    path: ./tests/out/*.html
```

hellt:
please check #3083

mgreve:
Thank you. I will check it asap. Btw. I think you mean https://github.com/srl-labs/containerlab/pull/3083 (not 3084).

hellt:
> Thank you. I will check it asap. Btw. I think you mean #3083 (not 3084).

oh, yeah, made a typo :D 

since you confirmed #3083 worked for you, I will close this one

PR #2932, opened 2025-11-12
Disable TX checksum offload on host-side veth for SRSIM mgmt intf

kaelemc:
Fixes #2894 

We needed to disable the checksum offload on the veth pair for the mgmt interface of the SR-SIM node. is not as simple as disabling it for the eth0 interface.

I will add some tests soon. @sacckth please confirm this works if you can :)

Issue #2894, opened 2025-10-16
Use disable tx offload on bridges and interfaces involved in sr-sim


hellt:
SR-SIM uses the strict validation of the tcp checksum offload on its management interface and this is what causes connectivity issues when containers outside of the same docker network try to talk to sr-sims.

We disable tx checksum offload on the veth links that we create for the datapaths - https://github.com/srl-labs/containerlab/blob/9d509f34d62254cf7405a7780377a36bfcd4fb6f/links/link_veth.go#L175

And we also try to disable offload on the docker network bridge - https://github.com/srl-labs/containerlab/blob/9d509f34d62254cf7405a7780377a36bfcd4fb6f/runtime/docker/docker.go#L459

But it seems that this does not help and we need to disable the tx checksum offload on the mgmt veth endpoint as well - see https://discord.com/channels/860500297297821756/865574858385784842/1428291519521226833

@sacckth @kaelemc

sacckth:
Yes, I believe it had been reported before, but somehow I failed to find the root cause... Will look into a fix or maybe @kaelemc is already on it?

kaelemc:
Not on it right now, but can do it in the coming days if you're busy.

steiler:
This is basically what needs to happen I guess.
I'm just not sure, is it required per linecard also or just per MDA....

I'm sure you guys will find the right spot in the functions called to add this snippet.
```
	// disable tx offloading
	ethtxoff := utils.NSEthtoolTXOff(n.Cfg.LongName, "eth0")
	err := n.ExecFunction(ctx, ethtxoff)
	if err != nil {
		return err
	}
```

PR #3045, opened 2026-02-09
Fix distributed SR-SIM unreachable with --skip-post-deploy

bayars:
# Summary

This happened on one of the SR-SIM distributed topologies. The topology included other Linux containers. The management address was reachable when I SSH from the host. However, the SR-SIM nodes were not reachable when the Linux containers tried to SSH into them.

These are the discovered issues that might be causing the problem:
- Fix distributed SR-SIM (CPM + line cards) becoming unreachable when deployed with --skip-post-deploy
- Fix PostDeploy timeout loop that never iterated due to recomputing the deadline each iteration
- Add detailed documentation for --skip-post-deploy in deploy and redeploy command references

## Root Cause

PR #2932 added TX checksum offload disabling inside PostDeploy() to fix SR-SIM management reachability (issue #2894). However, --skip-post-deploy skips PostDeploy() entirely, reintroducing the reachability problem for distributed setups where the CPM's host-side management veth keeps TX offload enabled, causing invalid checksums.

### Move TX offload disable to DeployEndpoints() in nodes/sros/sros.go

The TX offload disable is a networking concern, not a configuration/validation step. It is now performed in a new DeployEndpoints() override that calls DefaultNode.DeployEndpoints() first, then disables TX checksum offload on the host-side veth.

DeployEndpoints always runs (not gated by --skip-post-deploy) and executes after endpoints are created, so the veth pair exists.

### Fix PostDeploy timeout loop (nodes/sros/sros.go)

time.Now().Before(time.Now().Add(readyTimeout)) was always true — the deadline is now computed once before the loop.

### Documentation improvement for deploy.md and redeploy.md

Added detailed explanation of what the post-deploy phase does, listing the specific actions (readiness checks, TLS provisioning, config saving, overlay CLI, hosts population, TX offload) and guidance on when to use --skip-post-deploy.

### DIFF
--- a/docs/cmd/deploy.md
+++ b/docs/cmd/deploy.md
@@ -143,18 +143,20 @@ Read more about [node filtering](../manual/node-filtering.md) in the documentati
 
 #### skip-post-deploy
 
-The `--skip-post-deploy` flag can be used to skip the post-deploy phase of the lab deployment. This is a global flag that affects all nodes in the lab.
+The `--skip-post-deploy` flag skips the post-deploy phase of the lab deployment, affecting all nodes.
 
-The post-deploy phase runs after containers are created and network endpoints are deployed. Node startup configuration is already loaded at boot time (mounted during the deploy phase). The post-deploy phase performs validation and additional setup on top of the booted node:
+The post-deploy phase runs after containers and network endpoints are created. Depending on the node kind, it may include:
 
-- **Readiness and health checks:** Polling the node until it is fully booted and verifying it is healthy. Container logs are monitored for errors such as rejected configurations.
-- **TLS certificate provisioning:** Generating or loading certificates used for secure management access.
-- **Saving startup configuration:** Persisting the running configuration so it survives restarts (Nokia SR OS).
-- **Overlay CLI configuration:** Applying additional CLI configuration on top of the startup config and committing it (Nokia SR Linux).
-- **Populating `/etc/hosts`:** Adding peer node entries for in-band name resolution.
-- **Disabling TX checksum offload:** Adjusting management interface settings where required.
+- Readiness and health checks
+- TLS certificate provisioning
+- Saving startup configuration
+- Applying overlay CLI configuration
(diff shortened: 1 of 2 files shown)


docs/cmd/deploy.md
```diff
@@ -145,6 +145,17 @@ Read more about [node filtering](../manual/node-filtering.md) in the documentati
 
 The `--skip-post-deploy` flag can be used to skip the post-deploy phase of the lab deployment. This is a global flag that affects all nodes in the lab.
 
+The post-deploy phase runs after containers are created and network endpoints are deployed. Node startup configuration is already loaded at boot time (mounted during the deploy phase). The post-deploy phase performs validation and additional setup on top of the booted node:
```
sacckth: This seems to go very in depth while probably half of the bullet points apply to only a few kinds

docs/cmd/redeploy.md
```diff
@@ -86,6 +86,17 @@ Do not try to remove the management network during destroy phase. Usually the ma
 
 The `--skip-post-deploy` flag can be used to skip the post-deploy phase of the lab deployment. This is a global flag that affects all nodes in the lab.
 
+The post-deploy phase runs after containers are created and network endpoints are deployed. Node startup configuration is already loaded at boot time (mounted during the deploy phase). The post-deploy phase performs validation and additional setup on top of the booted node:
+
+- **Readiness and health checks:** Polling the node until it is fully booted and verifying it is healthy. Container logs are monitored for errors such as rejected configurations.
```
sacckth: See above...

docs/cmd/deploy.md
bayars: docs are updated again. I specified kinds names. Let me know if there is more kinds affected.

docs/cmd/redeploy.md
bayars: Same for here. I trimmed this part too.

sacckth:
any thoughts before merging @kaelemc ?

PR #3055, opened 2026-02-11
Start stop nodes

FloSch62:
The `start/stop/restart` command starts/stops/restarts one or more nodes in a deployed lab and restores parked dataplane
interfaces 

Tested:
- nokia_srl
- nokia_srsim (standalone and distributed)
- nokia_sros 
- juniper_vmx


core/stop.go
```diff
@@ -0,0 +1,513 @@
...
+	for nodeName := range allNodes {
+		nodeNames = append(nodeNames, nodeName)
+	}
+
+	slices.Sort(nodeNames)
+
+	return nodeNames
+}
+
+func (c *CLab) stopNode(ctx context.Context, n clabnodes.Node) error {
```
steiler: stopNode should become a Stop function on the node itself.

if there is different styles on how to stop a Node, that can be overwritten via the actual node, but the DefaultNode could cover the general case.

links/endpoint.go
```diff
@@ -46,6 +46,17 @@ type Endpoint interface {
 	SetIfaceAlias(string)
 	// GetVars returns the endpoint-level vars.
 	GetVars() map[string]any
+	// MoveBetween moves the endpoint interface between source and destination nodes.
+	MoveBetween(context.Context, Node, Node, *MoveOptions) error
+	// SetUpIn sets the endpoint interface up in the provided node namespace.
+	SetUpIn(context.Context, Node) error
+	// SetDownIn sets the endpoint interface down in the provided node namespace.
+	SetDownIn(context.Context, Node) error
```
steiler: The Node param is not required here, since the endpoint is actually assigned to a certain node. So

```
Up(context.context)
Down(context.context)
```
must be sufficent

links/endpoint.go
```diff
@@ -46,6 +46,17 @@ type Endpoint interface {
 	SetIfaceAlias(string)
 	// GetVars returns the endpoint-level vars.
 	GetVars() map[string]any
+	// MoveBetween moves the endpoint interface between source and destination nodes.
+	MoveBetween(context.Context, Node, Node, *MoveOptions) error
```
steiler: This should be MoveTo and the parameter should only carry the To Node.
The endpoint should already be attached to a certain node
https://github.com/srl-labs/containerlab/pull/3055/changes#diff-f40fa0e0a90864c8c28820a32323a480d0e778f510e6b31ad2bff39fde1202ddL17

links/endpoint.go
```diff
@@ -209,3 +280,43 @@ func CheckEndpointDoesNotExistYet(ctx context.Context, e Endpoint) error {
...
 	})
 }
+
+type endpointLookup interface {
+	GetIfaceName() string
+	GetIfaceAlias() string
+	String() string
+}
+
+func findLinkForEndpoint(ep endpointLookup) (netlink.Link, error) {
```
steiler: This is basically resolving the endpoint to a netlink interface from what I can tell... so it should probably be part of the endpoint itself... it is linked to a node, which contains the netns ... and can thereby resolve the netlink interface...

steiler (commented):
just a couple of comments

core/stop.go
FloSch62: Done. Stop(context.Context) error and Start(context.Context) error are now on the Node interface, with DefaultNode covering the general case.

links/endpoint.go
FloSch62: Done, there's no longer a Node parameter.

links/endpoint.go
FloSch62: For the restore-from-parking case we now use EndpointOnNode(ep, parkingNode).MoveTo(ctx, ownerNode, opts)

links/endpoint.go
FloSch62: Done. It's now (e *EndpointGeneric) resolveLink() (netlink.Link, error)

FloSch62:
vxlan tests fail, because of https://github.com/srl-labs/containerlab/issues/3056

FloSch62:
For srsim distributed, I may would need some help from you @kaelemc and @sacckth , but imho that can also be added later.

links/endpoint.go
steiler: why not init the parking namespace (Node) as a general Node and utilize the same means to move the interfaces back?
In the end it is the exact same action, just the source and destination nodes are swapped.

FloSch62: you are right, 

now parking netns is initialized as a regular GenericLinkNode and we use the same endpoint MoveTo flow in both directions, just swapping src/dst.

On start we first set endpoints to parking as source, because stop/start can happen in separate process invocations

nodes/default_node.go
```diff
@@ -66,6 +66,7 @@ type DefaultNode struct {
 	// State of the node
 	state      clabnodesstate.NodeState
 	statemutex sync.RWMutex
+	StopSignal string
```
steiler:
```
type Signal string

const (
	SIGTERM Signal = "SIGTERM"
	SIGINT  Signal = "SIGINT"
	SIGKILL Signal = "SIGKILL"
)
```

kaelemc: thx, done.

nodes/default_node.go
```diff
@@ -794,3 +797,260 @@ func (d *DefaultNode) GetContainerStatus(ctx context.Context) clabruntime.Contai
...
+// ParkEndpoints parks dataplane interfaces into a dedicated network namespace
+func (d *DefaultNode) ParkEndpoints(ctx context.Context) error {
+	cfg := d.Config()
+
+	parkName := clabutils.ParkingNetnsName(cfg.LongName)
+	parkPath, err := ensureNamedNetns(parkName)
+	if err != nil {
+		return fmt.Errorf("node %q failed creating parking netns: %w", cfg.ShortName, err)
+	}
+	parkingNode := clablinks.NewGenericLinkNode(parkName, parkPath)
```
steiler: Why not create a specific ParkingNode... that does all of this.

nodes/default_node.go
```diff
@@ -794,3 +797,260 @@ func (d *DefaultNode) GetContainerStatus(ctx context.Context) clabruntime.Contai
...
+
+	// Move dataplane interfaces into the parking netns while the container netns is still alive.
+	moved, err := moveEndpoints(
+		d.GetEndpoints(),
+		func(ep clablinks.Endpoint) error {
+			return ep.MoveTo(ctx, parkingNode, preMoveSetDownOptions())
+		},
```
steiler: I don't get this... via d.GetEndpoints you already have all endpoints at hand. Iterate over these and execute the MoveTo function.... why this weired indirection?!

nodes/default_node.go
```diff
@@ -794,3 +797,260 @@ func (d *DefaultNode) GetContainerStatus(ctx context.Context) clabruntime.Contai
...
+	)
+	if err != nil {
+		// Roll back any endpoints already moved to the parking namespace.
+		if len(moved) > 0 {
+			if rbErr := rollbackEndpoints(moved, func(ep clablinks.Endpoint) error {
+				return ep.MoveTo(ctx, d, nil)
```
steiler: Same here, why not call the MoveTo directly...

nodes/default_node.go
```diff
@@ -794,3 +797,260 @@ func (d *DefaultNode) GetContainerStatus(ctx context.Context) clabruntime.Contai
...
+		)
+	}
+	parkingNode := clablinks.NewGenericLinkNode(parkName, parkPath)
+
+	// Reparent endpoints to the parking node so MoveTo knows where to find them.
+	// Stop and Start are separate process invocations, so the in-memory node
+	// reference from Stop does not survive; we restore it here.
+	for _, ep := range d.GetEndpoints() {
+		ep.SetNode(parkingNode)
+	}
```
steiler: on the parking node implementation you could have a discovery or so to discover all the existing interfaces and simply move all of them back into the given Node, I envision.

nodes/default_node.go
```diff
@@ -794,3 +797,260 @@ func (d *DefaultNode) GetContainerStatus(ctx context.Context) clabruntime.Contai
...
+
+func setEndpointsUp(ctx context.Context, endpoints []clablinks.Endpoint) error {
+	for _, ep := range endpoints {
+		if err := ep.SetUp(ctx); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
steiler: This is also a pretty useless use the loop where this is called right now. In my view thats much cleaner.

links/generic_link_node.go
```diff
@@ -9,9 +9,27 @@ import (
 )
 
 type GenericLinkNode struct {
-	shortname string
-	endpoints []Endpoint
-	nspath    string
+	shortname    string
+	endpoints    []Endpoint
+	nspath       string
+	endpointType LinkEndpointType
```
steiler: What is this supposed to be?
You can throw all kinds if endpoint types into the endpoints slice... Whatever that is, it is bad.

links/parking_node.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		moved = append(moved, ep)
+	}
+
+	if err := p.RepointSymlink(); err != nil {
+		return fmt.Errorf("failed to repoint symlink for %q: %w", p.containerName, err)
+	}
+
+	return nil
+}
```
steiler: why delegate this sort of operation to the parkingNode ... to me that does not make much sense ...
execute the loop and moveTo from within the source and give the parking as a dst... done

links/parking_node.go
```diff
@@ -0,0 +1,110 @@
...
+		node:          NewGenericLinkNode(nsName, nsPath),
+		containerName: containerName,
+	}, nil
+}
+
+func (p *ParkingNode) NSPath() string {
+	return p.node.nspath
+}
+
+func (p *ParkingNode) RepointSymlink() error {
```
steiler, line 30: this is rather a private function ... make it lower case ... check the other functions as well

links/parking_node.go
kaelemc: u mean to move it into DefaultNode? sure.

FloSch62:
@kaelemc and @steiler I tried to drive this further. Feel free to revert the last commit if you had other plans or if its not satisfying to you

nodes/vr_node.go
```diff
@@ -141,3 +145,104 @@ func (n *VRNode) SaveConfig(_ context.Context) (*SaveConfigResult, error) {
...
+	if tag == "" {
+		return "", false
+	}
+
+	if !imageTagRE.MatchString(tag) {
+		return "", false
+	}
+
+	return tag, true
+}
```
kaelemc, line 252: Not really sure if we want all this in here. I feel maybe it can be something that lives in the vrnetlab side or something, but it would mean people have to build new images.

sacckth:
stalled or what? @steiler @kaelemc @FloSch62

kaelemc:
Yeah, stalled from my side. Discussed some stuff with Markus which I’ll try implement soon.

FloSch62:
I tried to rework this so parking and endpoint ownership are handled in the node lifecycle, with the parking netns treated like a regular node and the same MoveTo flow used in both directions. So far I only tested this with srl; before I spend more time, are there any concerns with this direction @steiler @kaelemc ?

hellt:
lets fucking go

PR #3062, opened 2026-02-18
Add exposed ports to events stream

FloSch62:
This adds the exposed ports of a container to the events stream.


hellt:
we can change the runner of the vxlan tests to ubuntu-22.04 to fix the issue with the iproute2 being not compatible until they fix it

hellt:
@dependabot rebase

I guess it won't do it :D

PR #3066, opened 2026-02-20
Add Spirent STC kind

kaelemc:
Adds the Spirent containerized test center

The version I am testing is 5.50.2906, and for some reason uses cgroups v1, which all major distros of today have gotten rid of (for cgroups v2).

Because of this, there is this mess of having to create a bunch of fake cgroup v1 paths else the test center won't boot correctly and you can't use the interfaces.

I also made a decision to force `portX` interface naming, the STC does this thing where upon boot it renames all the interfaces (decrementing them by one) as the management interface is `admin0`.

For this reason I believe forcing the `portX` naming reduces 'cognitive load' :D for the user as:

1. `portX` matches what they see in the spirent GUI (see below screenshot)
2. If we allow `ethX` and the decrementing occurs, it can cause some confusion in troubleshooting usecases, ie. pcap, looking at topoviewer etc. 

CleanShot 2026-02-21 at 04 34 11@2x

### DIFF
--- a/nodes/spirent_stc/stc.go
+++ b/nodes/spirent_stc/stc.go
@@ -76,9 +76,13 @@ func (n *spirentStc) Init(cfg *clabtypes.NodeConfig, opts ...clabnodes.NodeOptio
 
 	// STC uses cgroupv1, mount the fake dirs
 	cgroupDir := path.Join(n.Cfg.LabDir, cgroupV1Dir)
-	n.Cfg.Binds = append(n.Cfg.Binds,
+	n.Cfg.Binds = append(
+		n.Cfg.Binds,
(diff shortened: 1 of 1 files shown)


kaelemc:
Awaiting testers, once all good then will be ready to review.

sacckth:
Windows! yuck!

kaelemc:
> Windows! yuck!

go away.

Just so you don't judge me, it's RDP from a MacBook I had to sell all my internal organs to afford.

kaelemc:
think we will have to find some testers for this out in the wild. It works for me :)

hellt:
lets yolo and find stuff in prod

PR #3067, opened 2026-02-21
Fix typos and curl command in manual installation docs

earies:
The manual installation section had two issues:

- "linux distributive" corrected to "Linux distribution"
- The `curl` command used to determine the latest release version was
  scraping GitHub's HTML page without following redirects (`-L`), causing
  `LATEST` to always be empty. Replaced both occurrences (manual package
  installation and manual installation sections) with a redirect-based
  approach that is simpler and more reliable:

  ```bash
  LATEST=$(curl -sLo /dev/null -w '%{url_effective}' \
         https://github.com/srl-labs/containerlab/releases/latest | sed 's|.*/v||')
  ```

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -171,7 +171,7 @@ example:
 
 ```bash
 # manually install latest release with package managers
-LATEST=$(curl -s https://github.com/srl-labs/containerlab/releases/latest | sed -e 's/.*tag\/v\(.*\)\".*/\1/')
+LATEST=$(curl -sLo /dev/null -w '%{url_effective}' https://github.com/srl-labs/containerlab/releases/latest | sed 's|.*/v||')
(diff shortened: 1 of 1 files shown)


earies:
I'd also like to point out that since the installation script and `version upgrade` approach started to depend on pkg formats and distro discovery, this has broken such features for distros that fall outside of what is supported.

My 2c is that distro packages are complementary to what should be a standalone agnostic approach again.  We're really just talking about a single binary and examples currently.

hellt:
Thanks @earies, good catch!

The reason we went with packages is because gh's api gets throttled unexpectedly and this prevents CI runs from running reliably. With packages it has proven to be more resilient and stable, hence the installation method switched to it.

Issue #3056, opened 2026-02-13
vxlan tests failing because of ubuntu kernel upgrade


FloSch62:
See: https://github.com/actions/runner-images/issues/13669

hellt:
the workaround has been added in #3069

PR #3069, opened 2026-02-22
Add temporary workaround for #3056

FloSch62:
Lets use a workaround for the test until the iproute2 is fixed in the runner

### DIFF
--- a/.github/workflows/build-containerlab.yml
+++ b/.github/workflows/build-containerlab.yml
@@ -9,7 +9,7 @@ name: build-clab
 
 jobs:
   build-clab:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


.github/workflows/build-containerlab.yml
```diff
@@ -9,7 +9,7 @@ name: build-clab
 
 jobs:
   build-clab:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
```
hellt: I wouldn't want the build step to be pinned to 22.04 because of the vxlan issue.
Maybe instead lets follow their advice and install iproute from source

https://github.com/actions/runner-images/issues/13669#issuecomment-3915348281 which will be only in the vxlan test job and hopefully later on they will provide a newer package so that this workaround could be removed

FloSch62: Implemented the workaround instead.

PR #3091, opened 2026-03-05
Add ceos-copy-to-flash to .clab.yml schema

rrajpuro:
Introduce a new schema entry for `ceos-copy-to-flash`, allowing users to specify a list of CEOS-specific configuration files to be copied to the flash directory on startup.

NOTE: Dont see srl-agents property in the latest documentation


hellt:
thanks @rrajpuro 
yes, the srl-agents and mysockets extra opts are not in use anymore

PR #3096, opened 2026-03-07
Add extras processing for node configurations

### LINKED ISSUE #3094: Magic variable `__clabNodeName__` not expanded in `extras` paths (e.g. `topology.kinds.<kind>.extras`)
**Summary**
Magic variable expansion (e.g. `__clabNodeName__`) works for some fields (like `startup-config` and `exec`), but appears not to be applied to `extras` fields such as `extras.ceos-copy-to-flash`. This makes it hard to keep topologies DRY when `extras` contains per-node file paths.

**Example topology snippet**
```yaml
topology:
  defaults:
    startup-config: configs/__clabNodeName__.cfg

  kinds:
    arista_ceos:
      image: ceos:4.35.0F
      enforce-startup-config: true
      extras:
        ceos-copy-to-flash:
          - ceos-configs/__clabNodeName__/ceos-config
```

**Observed behavior**
Deploy fails because the path is used literally (not expanded):
`.../ceos-configs/__clabNodeName__/ceos-config: no such file or directory`

**Expected behavior**
`__clabNodeName__` should expand to each node short name, similarly to `startup-config` and `exec`, e.g.:
`ceos-configs/ce1/ceos-config`

**Question / request**
Why did we not implement in containerlab:
1) A generic pre-processor phase that walks the parsed topology and expands magic vars for all string fields
2) At least expand magic vars for known `extras` string slices (e.g. `ceos-copy-to-flash`, etc.)?

hellt:
This commit introduces a new function, `processNodeExtras`, to handle magic variable replacements in node extras, specifically for `ceos-copy-to-flash` and `srl-agents`. Additionally, a new test, `TestExtrasInit`, has been added to validate the functionality using a sample topology file. The topology file has also been updated to include the new extras structure for testing purposes.

fix #3094

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ build-with-cover:
 
 build-debug:
 	mkdir -p $(BIN_DIR)
-	go build -o $(BINARY) -gcflags=all="-N -l" -race -cover main.go
+	CGO_ENABLED=1 go build -o $(BINARY) -gcflags=all="-N -l" -race -cover main.go
 	sudo chown root:root $(BINARY)
 	sudo chmod 4755 $(BINARY)
 
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/opencontainers/runtime-spec v1.2.1
 	github.com/pmorjan/kmod v1.1.1
-	github.com/scrapli/scrapligo v1.3.3
+	github.com/scrapli/scrapligo v1.3.2
 	github.com/scrapli/scrapligocfg v1.0.0
 	github.com/spf13/cobra v1.10.2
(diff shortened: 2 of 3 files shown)


rrajpuro:
@hellt why not parse the YAML and replace all string fields. If variables are reserved and magic (i.e., globally defined placeholders), traversing the parsed YAML tree instead of using path-specific replacement because:
- Consistency: Ensures all instances of __VAR__ are replaced, avoiding missed substitutions. 
- Maintainability: No need to manually track or update specific paths when structure changes. 
- Simplicity: One uniform rule for replacement across the entire config. 
- Safety: Operates on scalar strings only, preserving structure and avoiding injection.

hellt:
because inputs to the replacer function are unknown until the topology is parsed and expanded. You can try to implement your approach

PR #3128, opened 2026-03-24
Fix startup-config cross-pollution in XRd, cEOS, VyOS, and c8000 nodes

micyen:
Problem
Nodes with global cfgTemplate variables (XRd, cEOS, VyOS, c8000) were overwriting the shared template when startup-config was provided, causing configuration cross-pollution between nodes in the same lab. For example, if xrd1 had a startup-config provided, it would affect xrd2's auto provisioned config (without startup-config provided).

Root Cause
Package-level cfgTemplate (populated via //go:embed) was being mutated directly: cfgTemplate = string(c) when reading user-provided startup files.

Solution
Use local template copies in config generation functions:

currentCfgTemplate := cfgTemplate (copy global)
If startup-config provided, assign to local currentCfgTemplate
Pass local to GenerateConfig()
This prevents global state mutation while preserving per-node customization.

Changes
XRd: Fixed in createXRDFiles() with local copy and comment
cEOS/VyOS/c8000: Applied same pattern in respective config functions

Testing
XRd: Manually verified with topology containing xrd1 (with startup-config) and xrd2 (without); configs isolated correctly
Other nodes: Pattern applied but not tested

### DIFF
--- a/nodes/c8000/c8000.go
+++ b/nodes/c8000/c8000.go
@@ -107,15 +107,17 @@ func (n *c8000) create8000Files(_ context.Context) error {
 	nodeCfg.MgmtIPv6Gateway = n.Runtime.Mgmt().IPv6Gw
 
 	// use startup config file provided by a user
+	// make copy of template to prevent provided startup config from mutating shared package template value
+	currentCfgTemplate := cfgTemplate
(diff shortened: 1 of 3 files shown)


hellt:
Good catch. Thank you

PR #3133, opened 2026-03-24
fix: prevent nil dereference panics in destroy, netconf, and sros

Aprazor:
## Summary

Three nil/bounds-check fixes that prevent panics:

- `core/destroy.go`: `cont.Names[0]` was accessed without checking `len(cont.Names)`. Orphaned or stopped containers under Podman can have empty names, causing `clab destroy` to panic rather than skip or warn.
- `netconf/netconf.go`: `r.Input` and `r.Failed` were accessed before checking the operation error `e`. If `operation(d)` returns a nil `r` along with a transport error (connection reset, timeout), this panics. Fixed the check order so the error is returned first.
- `nodes/sros/sros.go`: `calcComponentFqdn` sliced `n.Cfg.Fqdn[:fqdnDotIndex]` where `fqdnDotIndex` comes from `strings.Index`, which returns `-1` when `.` is not found. A short hostname without a domain suffix causes an index out of range panic during distributed SR-SIM deployment.

## Testing

- `go vet ./core/... ./netconf/... ./nodes/sros/...` — clean
- `go test -race ./core/... ./nodes/sros/...` — all pass
- `netconf/` has no unit tests in the project; fix verified by code inspection (error check moved before nil dereference)

### DIFF
--- a/core/destroy.go
+++ b/core/destroy.go
@@ -367,6 +367,10 @@ func (c *CLab) deleteContainersDirect(
 	containers []clabruntime.GenericContainer,
 ) {
 	for _, cont := range containers {
+		if len(cont.Names) == 0 {
+			log.Warnf("skipping container %s with no names", cont.ID)
+			continue
+		}
 		name := cont.Names[0]
 		log.Infof("Removing container: %s", name)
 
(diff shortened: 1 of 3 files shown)


hellt:
merging to `dev` to run sros tests

PR #3135, opened 2026-03-25
Feature/annotations endpoint for UI persistance

ldrozdz93:
## Notes

This is a draft PR to gather feedback from maintainers. A companion React-based UI that consumes this API exists but is not part of this PR — the endpoints are useful independently for any custom frontend or tooling.

Code will be thoroughly reviewed and improved before the PR is marked ready for review.

## Summary

Add `GET/POST /api/v1/annotations` endpoints to the `clab graph` HTTP server that persist node position data to `.clab-topo-editor.json` next to the topology file.

This enables interactive graph UI frontends to save and restore node positions across browser sessions — for example after dragging nodes into a preferred layout, positions survive a page refresh.

The file format is intentionally compatible with the [VSCode containerlab extension](https://github.com/srl-labs/vscode-containerlab)'s `topoViewer` position storage:

```json
{
  "nodeAnnotations": [
    { "id": "srl1", "position": { "x": 100, "y": 200 } }
  ]
}
```

## API

- `GET /api/v1/annotations` — returns saved positions, or `{"nodeAnnotations":[]}` if the file doesn't exist
- `POST /api/v1/annotations` — writes the request body JSON to `.clab-topo-editor.json`

### DIFF
--- a/core/graph.go
+++ b/core/graph.go
@@ -7,6 +7,7 @@ package core
 import (
 	"context"
 	"embed"
+	"encoding/json"
 	"fmt"
 	"html/template"
 	"io"
@@ -15,6 +16,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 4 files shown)


ldrozdz93:
@hellt could you please take a look? I'd like to know if the community would be interested in this change. If yes, then I would put more work into it.

hellt:
I think @FloSch62 planned to have a standalone web app (outside of the vsc extension) that seems to be doing what you have in mind. I will let Florian to comment further when he is back from his business trip

core/graph.go
```diff
@@ -333,6 +335,45 @@ func (c *CLab) ServeTopoGraph(tmpl, staticDir, srv string, topoD TopoData) error
...
+
+				return
+			}
+
+			w.WriteHeader(http.StatusOK)
+
+		default:
+			http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		}
+	})
```
hellt, line 375: clab doesn't have an embedded api server; the rest api for clab is in https://github.com/srl-labs/clab-api-server

ldrozdz93: @hellt thanks, I'll follow this feedback once I have a confirmation that this direction is acceptable.

ldrozdz93:
Thanks @hellt. In this case, I'll wait for @FloSch62 to get his feedback.

FloSch62:
Hi @ldrozdz93, 

I think I would not like to add api endpoints to the graph command as it's only used to serve the html based graph. I am currently working on a standalone version for the vscode ui which will use the clab-api-server. So I think this is the way to go for any frontend.

hellt:
I suppose the outcome is that this best to be handled outside of clab's core

PR #3136, opened 2026-03-25
Fix 3089 Allow credentials to be set via topology configuration

### LINKED ISSUE #2009: If SR OS node startup config contains a custom admin password, the vrnetlab bootstrap fails as it uses the default `admin:admin` credentials
Hello, 

I have a problem with the health status in SROS deployments with the containerlab. The `containerlab deploy` command hangs until the container health status turns `healthy`. However, the containerlab container hangs infinitely. It's not turning healthy. If I use `--skip-post-deploy`, it works but I cannot use it because of the project requirements.

I discovered that is coming from my configuration file. I am setting up a custom admin password in the configuration file. If I change this custom admin password to admin, then the container health status turns to `healthy`, and the `containerlab deploy` command successfully ends.

I was looking at how vrnetlab and containerlab use the username and the password for the health check. I can see vrnetlab library modify the `/health` file with the [update_health](https://github.com/vrnetlab/vrnetlab/blob/master/common/vrnetlab.py#L403) function. But the `username` or `password` definition is not used in this [VR class](https://github.com/vrnetlab/vrnetlab/blob/master/common/vrnetlab.py#L393).

I also know containerlab is only responsible for checking the `health` file content. It's using this file to define container health status.

The `--username` and `--password` are used in [launch.py](https://github.com/vrnetlab/vrnetlab/blob/master/sros/docker/launch.py#L382) in the [vrnetlab/sros/](https://github.com/vrnetlab/vrnetlab/tree/master/sros/docker) directory. I tried to add these variables in the containerlab container entry point like below. However, it is not working:

```bash
....
topology:
  nodes:
    sros1:
        kind: nokia_sros
        mgmt-ipv4: 172.100.101.21
        entrypoint: /launch.py --trace --connection-mode tc --hostname sros1 --username admin --password mY_Custom_Pass --variant sr-1
        startup-config: sros1.cfg
```

I also can see that vrnetlab does health checks with these commands. I tried this command in the containerlab container. But, I cannot take any response. At the same time, I can open a terminal with telnet in the containerlab container.

```bash
socat TCP-LISTEN:22,fork TCP:127.0.0.1:22
socat UDP-LISTEN:161,fork UDP:127.0.0.1:161
socat TCP-LISTEN:830,fork TCP:127.0.0.1:830
socat TCP-LISTEN:80,fork TCP:127.0.0.1:80
socat TCP-LISTEN:443,fork TCP:127.0.0.1:443
```

I just want to know how can I fix the containerlab SROS container's health status. I am using the Nokia SROS 24.03 in the Ubuntu-22.04. Containerlab version 0.50.0. But I tried with version 0.54.2. In both cases, vr-sros health status is not turning healthy with a custom password.

### LINKED ISSUE #2146: Allow for a custom default admin password to be set via topology attributes
Some lab environments restrict the users to have nodes with the default credentials shipped with the node (admin:admin, etc).

We must have a way to set the default password via the topology setting to mitigate this. Maybe an env var set for a node/kind/default that would set the credentials during the node registration.

### LINKED ISSUE #3089: Allow configuring credentials with ENV
# Allow configuring credentials for `save` command via env variables (with fallback to default)

## Description

We are using Containerlab to simulate our full production configuration with vr based devices, including user management. As part of this process, we modify or remove the default user that Containerlab relies on for device access.

After removing or changing the default user, the `containerlab save` command no longer works, and startup configurations are not saved anymore.

## Problem

The `save` command appears to rely on the default credentials configured by Containerlab. When those credentials are modified or removed (to better reflect production setups), the command fails and configurations are no longer persisted.

## Proposed Enhancement

Allow defining credentials via the node `env` system so that Containerlab can use custom authentication when interacting with nodes.

Reference documentation:  
https://containerlab.dev/manual/nodes/#env

Example (conceptual):

```yaml
nodes:
  r1:
    kind: <kind>
    env:
      CLAB_USERNAME: myuser
      CLAB_PASSWORD: mypassword

```

## Expected Behavior
- If custom credentials are defined via environment variables, Containerlab should use them for operations such as save.
- If no custom credentials are defined, Containerlab should fall back to the default username/password behavior.
- This would allow users to remove or modify the default user without breaking lifecycle operations.

## Benefits
- Enables production-like authentication models in lab environments.
- Removes dependency on default credentials.
- Improves flexibility and security alignment with real-world deployments.

Is this something that could be supported?
Alternatively, is there a recommended approach to handle this use case?

antoinekh:
Fixes #3089
Fixes #2146 
Fixes #2009

## Summary

Allow credentials to be set via topology configuration. Adds `username` and `password` fields at any level of the topology hierarchy, with a well-defined resolution order.

## New topology fields

Credentials can now be configured at the `defaults`, `kinds`, or `nodes` level:

```yaml
topology:
  defaults:
    username: admin         # applies to all nodes
    password: mypassword
  kinds:
    nokia_srlinux:
      username: srl-admin   # overrides defaults for SRL nodes
      password: srl-pass
  nodes:
    router1:
      username: router1-user  # overrides everything for this node
      password: router1-pass
```

## Credential resolution order

Most specific to least specific:

`node` → `group` → `kind` → `defaults` → kind's hardcoded default

### Core types (`types/node_definition.go`, `types/topology.go`, `types/types.go`)

- Added `Username`/`Password` fields to `NodeDefinition` (YAML-serializable)
- Added `GetNodeUsername()`/`GetNodePassword()` methods to `Topology` with hierarchy resolution
- Added `Username`/`Password` to `NodeConfig` — `Password` is excluded from JSON marshaling for security

### Core config (`core/config.go`)

- `createNodeCfg()` now populates credentials from topology, falling back to the kind's hardcoded registry defaults when not set

### Node kinds

- All 40+ node kinds updated to use `n.Cfg.Username`/`n.Cfg.Password` instead of the package-level `defaultCredentials` variable

### Inventory generation (`core/inventory.go`)

- Ansible and Nornir inventories now use per-node resolved credentials

### Schema & docs

- Updated `clab.schema.json` and `docs/manual/nodes.md`

### Tests (`types/topology_test.go`)

- Added `TestGetNodeCredentials` covering all hierarchy levels


antoinekh:
Used in lab for a couple of days, no issues

sacckth:
Great contribution. I'll let our benevolent containerlab leader to comment but there are some things left.
I found an issue with a potential rabbit hole. There are certain NOSes that don't have explicit initial configuration that sets the password and username (e.g. SRLinux). For this case, we need to add templates to render the username and password accordingly.

Similarly, in some other NOSes, we explicitly set the password in the default configuration (e.g. SROS SR-SIM). 

A compromise would be to set the password/user explicitly  in the startup-config which I believe most of the people did already.  

We should probably add  a few Robot tests that perform checks using SSH using the "default" creds. I suggest to cover a few usecases with this new variable.

antoinekh:
@hellt pipeline failed for GitHub ratelimit error 429. Nothing related to this code

hellt:
Thanks. Will check it out when time permits. 
In the meantime others can check out it from the branch - that would help to get some real world converage

hellt:
Hi @antoinekh would you mind trying this adaptation of this PR where the credentials are put under the `credentials` container:

```
credentials:
  username: foo
  password: bar
```

hope it still works :D

hellt:
I assume it is

antoinekh:
Sorry I was ouf for few days. I will test this afternoon

antoinekh:
@hellt tests are ok on my side, thanks for the update on the PR and the merged :)

PR #3145, opened 2026-03-30
feat(deploy): add --export-rendered flag to save expanded topology YAML

### LINKED ISSUE #3013: Add an option to save the topology file generated from the topo template and provided vars file
### Discussed in https://github.com/srl-labs/containerlab/discussions/3009

<div type='discussions-op-text'>

<sup>Originally posted by **ondrej-fabry** January 26, 2026</sup>
I’m using a templated topology + vars workflow and I’m trying to make it more robust/ergonomic.

## Current setup

* Topology template: `lab0.clab.gotmpl`
* Vars file: `lab0.clab_vars.yml`

Today, I run containerlab commands directly against both files:

```sh
containerlab deploy   -t lab0.clab.gotmpl --vars lab0.clab_vars.yml
containerlab destroy  -t lab0.clab.gotmpl --vars lab0.clab_vars.yml
```

This works, but it makes lifecycle operations depend on **two separate inputs staying in sync** (and being available) for as long as the lab exists. That’s fragile if the lab is “long-lived” on a user machine and the repo moves or changes.

## Desired workflow

Use template + vars **only at creation time**, then store the rendered topology in the user’s lab directory and manage the lab from that single rendered file:

```sh
# once
containerlab generate \
  -t lab0.clab.gotmpl \
  --vars lab0.clab_vars.yml \
  --file ~/labs/lab0.clab.yml \
  --name lab0

# repeat
containerlab deploy  -t ~/labs/lab0.clab.yml
containerlab destroy -t ~/labs/lab0.clab.yml
```

## Questions

1. Is there already a recommended or idiomatic way to do this in containerlab (render once → manage from rendered file)?
2. If not, is this approach sound, or does containerlab intentionally require the original template + vars for correct lifecycle management?
3. Are there any pitfalls to be aware of (e.g. lab name/ID derivation, labdir semantics, inventory/state tracking) when managing from a rendered topology?

Any suggestions for a best-practice workflow are welcome.
</div>

Aprazor:
## Summary

When using topology templates (`*.clab.gotmpl`) with variables files, the final rendered YAML only exists in memory and is lost after deploy. This makes it hard to debug template issues or version-control the rendered output.

Adds `--export-rendered <path>` flag to the `deploy` command that saves the fully expanded YAML (after template execution and env var substitution) to the specified file path before unmarshaling.

```bash
clab deploy -t lab.clab.gotmpl --export-rendered /tmp/lab-rendered.yaml
```

The saved file contains the final YAML exactly as containerlab would parse it — all template logic resolved and all `${ENV_VARS}` substituted.

Closes #3013

## Testing

- `go vet ./cmd/... ./core/...` — clean
- `go test -race ./cmd/... ./core/...` — all pass
- The flag is a pure additive feature with no impact on existing behavior when omitted


cmd/options.go
```diff
@@ -212,6 +212,13 @@ func (o *Options) ToClabDestroyOptions() []clabcore.DestroyOption {
 		}
 	}
 
+	if o.Global.VarsFile != "" {
+		destroyOptions = append(
+			destroyOptions,
+			clabcore.WithDestroyVarsFile(o.Global.VarsFile),
+		)
+	}
```
amtelekom, line 220: Why copy the VarsFile to Destroy Options here? The globally set vars file was already used via `c.TopoPaths.VarsFilenameAbsPath` in the destroy operation, so to keep a copy of this in the destroy options (with always the same value as the global option) seems like a redundant copy of the same parameter, but allowing that a relative path might slip through, where the code previously could rely on always receiving an absolute path?

hellt: good call. I will amend that later

PR #3157, opened 2026-04-06
fix(srl): accept ixr-x4 as alias for ixr-x4-d topology type. Fixes #3149

### LINKED ISSUE #3149: IXR-X4 Topology Accepted Values vs Schema Mismatch
It appears there is a mismatch in what the topology JSON schema says is a valid value for the X4 and what containerlab accepts:

JSON Schema:
`ixr-x4`

Containerlab Acceptable Value:
`ixr-x4-d`

sacckth:
Align runtime srlTypes with clab.schema.json (issue #3149) and allow ixr-x4-d in schema validation for users who use the documented name.


hellt (dismissed):

hellt:
will need to be updated in srlinux.md as well I think

hellt (approved):

PR #3158, opened 2026-04-07
Add link IP support to VyOS and cEOS

evilhamsterman:
As the title says, this adds support for the Link IPs feature added in 0.70 to the VyOS and cEOS kinds


hellt:
Thanks @evilhamsterman

PR #3167, opened 2026-04-21
feat(cmd): add --name flag to tools vxlan delete for exact-match deletion

Rucdev:
## Summary

- Add a new `--name | -n` flag to `containerlab tools vxlan delete` to delete a single VxLAN interface by exact name.
- `--name` and `--prefix` are mutually exclusive; at least one of them is required.
- When `--name` refers to a non-existent interface, or an interface that is not a VxLAN, the command now returns an error instead of silently doing nothing.
- `--prefix` behavior is unchanged for backward compatibility.

## Motivation

The existing `--prefix` flag uses `strings.HasPrefix` for matching. When two VxLAN interfaces share a common naming stem (for example `vx-srl2-e1-1` and `vx-srl2-e1-10`), deleting one with `--prefix vx-srl2-e1-1` also removes `vx-srl2-e1-10`, which is surprising and has caused issues downstream.

clabernetes uses this command as a safety net before creating VxLAN interfaces. See srl-labs/clabernetes#252 for the concrete failure mode this addresses.

With `--name`, callers that know the exact interface name (like clabernetes) can request a precise deletion without affecting siblings.

## Test plan

- [x] `go build ./...` passes
- [x] Manually verified with dummy VxLAN interfaces:
  - `--name vx-srl2-e1-1` deletes only that interface, leaving `vx-srl2-e1-10` intact
  - `--name <non-existent>` returns an error
  - `--name <non-vxlan interface>` returns an error
  - `--name` and `--prefix` together → cobra rejects
  - Neither flag → cobra rejects
  - `--prefix` alone still works as before (including the prefix-overlap behavior, unchanged)

### DIFF
--- a/cmd/options.go
+++ b/cmd/options.go
@@ -474,6 +474,7 @@ type ToolsVxlanOptions struct {
 	Remote         string
 	ParentDevice   string
 	DeletionPrefix string
+	DeletionName   string
 }
 
 type ToolsSnapshotOptions struct {
--- a/cmd/tools_vxlan.go
+++ b/cmd/tools_vxlan.go
@@ -118,6 +118,15 @@ func vxlanCmd(o *Options) (*cobra.Command, error) { //nolint: funlen
 		o.ToolsVxlan.DeletionPrefix,
 		"delete all containerlab created VxLAN interfaces which start with this prefix",
 	)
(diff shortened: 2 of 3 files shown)


hellt:
thanks @Rucdev

PR #3169, opened 2026-04-23
Enable topology template composition from multiple files

amtelekom:
# Summary
Allows defining Template Variables in multiple yaml files, and rendering additional .gotmpl files from a topology gotmpl file.

# Use-Case
We're specifying a test environment for an Orchestrator that needs to accomodate a few combinations of different blocks of network devices (think combinations of PE Setup A, B, C with CE setup 1, 2, 3, with a few more axis thrown in).
Until now, we wrote one big topology template with large `if`-Blocks, but handling this huge file has become very unwieldy and also leads to bad reuse of very similar, but sightly different definitions, because editors refrain from defining templates inside the template file since it makes navigation even harder.
Also, we'd like to share some Template Variables between some LABs with only slight differences, but currently each one needs its complete definition (or we need to do ENV-Vars, but to keep these consistent you effectively neet to rely on some external tooling).

# Implemented additions
- The `--vars` option can now be specified multiple times, contents of the referenced files will be merged with later files overwriting previous keys
- If the vars are autodetected based on the topology template filename, `template.clab_vars.*.[yaml|yml|json]` will also be loaded
- If there is a directory called `clab_templates` next to the topology template, all `*.gotmpl` inside of it are also loaded and are available during template execution via the `template` built-in of go's template engine

# Impact

## Binary size
Since this is done without additional dependencies a la gomplate, the impact is minimal; current main branch compiles to a 28,766,392 byte binary in my setup; this branch to 28,786,872 bytes (+20kb / less than+1‰)

## Backward compatibility
No existing options are changed, just added to.
Theoretically, if users had a file that was previously not picked up by the automatic discovery (either an additional variables file or a gotmpl in a clab_templates subdirectory) and those contained syntactical errors or overwrite options from the main variables file, it could change behavior, but I'd like to think this unlikely.
If a syntax error due to a newly parsed file happens, the error message points the user towards the included file, e.g.,

> Failed to read topology file: template: invalid_subtempl.gotmpl:1: undefined variable "$TEST"

for a subtemplate in the clab_templates directory, or 

> Failed to read topology file: variables file 'my_topo.clab_vars.fail.yaml': yaml: line 3: found unexpected end of
  stream.


hellt:
thanks @amtelekom 
can you merge main into your branch so that the changes from https://github.com/srl-labs/containerlab/pull/3171 are soked in?

amtelekom:
Done. I'm not sure why the docs check fails with 403 errors on remotely hosted images, since I only added text blocks to hint at existing and new functionality of the templating system.

hellt:
thnx; the docs 403 errors are expected. Nothing to worry about

PR #3186, opened 2026-05-07
container runtime lifecycle and netavark support

orrious:
Add first-class topology support for container runtime options needed by systemd-style nodes, including privileged mode, cgroup namespace mode, pid mode, tmpfs mounts, and security options. Wire those fields through Docker and Podman container creation while preserving existing behavior when they are omitted.

Extend lifecycle ordering so start honors stage wait-for dependencies and external containers participate in dependency ordering. This lets topologies gate dependent nodes on prerequisite lifecycle states without lab-side startup scripts.

Fix management bridge detection when using Podman/netavark through the Docker compatible API. Containerlab now prefers netavark's network_interface metadata when available, falls back to Podman's libpod network metadata when Docker compatibility strips that field, and retains Docker's existing bridge-name fallbacks.

Validation:
- go test ./runtime/docker -run "TestMgmtBridgeNameFromInspect|TestNetavarkNetworkInterface"
- go test ./core -run "TestLifecycle"
- go test ./types ./runtime/docker
- representative topology deploy with netavark workaround disabled


hellt:
Let's split this in three PRs as they are all not related

PR #3190, opened 2026-05-15
Make concurrent deploys on the same host more reliable

torbbang:
Running two `clab deploy`s (or a deploy and a destroy) at the same time on one host tends to fail in the following ways:

- `/etc/hosts` can end up with a half-written block, or lose its baseline `localhost` lines.
- The second deploy's `docker network create` races with the first and errors out with "network with name clab already exists".
- A destroy that removes a container while another deploy is listing containers makes that deploy abort with "container cannot be found".

This serializes the `/etc/hosts` edits with a file lock, proceeds on "already exists" 409 and skips containers that vanish mid-enumeration. Each one has a regression test.

### DIFF
--- a/core/hostsfile.go
+++ b/core/hostsfile.go
@@ -8,7 +8,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
+	"sync"
+
+	"golang.org/x/sys/unix"
 
 	clabconstants "github.com/srl-labs/containerlab/constants"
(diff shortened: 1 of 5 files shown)


torbbang:
PR that fixes the failing smoke tests #3192

hellt:
@torbbang whenever you have time merge in the `main` in this one so that we can re-run the pipeline
thanks!

hellt:
thanks @torbbang

PR #3192, opened 2026-05-15
Switch to HTTPS in api-server smoke tests

torbbang:
clab-api-server v0.3.0+ defaults TLS_ENABLE to true and serves HTTPS with a self-signed cert. The smoke test was still curling plain HTTP and getting "Client sent an HTTP request to an HTTPS server."


FloSch62 (approved):

hellt:
/cc @FloSch62 
seems reasonable, if that change was intentional on the clab-api side

FloSch62:
Yes, default is now https, sadly I missed to change this, thank you @torbbang for this! 
@hellt from my side you can merge it

PR #3198, opened 2026-05-20
add srsim hw schema sidecar

FloSch62:
This adds a full compability matrix for srsim components, generated by https://github.com/FloSch62/srsim-hw-schema with the input from https://documentation.nokia.com/sr/26-3/7x50-shared/srsim-installation-setup/appendices.htm 

I know its a massive 8k lines sidecar, but if we want default yaml validations I fear it only works with the massive if`s ( but this is anyhow auto generated ) But because of that the idea is just to have the ref in current schema and just load the sidecar schema.

### DIFF
--- a/schemas/clab.schema.json
+++ b/schemas/clab.schema.json
@@ -323,122 +323,12 @@
                 "components": {
                     "type": "array",
                     "items": {
-                        "type": "object",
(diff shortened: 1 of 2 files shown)


FloSch62:
Lets not do it like that.

PR #3199, opened 2026-05-21
Modify Schema to make endpoints mandatory and to add the debug field

alexandrehassan:


schemas/clab.schema.json
alexandrehassan, line 742: Looking into the jsonschema spec better endpoints should probably be treated as a tuple not a list [jsonschema tuple docs](https://json-schema.org/understanding-json-schema/reference/array#tupleValidation)

This would would save having to write min and max lengths.

The schema would become:
```json
"endpoints": {
    "type": "array",
    "description": "endpoints list",
    "markdownDescription": "[endpoints](http://localhost:8000/manual/topo-def-file/#links) list",
    "uniqueItems": true,
    "prefixItems": [
        {
            "type": "string",
            "pattern": "^\\S+:\\S+$"
        },
        {
            "type": "string",
            "pattern": "^\\S+:\\S+$"
        }
    ]
},
```

We generate a topology validator and as a more concrete example its the difference between

```python
class LinkConfigShort(BaseModel):
    endpoints: Annotated[
        list[Endpoint] | None, Field(description="endpoints list", min_length=2, max_length=2)
    ] = None
    ...
```

and

```python
class LinkConfigShort(BaseModel):
    endpoints: Annotated[tuple[Endpoint, Endpoint], Field(description="endpoints list")]
```

hellt: I am ok with the tuple as well. Do you want to change this PR?

alexandrehassan: Added the change to the PR

PR #3202, opened 2026-05-24
Quote DOT graph name in graph output

fabriciojm:
This fixes DOT graph generation for `clab graph --dot` by quoting the top-level graph name.

Because topology files are normally named with the `.clab.yml` / `.clab.yaml` pattern, the generated graph name normally includes `.clab` in its name. That means the unquoted DOT header was using a graph name containing a dot by default, which can break Graphviz parsing and PNG generation.

This change quotes the graph name before writing the DOT file and adds a small unit test to cover that case.


core/graph_test.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2020 Nokia
```
hellt: thanks @fabriciojm !

would you mind removing the test (as silly as this sounds ;)), it just doesn't test much and I will be fine not having it.
The rest looks good, thanks for providing the workaround.

I must admit, the dot generation took the back seat after we got the vscode extension that provides much more usable graphics

fabriciojm: Done -- happy to contribute!

PR #3206, opened 2026-06-01
Add full GUI Docs

FloSch62:
image


hellt:
let's enable git-lfs so that pngs are not stored in git pack

cmd/options.go
```diff
@@ -15,6 +15,7 @@ const (
 	multiToolImage             = "ghcr.io/srl-labs/network-multitool"
 	defaultTimeout             = 120 * time.Second
 	defaultToolsServerPort     = 8080
```
hellt, line 17: is this one still needed then?

FloSch62: yes thats the default tools port for example for gotty... but to avoid collision api server default moved to 8090

FloSch62:
> let's enable git-lfs so that pngs are not stored in git pack

@hellt I think I added git lfs before the first commit, can you please crosscheck, if this was right?

hellt:
yeah, all good then
looks fine to me, if that is ready I can give it a review and merge if you don't have any pending changes

FloSch62:
> yeah, all good then looks fine to me, if that is ready I can give it a review and merge if you don't have any pending changes

All ok from my side

PR #3207, opened 2026-06-01
Apply

FloSch62:
## Summary

Adds `containerlab apply`, a lifecycle command that applies a topology file to a lab.

If the lab is not running, `apply` behaves like `deploy`. If the lab is already running, it compares the desired topology with the current lab runtime and applies
supported changes without destroying and recreating the whole lab.

The first supported scope is topology reconciliation:

- add nodes
- delete nodes
- add supported links
- delete stale containerlab-owned link endpoints
- start stopped existing nodes and restore parked dataplane interfaces
- restart or recreate existing nodes for the limited node-diff cases supported by the new reconcile plumbing

## How Apply Plans Changes

For a running lab, apply uses:

- runtime containers and containerlab labels
- live network interfaces carrying containerlab ownership markers
- the desired topology file
- the lab state file, `clab-<lab>/.state.clab.yaml`, when available

The state file is written by deploy/apply and is used to compare selected existing node configuration fields. Link reconciliation still uses live runtime
interfaces rather than historical link definitions.

## Main Flow

For a running lab, apply roughly:

1. Discovers lab containers from Docker/Podman labels.
2. Groups distributed node containers by root node label, including SR-SIM components.
3. Resolves desired topology links.
4. Validates that the topology/runtime shape is safe for apply.
5. Loads the previous lab state when available.
6. Plans added/deleted/recreated/started/restarted nodes.
7. Discovers live containerlab-owned interfaces.
8. Plans stale owned endpoints for deletion.
9. Deletes stale endpoints and removed nodes.
10. Creates added or recreated nodes.
11. Creates missing supported links.
12. Restores parked interfaces and starts stopped nodes where needed.
13. Runs post-deploy work for added/recreated nodes.
14. Regenerates lab outputs and writes the updated state file.

## Supported Link Changes

Supported link types include:

- `veth`
- brief links
- `host`
- `mgmt-net`
- `macvlan`
- `vxlan`
- `vxlan-stitch`
- `dummy`
- `bridge`

For special endpoints such as `host` and `mgmt-net`, apply only checks expected interface names so it does not delete unrelated host interfaces.

## Node Lifecycle Behavior

Node kinds can choose how link changes are handled:

- live link changes: Linux, SR Linux, SR OS/SR-SIM
- restart for link changes: cEOS
- recreate for link changes: default/unknown and vrnetlab-backed nodes

Apply also has state-backed node diff plumbing. Container creation fields such as image, type, env, command, entrypoint, runtime, resources, devices, ports, binds,
license, and SR-SIM components can be planned as recreate actions where supported. `exec` changes are planned as restart actions.

This is not a general in-place node mutation engine. Startup config changes and clean artifact changes still require `redeploy` or `deploy --reconfigure`.

## Lab Artifacts

Apply does not delete the lab directory.

Existing node artifacts, generated configs, certificates, licenses, and node directories remain in place unless a normal deploy/recreate path overwrites or reuses
them. Deleting a node removes the runtime container but does not remove that node's lab directory.

Use `redeploy` or `deploy --reconfigure` when a clean artifact/config state is required.

## Output

Apply prints a table summary or plan.

Common rows include:

- `deployed lab`
- `added nodes`
- `deleted nodes`
- `recreated nodes`
- `started nodes`
- `added links`
- `deleted endpoints`
- `restarted nodes`
- `no changes`

Deleted runtime link state is reported as endpoints because apply removes concrete live interfaces.

## Flags

Adds local apply flags:

- `--dry-run`
- `--max-workers`
- `--skip-post-deploy`
- `--export-template`

`--dry-run` prints the planned actions without mutating the lab.

## Limitations

Use `redeploy` or `deploy --reconfigure` for:

- startup config changes on existing nodes
- forcing clean lab artifacts
- link parameter changes where endpoint names stay the same
- MTU changes on existing links
- changing link type while keeping the same endpoint names
- unsafe ownership/namespace/container-creation changes

Known unsupported or unsafe shapes include root namespace links, auto-remove nodes, pre-existing external containers, shared `network-mode: container:<...>` cases,
and runtime state that cannot be matched safely to the desired topology.

## Testing

Adds unit coverage for apply planning, output, link endpoint selection, node link-apply modes, state-backed behavior, distributed component grouping, and ownership
handling.

Adds smoke coverage for:

- deploy fallback through apply
- dry-run plans
- add/delete links
- add/delete nodes
- special link types
- preserving existing node runtime identity for live-capable link changes
- SR-SIM apply scenarios, including stopped-node recovery and adding a components-based node


.cursor/skills/karpathy/SKILL.md
```diff
@@ -0,0 +1,72 @@
+---
+name: karpathy-guidelines
+description: Behavioral guidelines to reduce common LLM coding mistakes. Use when writing, reviewing, or refactoring code to avoid overcomplication, make surgical changes, surface assumptions, and define verifiable success criteria.
```
hellt: @FloSch62 this one is already there, but under the different name (as a rule) https://github.com/srl-labs/containerlab/blob/main/.cursor/rules/karpathy-guidelines.mdc

What is better a rule or a skill?

types/node_definition.go
```diff
@@ -88,6 +88,9 @@ type NodeDefinition struct {
 	// Network aliases
 	Aliases    []string     `yaml:"aliases,omitempty"`
 	Components []*Component `yaml:"components,omitempty"`
+	// how `containerlab apply` handles dataplane link changes for this node:
+	// live, restart or recreate. Overrides the kind's own declaration.
+	LinkApplyMode string `yaml:"link-apply-mode,omitempty"`
```
hellt: instead of a string, lets have a stringed type so that only the allowed values can be passed to it

.cursor/skills/karpathy/SKILL.md
FloSch62: You added it... I dont use cursor... I am not a big fan having skills in the repo... Lets remove it from this pr and keep the rule...

FloSch62: done

types/node_definition.go
FloSch62: done

PR #3218, opened 2026-06-05
Check if modules could be built into kernel.

kaelemc:
For some use case where maybe a custom kernel is required to be used, it might make sense to build the kernel directly into the kernel instead of loading it as needed.

Right now we only support the latter when doing module checks, so if a module is built into kernel we will not catch it and erroneously throw errors if it's not in `/proc/modules`.

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -503,6 +503,15 @@ func (*CLab) loadKernelModules() error {
 		// trying to load the kernel modules.
 		km, err := kmod.New(opts...)
 		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				log.Debugf(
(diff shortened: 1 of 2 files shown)


core/config.go
```diff
@@ -503,16 +503,25 @@ func (*CLab) loadKernelModules() error {
...
 			return nil
 		}
 
 		err = km.Load(m, "", 0)
 		if err != nil {
-			log.Warnf("Unable to load kernel module %q automatically %q", m, err)
-
-			return nil
+			return fmt.Errorf(
+				"kernel module %q is not loaded, not built-in and could not be loaded: %w", m, err,
```
kaelemc, line 523: I made this one an error as it maybe makes sense to fail here. but can revert back to warn upon your wish.

Issue #3037, opened 2026-02-06
[arista_ceos] Change the management interface to `Management1`


mtache:
Using `Management0` in cEOSlab causes some issues that we cannot workaround or fix on the cEOSlab side.

In EOS, `Management0` represents the management interface on the active supervisor for modular systems. As virtual a virtual interface, some configuration related to physical interfaces can not be used on that interface like LLDP or speed. This behavior will not be changed by the cEOSlab team.

There is no default management interface in cEOSlab (this [comment](https://github.com/srl-labs/containerlab/blob/main/nodes/ceos/ceos.go#L299) is in fact imprecise).
When there is no [JSON interface mapping file](https://containerlab.dev/manual/kinds/ceos/#user-defined-interface-mapping) bound to the container, the management interface name is inferred using the `MGMT_INTF` environment variable value (which is defined [here](https://github.com/srl-labs/containerlab/blob/main/nodes/ceos/ceos.go#L52)): cEOSlab picks up the management interface ID from the kernel interface defined as management interface. Example: if the kernel interface is `ma1` (`MGMT_INTF=ma1`), the management interface will be `Management1`; if the kernel interface is `eth0` (`MGMT_INTF=eth0`) the management interface will be `Management0`.
Also, it seems that Docker really wants the first interface to be `eth0` and that cannot be changed.

The consequence of all of this is that we are using `Management0` interface "as the default" for years and we are hitting issues when configuring LLDP on that interface. In practice, we always use the [JSON interface mapping file](https://containerlab.dev/manual/kinds/ceos/#user-defined-interface-mapping) and use `Management1` when we need to disable LLDP on the Management interface, which happens very often.

Currently, one way to fix this situation in containerlab would be to always build the [JSON interface mapping file](https://containerlab.dev/manual/kinds/ceos/#user-defined-interface-mapping) on the fly based on the topology file interface names, make sure the mapping to `eth0` is set to `Management1` and bind it to the container.

hellt:
this decision should be made by some EOS folks, following the best practices. In fact, I think this code was delivered by them partially, so I will defer to someone like @ankudinov @bjmeuer

ankudinov:
@mtache please let's not rush it and finish the internal discussion first
[...]

Issue #2019, opened 2024-04-27
Management Interfaces  Links on Ceos


mahmutaydin1:
Hello,

I need a help about management interfaces connectivity. When I do links,  they are working well. However, show lldp neighbors  gives an output that Ma1 interfaces are connected to all nodes. I want to make them host ports.I think since it is bridge by default, it is seen as connected to each other.
Is there any paretemer to provide it?

`leaf1#show  lldp neighbors 
Last table change time   : 0:03:16 ago
Number of table inserts  : 12
Number of table deletes  : 0
Number of table drops    : 0
Number of table age-outs : 0

Port          Neighbor Device ID       Neighbor Port ID    TTL
---------- ------------------------ ---------------------- ---
Et1           leaf2                    Ethernet1           120
Et2           spine1                   Ethernet2           120
Et3           spine2                   Ethernet2           120
Et4           host1                    Ethernet1           120
Et6           leaf2                    Ethernet6           120
Ma1           host1                    Management1         120
Ma1           leaf3                    Management1         120
Ma1           spine1                   Management1         120
Ma1           leaf4                    Management1         120
Ma1           leaf2                    Management1         120
Ma1           spine2                   Management1         120
Ma1           host2                    Management1         120`

Issue #1469, opened 2023-07-14
Arista cEOS Management1 Interface


paullavelle:
Currently, Arista cEOS nodes launched in Containerlab use Management0 as the default management interface. Although this can be overridden by creating an EosIntfMapping.json file, it is preferable for Containerlab to use Management1 as the default management interface, aligning it with vEOS and EOS running on physical switches.

hellt:
@burnyd @sulrich does it match your view, guys?

ankudinov:
I'd rather prefer a simple knob TBH as that can break some existing labs if the default will be simply changed.

ankudinov:
One more concern is that if the fix will be to set a default mapping JSON, we have to make sure that "e1_1" -> "Ethernet1/1" conversion still works. As mapping JSON kills this possibility when set. And building complex topology with underscore is much easier than defining a full interface map.

paullavelle:
Closing issue as it is possible to specify Ma1 via EosIntfMapping.json if required:

{
  "ManagementIntf": {
    "eth0": "Management1"
  },
...

paullavelle:
Closing issue as it is possible to specify Ma1 via EosIntfMapping.json if required:

{
"ManagementIntf": {
"eth0": "Management1"
},
...

PR #3220, opened 2026-06-09
ceos: support remapping and manual wiring of the management interface

philsphicas:
Adds two opt-in management-interface options to the cEOS kind.

- `MGMT_INTF: maN` exposes the management interface as `ManagementN` (e.g. `Management1`) while still using the containerlab management network and its assigned address. `Management1` is the management interface name used by vEOS and physical EOS switches, so startup-configs and automation written for those devices can be reused without renaming the management interface.
- With `network-mode: none`, the management interface can be wired as a regular link to another node — cabled directly to a management host, or to a management switch that aggregates several nodes' management ports. No management address is assigned unless the link (or startup-config) provides one, so the interface only comes up addressed when the wiring is actually in place.

The default is unchanged: without `MGMT_INTF` (or with `eth0`) the node keeps `Management0` on the management network. Any non-`maN` value is passed through to cEOS with a warning. The legacy `MAPETH0` flag (a no-op on current cEOS images) is dropped automatically when `maN` is used.

Docs: new Management interface section on the cEOS kind page.

Tested with `arista_ceos`:
- default (`Management0` on the management network),
- `MGMT_INTF: ma1` on the management network (usable `Management1`, reachability preserved),
- `ma1` wired to a Linux node with `network-mode: none` (no auto address; address from the link).

Relates to #3037, #1469, #389, #2019.

### DIFF
--- a/docs/manual/kinds/ceos.md
+++ b/docs/manual/kinds/ceos.md
@@ -228,6 +228,75 @@ This topology will be equivalent to `ceos1:Ethernet1/1` connected to `ceos2:Ethe
 This feature can not be used together with interface mapping. If the interface mapping is in use, all names must be redefined in the map and the underscore naming option will not work. Also, it's only possible to rename Ethernet interfaces this way, not management ports.
 ///
 
+### Management interface
+
+By default a ceos node uses `Management0` as its management interface, reachable over the [containerlab management network](../network.md). cEOS derives the EOS management interface name from the kernel netdev that backs it, which is controlled by the `MGMT_INTF` environment variable:
+
+* `MGMT_INTF=eth0` (the default) --> `Management0`
+* `MGMT_INTF=ma1` --> `Management1` (and `maN` --> `ManagementN` in general)
+
+`Management1` is the management interface name used by vEOS and physical EOS switches. Exposing the management interface as `Management1` lets you reuse startup-configs and automation written for those devices without renaming the management interface.
+
+Only `eth0` (the default) and `maN` values are handled by containerlab. Any other value (for example `eth1`) is passed through to cEOS unchanged and logged with a warning, as cEOS would consume that interface as the management interface.
+
+/// note
+The legacy `MAPETH0` environment variable has no effect on current cEOS images and is automatically omitted when `MGMT_INTF` is set to a `maN` value. The management interface is selected solely by `MGMT_INTF`.
+///
+
+#### Keep the management network, expose it as `Management1`
+
+To keep the automatic containerlab management network (and its assigned management address) but have it appear as `Management1`, set `MGMT_INTF` to `ma1`. Containerlab renames the runtime-provided `eth0` netdev to `ma1` during boot so cEOS maps it to `Management1`; the management address, default route and `ssh`/`gNMI` reachability are preserved.
+
+```yaml
+name: ceos
+topology:
+  kinds:
+    arista_ceos:
+      env:
+        MGMT_INTF: ma1 # all ceos nodes use Management1 as their mgmt interface
+  nodes:
+    ceos1:
+      kind: arista_ceos
+      image: ceos:4.32.0F
+    ceos2:
+      kind: arista_ceos
+      image: ceos:4.32.0F
+  links:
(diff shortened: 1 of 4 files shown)


philsphicas:
Hi @hellt and @ankudinov - would appreciate your feedback on this when you get a chance. I know mgmt-interface naming has been discussed before (#3037 et al.), and I tried to address the concerns raised in those earlier discussions - it's explicit opt-in (`MGMT_INTF: ma1` gives you `Management1`), the `Management0` default is untouched, and the `eth1_1` shorthand still works. It also lets you wire the mgmt interface as a real link with `network-mode: none`, which makes it possible to test ZTP in more interesting, realistic topologies. Open to changes.

ankudinov:
@philsphicas Thank you for submitting this PR. Mgmt interfaces on cEOS-lab definitely need fixing.
We might be in the same event with @hellt this week and will try to find some time to discuss this.
I'd prefer to avoid Containerlab doing heavy lifting though and rely on cEOS-lab image:
1) interface map - that was always working, but requires a bit of work to be defined and mounted
2) INTF_MAP_ETH0=Management1 - this is the new env var we are planning to add to solve the problem. However it is not yet shipped.
So, as a starting point it would be enough to document the new env variable when it's shipped.
2nd step could be a deprecation warning for Ma0 as the default.
3rd step - swap the default and document INTF_MAP_ETH0=Management0 as opt-out knob.

philsphicas:
@ankudinov Thanks, that helps. Letting the image take care of the mapping definitely sounds like the best approach.

Two smaller things I'd still like to do, separately:

1. Allow wiring eth0 as a normal link (with `network-mode: none`). Right now the cEOS kind rejects eth0 as a link endpoint, while other platforms (linux/sros/6wind_vsr/fdio_vpp/veesix_osvbng) allow it. Any objection to a small PR for that?

2. Auto-generating the interface map. You said there might be appetite if it's opt-in. Is that still the case, or is the timeframe for the eth0-mapping so close that it's not worth it? For the opt-in I was thinking either a clab-side knob (e.g. `mgmt-interface: Management1`) or just honoring the future env var name so the same topology keeps working once the image supports it. Any preference?

hellt:
we've discussed this with @ankudinov at the conference, and it seems we agree that ceos will handle the interface mapping of `maN` in the image itself;

the bits related to mapping of the maN interfaces with `network-mode: none` are good to be proposed as a standalone PR

ankudinov:
@philsphicas
INTF_MAP_ETH0 is just around the corner and I'd definitely avoid adding interface map logic to the container lab. It's also an additional complexity that can be solved externally.
`network-mode: none` looks quite useful. Please make sure that you have all test case in place and short description for a lab topology to represent the use case and I'll test it as part of the review.
You can reduce the scope of this PR and rename it. But it could be easier to close this one and open another for `network-mode: none` only.

philsphicas:
Hi @ankudinov and @hellt ... Thanks very much for the feedback! I created #3237 and will abandon this one.

My feelings on the new PR are slightly mixed. It maintains the invariant that `eth0` is always `ManagementX` (whatever cEOS decides it should be), which is good, but it forces the user to patch it as `eth0` (not `maX`), which is slightly less user-friendly.

Either way, we need to manually insert an interface - the other option is to insert it as `maX` (still letting cEOS do the mapping to `ManagementX`), but that is basically the original PR.

Unfortunately, I won't be able to work on it for a couple of weeks ... if you need to make edits, feel free, otherwise I will resume in the first week of July.

hellt:
this begs the question if then we should wait for Ma1 to arrive first and then revisit this problem with a cleaner approach?

philsphicas:
This is slightly orthogonal to the future change.

Once you do network mode none, eth0 is gone. To wire the management interface, we need to plumb a new interface.

The main question is: do you *care* what the interface is named at the OS level?

Creating it as eth0 (what the change does) is compatible with INTF_MAP_ETH0 and lets cEOS map correctly (to ManagementX) but you still reference it as eth0 in the links.

Creating the interface as ma0 (or ma1 etc.) still lets cEOS do the mapping, and allowing wiring as maX, but it violates the principle that "eth0 is always the management interface".

PR #3222, opened 2026-06-10
fix Nokia SRSIM default config for integrated models

### LINKED ISSUE #3221: Nokia SRSIM default config for integrated models
Integrated SR-SIMs like SR-1 / SR-1s come up with empty card/mda/power-supply config
Would be nice to have this preconfigured as well

FloSch62:
Fixes #3221

### DIFF
--- a/docs/manual/kinds/sros.md
+++ b/docs/manual/kinds/sros.md
@@ -241,6 +241,8 @@ topology:
 
 ///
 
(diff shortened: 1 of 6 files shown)


kaelemc (approved):
big fan of your work @FloSch62 :)

hellt (approved):

PR #3227, opened 2026-06-11
Fix node-scoped magic vars in inherited extras

rrajpuro:
Fixes a regression in magic variable expansion for node extras inherited from `kinds`, `groups`, or `defaults`.

PR #3096 added magic variable expansion for `extras`, but `processNodeExtras` mutated the inherited `*Extras` object in place. When multiple nodes inherited the same kind-level `ceos-copy-to-flash` path containing `__clabNodeName__`, the first node's replacement was reused by later nodes.

This changes `processNodeExtras` to copy the extras value and mutable slices before replacement, keeping expansion node-scoped.

Tested with:

```bash
go test ./core -run 'Test(KindExtrasMagicVarsAreNodeScoped|ExtrasInit)$'
```

### DIFF
--- a/core/config.go
+++ b/core/config.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


hellt:
thanks @rrajpuro

PR #3238, opened 2026-06-16
Release v1.57.0-2

biplamal:


hellt:
Hi @biplamal 
I think it is time we move out the ixia example from the clab repo and you host it outside of it.

There is little sense to maintain a sub package for the ixia example, instead I propose you create a repo and we link to it. That will enable you to introduce any code changes you want at any time and won't induce code churn on clab core.

PR #3240, opened 2026-06-18
Add components: support for vSIM

kaelemc:
Some of us who work with SR OS a lot got addicted a bit to the `components:`, however this is currently SR-SIM only which is only 25.7 onwards.

While some of the older versions of SR OS are within support, we sometimes need to use them in clab, thus meaning we have to use the vSIM which brings the *fun* of building the TiMOS line.

Instead we can parse `components:` and let clab try to build it for us and make life easy.

Few things are wonky since vSIMs are VMs, decided to put the `max_nics`, `cpu` and `ram` parameters as env vars under the slot.

### DIFF
--- a/docs/manual/kinds/vr-sros.md
+++ b/docs/manual/kinds/vr-sros.md
@@ -93,8 +93,11 @@ The interface naming convention is: `1/1/X`, where `X` is the port number.
 
 /// admonition
     type: warning
-Nokia SR OS VM nodes currently only support the simplified interface alias `1/1/X`, where X denotes the port number.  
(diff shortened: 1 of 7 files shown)


hellt:
@kaelemc @FloSch62 is this good to go?

kaelemc:
Almost. will request your review when ready.

sacckth:
do we have a death wish?

kaelemc:
> do we have a death wish?

Yes, unless you can get me the 24.10.R1 srsim :)

kaelemc:
Seems to work well enough for me.

hellt:
probably needs a doc addition

PR #3241, opened 2026-06-21
feat: support SSH IdentityFile via node credentials

### LINKED ISSUE #3224: Support SSH identity file in generated SSH config (e.g. via `credentials`)
Currently clab generates a per-lab SSH config (`/etc/ssh/ssh_config.d/clab-<lab>.conf`) that sets User, host-key options and optionally `PubkeyAuthentication`. 
However, when nodes only accept a specific SSH key, users still have to pass it manually.
Currently there is no way to specify the SSH IdentityFile in the clab topology.

I would propose to extend the `credentials` mapping  (which already feeds the `User` directive of the generated SSH config) with an identity file field:

```yaml
topology:
  defaults:
    credentials:
      username: root
      identity-file: ~/path/to/id_ed25519
```

this would then render in the SSH config as follows:

```
Host clab-mylab-node1
    User root
    IdentityFile ~/path/to/id_ed25519
    StrictHostKeyChecking=no
    UserKnownHostsFile=/dev/null
```

Would that be acceptable to add in such a way? Or is that beyond the scope of the credentials section and it would be cleaner to have such an option in a separate section?

I'm happy to contribute a patch if we agree on a solution.

mvanhorn:
## Summary

Adds an optional `identity-file` key to the `credentials` mapping so a topology can declare the SSH private key a node expects. When set, containerlab renders an `IdentityFile` directive into the generated per-lab SSH client config (`/etc/ssh/ssh_config.d/clab-<lab>.conf`), so `ssh <node>` uses that key automatically.

The new field is threaded through the same precedence chain as `username`/`password`:

`NodeCredentials.IdentityFile` → `resolveTopologyCredentials` / `GetNodeIdentityFile` (node > group > kind > defaults) → `createNodeCfg` → `addSSHConfig` (`SSHConfigNodeTmpl.IdentityFile`) → `ssh_config.go.tpl`.

`credentialsPresent` now also treats `identity-file` as a present-credentials signal, so a node that sets only `identity-file` still wins precedence at its level (no merging across levels, consistent with the existing username/password behavior).

## Why this matters

Today, when a lab node only accepts a specific SSH key, users have to pass `-i <key>` on every `ssh` invocation because there is no way to declare the key in the topology. This adds that declaration in the natural place — alongside `username`/`password` under `credentials` — and emits a standard `IdentityFile` directive.

Ref #3224. Maintainer @hellt greenlit the addition on 2026-06-13 ("This sounds reasonable").

The rendered path is double-quoted so paths containing spaces are handled correctly; OpenSSH still performs its own `~` expansion at connect time. The new key is also added to `schemas/clab.schema.json` so editor/VS Code schema validation picks it up.

## Testing

- `core/sshconfig_test.go` (new): renders the embedded `ssh_config.go.tpl` and asserts the `IdentityFile` directive is emitted when set (quoted), omitted when unset, rendered without a `User` line when only `identity-file` is set, and quoted for paths containing spaces.
- `types/topology_test.go`: extends `TestGetNodeCredentials` to cover identity-file precedence — defaults-level applies to a node, node-level overrides defaults, and a node setting only `identity-file` still wins precedence over a kind that sets username/password.
- `gofmt`, `go vet`, and test compilation pass for the `types` and `core` packages (verified for the `linux` target — the upstream CI platform).
- The quoted template output was validated with `ssh -G` to confirm both `~` expansion and space-containing paths parse correctly.

Fixes #3224

### DIFF
--- a/core/assets/ssh_config.go.tpl
+++ b/core/assets/ssh_config.go.tpl
@@ -6,6 +6,9 @@ Host{{- range .Names }} {{ . }}{{- end }}
 	{{-  if ne $node.Username ""}}
 	User {{ $node.Username }}
 	{{- end }}
+	{{- if ne $node.IdentityFile "" }}
+	IdentityFile "{{ $node.IdentityFile }}"
+	{{- end }}
 	StrictHostKeyChecking=no
 	UserKnownHostsFile=/dev/null
 	{{- if ne $node.SSHConfig.PubkeyAuthentication "" }}
--- a/core/config.go
+++ b/core/config.go
@@ -253,6 +253,14 @@ func (c *CLab) createNodeCfg( //nolint: funlen
 	// defaults from the node registry.
 	nodeCfg.Credentials.Username = c.Config.Topology.GetNodeUsername(nodeName)
 	nodeCfg.Credentials.Password = c.Config.Topology.GetNodePassword(nodeName)
+	// Resolve the SSH identity-file path against the topology directory (and expand ~) so the
+	// rendered ssh_config works regardless of the cwd ssh is later invoked from.
+	if identityFile := c.Config.Topology.GetNodeIdentityFile(nodeName); identityFile != "" {
+		nodeCfg.Credentials.IdentityFile = clabutils.ResolvePath(
+			identityFile,
+			c.TopoPaths.TopologyFileDir(),
+		)
+	}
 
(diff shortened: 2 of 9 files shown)


mvanhorn:
Thank you for the second one, @hellt. Supporting SSH IdentityFile via node credentials is a nice convenience for topology authors.

PR #3244, opened 2026-06-22
fix: skip non point-to-point links in graph generation (dummy link panic)

### LINKED ISSUE #3226: [BUG] `graph` issues with a dummy link type
### Describe the bug

When launching `containerlab graph` with a file with a dummy link type, the application encounters an error:

```bash
13:14:07 INFO Parsing & checking topology file=dummy.clab.yml
13:14:07 INFO building graph from topology file
panic: runtime error: index out of range [1] with length 1
```
Above issue is observed also with the following flags:

- `--mermaid`
- `--dot`

**Important**: there is no issue when  creating a Draw.io version with `--drawio` flag. Also Visual Studio Code Extension is able to draw the topology.

### Environment

OS: Linux Ubuntu 24.04
Containerlab: 0.75.0

### Steps to reproduce

1. Create a file with the following content:
```yaml
name: dummy-demo

topology:
  nodes:
    router1:
      kind: linux
      image: alpine:latest

  links:
    - type: dummy
      endpoint:
        node: router1
        interface: eth1
```
2. Run the following command:
```bash
containerlab graph --offline -t ~/pathToTopology/dummy.clab.yml
```

### Expected behavior

Generate a HTML file (or another format, depending on an additional flags) with topology with a dummy link type.

mvanhorn:
## Summary

`containerlab graph` panics with `index out of range [1] with length 1` when the topology contains a `dummy` link. A dummy link attaches a single node interface and therefore exposes only one endpoint, but every graph generator iterated `c.Links` and unconditionally indexed `eps[1]`. This adds a point-to-point guard so single-endpoint (and zero-endpoint) links are skipped instead of crashing graph generation.

## Why this matters

A `type: dummy` link makes `link.GetEndpoints()` return a slice of length 1, so indexing the second endpoint panics. The crash affected all three graph code paths:

- `core/graph.go:131` in `GenerateDotGraph` (the default `--dot` output)
- `core/graph.go:293` in `GenerateMermaidGraph` (`--mermaid`)
- `cmd/graph.go:178` in the default HTML topology builder (`containerlab graph` with no flags)

The HTML/drawio renderers already avoided the crash by only enumerating nodes, so the fix mirrors that behavior for the link loops: links that are not point-to-point are logged at debug level and skipped rather than rendered as an edge. This keeps `dummy`, and other single-host link types (`host`, `macvlan`, `mgmt-net`) from breaking graph generation while still rendering every genuine node-to-node edge.

## Testing

- Added `core/graph_test.go` covering single-endpoint (dummy) and zero-endpoint links for both the dot and mermaid generators, plus a two-node veth regression case for mermaid. The tests build the real links and call the real `GenerateDotGraph` / `GenerateMermaidGraph` functions. Without the fix the dummy case reproduces the exact `index out of range [1] with length 1` panic from the issue; with the fix the generators produce output with no edge for the single-endpoint link.
- `go test ./core/`, `go vet ./cmd/ ./core/`, `go build ./cmd/... ./core/...`, and `gofmt -l` all pass (run with GOOS=linux).

Fixes #3226

### DIFF
--- a/cmd/graph.go
+++ b/cmd/graph.go
@@ -166,6 +166,13 @@ func graphFn(ctx context.Context, o *Options) error {
 
 	for _, l := range c.Links {
 		eps := l.GetEndpoints()
+		// Skip links that are not point-to-point (e.g. dummy, host, macvlan,
+		// mgmt-net links expose a single endpoint). Indexing eps[1] on such
+		// links would panic, so they are not rendered as edges.
+		if len(eps) != 2 {
+			log.Debugf("skipping non point-to-point link with %d endpoint(s) in graph topology", len(eps))
+			continue
+		}
 
 		ifaceDisplayNameA := eps[0].GetIfaceDisplayName()
 		ifaceDisplayNameB := eps[1].GetIfaceDisplayName()
--- a/core/graph.go
+++ b/core/graph.go
@@ -120,6 +120,13 @@ func (c *CLab) GenerateDotGraph(ctx context.Context) error {
 		attr["color"] = black
 
 		eps := link.GetEndpoints()
+		// Skip links that are not point-to-point (e.g. dummy, host, macvlan,
+		// mgmt-net links expose a single endpoint). Indexing eps[1] on such
+		// links would panic, so they are not rendered as edges.
+		if len(eps) != 2 {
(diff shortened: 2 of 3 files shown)


mvanhorn:
Thanks @hellt, and that's three - skipping non point-to-point links in graph generation keeps containerlab graph from panicking on dummy links.

PR #3256, opened 2026-06-29
Add containerlab best practices skills

FloSch62:

PR #3260, opened 2026-06-30
Improve apply link reconciliation

FloSch62:
## Summary
- move apply runtime endpoint selection onto the `links.Link` contract with `GetRuntimeEndpoints()`
- keep stitched-link capability in the `links` package and update apply/deploy callers
- persist apply state and reset resolved links for idempotent apply paths
- remove short kind aliases from the apply docs table
- select the VXLAN stitch endpoint by interface name instead of endpoint slice order

## Validation
- `go test ./links ./core ./cmd`

Split out from #3256 so that PR can stay focused on agent skills only.

### DIFF
--- a/cmd/tools_veth.go
+++ b/cmd/tools_veth.go
@@ -131,7 +131,7 @@ func vethCreate(o *Options) error {
 
 	// deploy the endpoints of the Link
 	for _, ep := range link.GetEndpoints() {
-		if err := clablinks.DeployEndpoint(ctx, ep); err != nil {
+		if err := link.Deploy(ctx, ep); err != nil {
 			return fmt.Errorf("failed to deploy veth endpoint: %w", err)
 		}
 	}
--- a/cmd/tools_veth.go
+++ b/cmd/tools_veth.go
@@ -131,12 +131,7 @@ func vethCreate(o *Options) error {
 
 	// deploy the endpoints of the Link
 	for _, ep := range link.GetEndpoints() {
-		deployable, ok := ep.(clablinks.DeployableEndpoint)
-		if !ok {
-			return fmt.Errorf("endpoint %q is not deployable", ep.GetIfaceName())
-		}
-
-		if err := deployable.Deploy(ctx); err != nil {
+		if err := clablinks.DeployEndpoint(ctx, ep); err != nil {
 			return fmt.Errorf("failed to deploy veth endpoint: %w", err)
 		}
 	}
--- a/cmd/tools_vxlan.go
+++ b/cmd/tools_vxlan.go
@@ -5,6 +5,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"net"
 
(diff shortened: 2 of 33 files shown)


steiler:
**Dead code left behind** 
Endpoint Deploy() methods on EndpointVeth, EndpointHost, etc. are now unused. Harmless, but could be cleaned up in a follow-up.

cmd/tools_veth.go
```diff
@@ -131,12 +131,7 @@ func vethCreate(o *Options) error {
 
 	// deploy the endpoints of the Link
 	for _, ep := range link.GetEndpoints() {
-		deployable, ok := ep.(clablinks.DeployableEndpoint)
-		if !ok {
-			return fmt.Errorf("endpoint %q is not deployable", ep.GetIfaceName())
-		}
-
-		if err := deployable.Deploy(ctx); err != nil {
+		if err := clablinks.DeployEndpoint(ctx, ep); err != nil {
```
steiler: Why use a Function instead of the Method that was there?
But im my view you should even just call link.Deploy() ... and the link should figure out what needs to be done...

hellt: this func indeed calls `link.Deploy(ctx, ep)`. It is a helper method that adds a couple of nil checks prior to calling `link.Deploy(ctx, ep)`

steiler: I see that... but if you have an endpoint, there should by contract be a Link associated... and I don't see why you would have a nil endpoint...
So if you then simply call Deploy on endpoint, the links is implicitely being deployed ...

steiler: by now there is just way to many functions floting around somewhere that magically do things that should be covered as an integrated part of the standards workflow.

```
// activateEndpoint brings the endpoint's interface up in its current namespace.
func activateEndpoint(ctx context.Context, e Endpoint) error {
	return e.GetNode().ExecFunction(ctx, func(_ ns.NetNS) error {
		link, err := netlink.LinkByName(e.GetIfaceName())
		if err != nil {
			return err
		}
		return netlink.LinkSetUp(link)
	})
}
```
why is this not implemented as Activate function on the GenericEndpoint... but referenced in every Endpoint and called specifically... things like that... the cohesion of this stuff is quite low and therefore unfortunately the complexity is high.

move endpoint. the same
```
func moveEndpoint(ctx context.Context, e Endpoint, dst Node) error {
```

hellt: removed the deploy helper in 84f3c64a7

hellt: I'll put move and activate on the generic endpoint, good call.

hellt: the move func turned to be dependent on the concrete endpoint type, so promoting it to generic endpoint did not work. But I removed the interface MoveTo method that was redundant [eb5578f](https://github.com/srl-labs/containerlab/pull/3260/commits/eb5578f58ceee9d3c56abe902e06646b96a86db7)

hellt:
lets have it in and soak it through user testing

PR #3262, opened 2026-06-30
Add ciena_saos10 kind for Ciena SAOS 10.x (#1)

jgroom33:
* Add ciena_saos10 kind for Ciena SAOS 10.x

Adds a first-class VM kind for Ciena SAOS 10.x routers run as a QEMU VM in a container.

- nodes/ciena_saos10: VRNode-based kind registering ciena_saos10; numeric-port interface mapping (1 -> eth1); a   supported-variant (type) allowlist; partial-startup-config-only handling that exports SAOS_STARTUP_CONFIG_PATH to the launcher.
- core/register.go: register the kind.
- schemas/clab.schema.json: add name to the kind enum and per-kind block.
- docs: new ciena_saos10 kind page (incl. partial-config apply behavior),
  kinds index row, and mkdocs nav entry.
- lab-examples/ciena_saos10: minimal two-node example topology.


mkdocs.yml
```diff
@@ -56,6 +56,7 @@ nav:
       - Ostinato: manual/kinds/ostinato.md
       - Spirent TestCenter: manual/kinds/spirent_stc.md
       - Check Point Cloudguard: manual/kinds/checkpoint_cloudguard.md
+      - Ciena SAOS 10: manual/kinds/ciena_saos10.md
```
hellt: is SAOS 10 the name of the Network OS? Should `10` be in the name or is it more like a version family?
In latter case I think the `10` can be dropped?

docs/manual/kinds/ciena_saos10.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Managing ciena_saos10 nodes
+
+Ciena SAOS 10 nodes launched with containerlab can be managed via the following interfaces:
+
+=== "CLI via SSH"
+    to connect to the SAOS CLI (password `ciena123`)
+    ```bash
+    ssh diag@<container-name/id>
+    ```
```
hellt: we use the new block syntax for tabs and admonitions.

e.g.

```
/// tab | CLI via SSH
to connect to the SAOS CLI (password `ciena123`)
```bash
ssh diag@<container-name/id>
```
///
```

File: docs/manual/kinds/ciena_saos10.md [hunk 3]
```diff
@@ -0,0 +1,127 @@
...
+    ssh diag@<container-name/id> -p 830 -s netconf
+    ```
+=== "Console via telnet"
+    to connect to the serial console
+    ```bash
+    telnet <container-name/id> 5000
+    ```
+
+!!!info
+    Default user credentials: `diag:ciena123`
```
hellt: we use the new block syntax for tabs and admonitions.

e.g.

```
/// info
Default user credentials: `diag:ciena123`
///
```

mkdocs.yml
jgroom33: Ciena has SAOS 6, SAOS 8, SAOS 10 for their routers/switches. They differ significantly is syntax and capability. (i.e. SAOS 6 has limited routing capability) There's a low probability that 6 or 8 will have a simulator, but the differentiation is beneficial.

jgroom33: @hellt Either approach is acceptable. What is the preference?

hellt: Let's name them just `ciena_saos` and call out this differentiation in the docs.

PR #3266, opened 2026-07-07
Add validate command.

kaelemc:
Add `clab validate` command which lets us validate the topology file is valid.

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/charmbracelet/log"
 	"github.com/spf13/cobra"
 	clabcore "github.com/srl-labs/containerlab/core"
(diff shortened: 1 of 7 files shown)


cmd/validate.go
```diff
@@ -0,0 +1,47 @@
+package cmd
```
hellt, line 1: what is the operational pattern for this command?
I am not against it, just wondering when it is useful?

kaelemc: My specific usecase is some automation where some script may get any topo file passed and it should check if it's all good/valid before continuing with it's next steps. Prevents having to implement a separate parser to validate the topo.

Another one is a little QoL thing where the YAML schema is not available. either CLI only environment with non-fancy editors or maybe you've got a massive topology where the YAML schema extension won't activate since the topo file is too huge.

PR #3269, opened 2026-07-08
sonic-vm: inject authorized_keys via scrapligo SSH

steiler:
## Summary
- switch `sonic-vm` SSH key deployment from interactive console/telnet flow to password-authenticated scrapligo SSH against the management IP
- add robust shell-safe key writing (`printf` with escaping) and remove console-specific command/event handling
- move vrnetlab readiness probing to shared `VRNode.IsVrnetlabHealthy()` and use it from both `sonic-vm` and `vr-sros`
- add/update unit tests for generated authorized_keys command payloads

## Test plan
- [x] `go build ./nodes/...`
- [x] `go test ./nodes/sonic_vm/... ./nodes/vr_sros/...`
- [x] deploy a `sonic-vm` lab and verify passwordless SSH login works after post-deploy key injection

### DIFF
--- a/docs/manual/kinds/sonic-vm.md
+++ b/docs/manual/kinds/sonic-vm.md
@@ -5,6 +5,7 @@ kind_code_name: sonic-vm
 kind_display_name: SONiC (VM)
 ---
 # -{{ kind_display_name }}-
+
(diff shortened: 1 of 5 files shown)


hellt:
LGTM

I changed the healthcheck method to use the container runtime, worked well on my end. Check if that stays OK on your end as well and then we can merge

hellt (approved):

steiler:
@hellt lgtm and works!

PR #3270, opened 2026-07-08
Add veth-stitch link type to support SR-SIM link impairments (prev #3268)

kaelemc:
Copied from #3268 so we can run the SRSIM test suite.

This PR is changed now to implement a `veth-stitch` link type instead of adding the `links.impairments` and the related mode. `veth-stitch` basically creates a link which has a veth inserted in the middle inside it's own 'stitching netns'. 

The veth can probably live in the host netns but putting this new veth inside it's own netns is just so we don't hit the 15 char name limit on interfaces and the link is staying isolated in a sense. You can also quite easily then tcpdump based on node name when in the netns for the stitched link.

Downside is you have a chance to have orphan netnses on the system.

The reason this is needed is because SR-SIMs magical packet processing means we have had two limitations:

- tc netem impairments didn't work.
- packet capture only worked in one direction?/was unreliable.

With this new link type both are fixed as basically we now insert a new veth on the link of which each end on this new veth would act as a 'delegate'/psuedo-interface of the actual link.

The next thing to add is probably detect if the node is SR-SIM and then create the link as a `veth-stitch` instead of a `veth`, unless link type is explicitly set by the user.


docs/manual/topo-def-file.md
```diff
@@ -272,6 +272,31 @@ links:
     labels: <link-labels>                   # optional (used in templating)
 ```
 
+###### veth-stitch
+
+The veth-stitch link type transparently stitches its two endpoints by using three veth links. One per node interface, and a third veth which stitches the two together.
+This is useful for nodes whose datapath interfaces cannot be captured or impaired directly (such as SR-SIM).
```
hellt: @kaelemc lets add a mermaid diagram or something that would explain what exactly is being done here?

hellt:
## veth-stitch implementation review — updated at `e34c5b2cb`

### Goal and resulting datapath

This PR gives SR-SIM links ordinary Linux interfaces that can be captured and impaired:

`node A ↔ host far-A ⇄ tc mirred redirects ⇄ host far-B ↔ node B`

The two `clab-s-*` far ends live in the root namespace. Traffic from node A enters far-A, is redirected out of far-B, and reaches node B; the reverse rule handles B → A. The tools altname for node A is intentionally placed on far-B, because netem on far-B egress impairs A → B traffic.

Brief-link behavior:

- SR-SIM ↔ SR-SIM: automatically resolves to `veth-stitch`.
- SR-SIM ↔ SR Linux: automatically resolves to `veth-stitch`.
- SR Linux ↔ SR Linux: remains a normal `veth`.
- Explicit `type: veth`: remains a normal veth and acts as the SR-SIM opt-out.
- Explicit `type: veth-stitch`: opts any supported endpoint pair into stitching.

### Issues resolved during review

- Replaced the broad `Node.DefaultLinkType` requirement with an optional raw-link capability; only SR-SIM opts in.
- Allowed MAC, IPv4, and IPv6 metadata on node-facing stitched endpoints and preserved full `LinkCommonParams`.
- Aligned the schema with Go by making explicit `veth-stitch` endpoints object-only. Typeless brief links still support string endpoints.
- Made `LinkVEthStitched` the lifecycle owner for resolved links:
  - node-facing endpoints bind to the composite;
  - composite deploy routes to the appropriate segment;
  - synchronized remove tears down both segments exactly once;
  - full deployment state is set only after both TC redirects succeed.
- Added both host far ends to `GetRuntimeEndpoints`, so apply can inventory and clean all four material endpoints.
- Moved filtered-destroy cleanup into the links package. Core only invokes the generic `CleanupFilteredLinks` hook; veth-stitch owns altname selection and host-interface removal.

### Remaining review points

1. **Initial deploy still treats link `PostDeploy` failure as best-effort.** `core/deploy.go` logs and continues. If TC lookup, altname setup, or redirect installation fails, deploy can report success with two disconnected segments. This needs an explicit decision on whether initial deploy should fail or remain best-effort.
2. **Tools identity depends on altname support.** `markFarEnd` tolerates `EOPNOTSUPP`; tools lookup then falls back to the container interface, which is the path SR-SIM cannot reliably capture or impair. A clear error or deterministic primary-name fallback would be safer.
3. **The PR description is stale.** It still describes a private stitching namespace and says SR-SIM auto-defaulting is future work; the implementation now uses root-namespace far ends and already auto-promotes brief SR-SIM links. The auto-default and explicit `type: veth` escape hatch should also be documented.

### Verification

- `go test ./...`
- `go test -race ./links ./core`
- SR-SIM Robot suite covers explicit stitch deploy, ping, inspect overlay, netem set/show/reset, recovery, and full cleanup.

High-value remaining integration coverage: brief auto-promotion in Robot, one-node filtered destroy, apply remove/recreate of a stitched link, and packet capture.

hellt:
srsim-test 06* fails as I have updated the srsim image to 26.7.r1:

```
Current Time: Thu Aug 06 11:32:49 UTC 2026
Loading TiMOS License for 'TiMOS-B-26.7.R1' from location:
   'cf3:\license.txt' ... OK

Loading primary configuration file "cf3:\config.cfg"
Loaded 202 lines in 0.1 seconds from file "cf3:\config.cfg"
WARNING: MGMT_CORE #2204: Obsolete element - configure system grpc md-cli
WARNING: MGMT_CORE #2204: Obsolete element - configure system grpc md-cli admin-state
WARNING: MGMT_CORE #2204: Obsolete element - configure system security aaa local-profiles profile "administrative" grpc rpc-authorization md-cli-session
WARNING: MGMT_CORE #2206: Obsolete element value - configure system security tls server-cipher-list "clab-all" tls13-cipher 5 name tls-aes128-ccm8-sha256 - obsolete enumeration value
Committing configuration WARNING: MGMT_CORE #2208: configure system security tls server-cipher-list "clab-all" tls13-cipher 5 name - Unsupported element - field value, must delete from configuration

CRITICAL: MGMT_CORE #2508: Configuration load failed - using default configuration
```
this is on sar1 node

/cc @sacckth

nodes/sros/version.go
```diff
@@ -25,6 +25,8 @@ const (
 	// vrnetlabVersionLabel is set on vrnetlab-built images; nokia_srsim kind must not use such
 	// images.
 	vrnetlabVersionLabel = "vrnetlab-version"
+
+	srosDefaultVersion = "25.7.R1"
```
hellt, line 32: why is this a const? we should get the accurate version from the system itself

kaelemc: We do, it's just a fallback as a nil version caused some panic somewhere as some of the first releases didn't have the version meta in the image.

hellt: ah, yeah, I missed the part where it defaults to this val only when fetching fails. WIll add comments

PR #3275, opened 2026-07-14
Add first-class container runtime options

orrious:


orrious:
Rebased this branch onto current `main` and force-pushed the updated head.

Current head: `50d51d076`

- x86 runner: passed focused branch validation
- ARM runner: passed focused branch validation

PR #3276, opened 2026-07-14
Honor lifecycle dependencies when starting nodes

orrious:
## Summary

Make node startup honor lifecycle dependencies instead of starting every selected node concurrently.

- derive startup ordering from `wait-for` lifecycle dependencies
- order shared-network-namespace children after their provider
- validate missing dependencies and dependency cycles
- preserve existing behavior for independent nodes by starting them in the same batch
- cover ordering, validation, and external-container dependencies with unit tests

## Validation

- Linux Go 1.26.5: `go test ./core`

### DIFF
--- a/core/lifecycle.go
+++ b/core/lifecycle.go
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


orrious:
Rebased this branch onto current `main` and force-pushed the updated head.

Current head: `cbc2d0b5d`

- x86 runner: `go test ./core` passed
- ARM runner: `go test ./core` passed

PR #3277, opened 2026-07-14
Detect Podman netavark management bridges

orrious:
## Summary

Detect the actual Linux bridge backing a Podman netavark management network instead of assuming Docker's generated `br-<id>` naming convention.

- read netavark's `network_interface` field from the raw network inspect response
- fall back to Podman's libpod network endpoint when the compatibility response omits that field
- retain Docker bridge-name handling as the final fallback
- preserve current behavior for non-bridge Docker networks
- document the fix in the release notes

## Validation

- Linux Go 1.26.5: `go test ./runtime/docker`

### DIFF
--- a/docs/rn/0.77.md
+++ b/docs/rn/0.77.md
@@ -67,5 +67,6 @@ Thanks @FloSch62 #3242 #3245 #3246
 
 ## Miscellaneous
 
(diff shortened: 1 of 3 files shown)


orrious:
Rebased this branch onto current `main`, added the missing release-note entry, and force-pushed the updated head.

Current head: `ccf195fef`

- x86 runner: `go test ./runtime/docker` passed
- ARM runner: `go test ./runtime/docker` passed

PR #3280, opened 2026-07-14
Use runtime Exec for cEOS post-deploy configuration

orrious:
## Summary

Use containerlab's configured runtime connection for cEOS post-deploy configuration instead of spawning an interactive host-side `docker/podman exec -it` process.

Existing path:

```text
deploy -> cEOS PostDeploy -> host docker/podman exec -> interactive EOS CLI
```

New path:

```text
deploy -> cEOS PostDeploy -> Runtime.Exec -> batched EOS CLI command
```

This keeps execution inside the Docker or Podman runtime abstraction, avoids interactive TTY behavior, captures the return code/stdout/stderr, and bounds CLI-readiness retries at 60 attempts. The EOS command batch explicitly enters configuration mode, uses `--abort-on-error`, and finishes with `end` and `write memory`.

The behavior remains cEOS-specific: post-deploy still applies the runtime-assigned management address and topology-defined data-interface addresses.

## Validation

- added unit coverage for command construction, success, retry, and error handling
- added cEOS Robot coverage for runtime-assigned Management0 IPv4
- Linux Go 1.26.5: `go test ./nodes/ceos`
- strict Zensical documentation build: `make docs`

### DIFF
--- a/nodes/ceos/ceos.go
+++ b/nodes/ceos/ceos.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	_ "embed"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -16,6 +15,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/charmbracelet/log"
 	clabconstants "github.com/srl-labs/containerlab/constants"
@@ -347,16 +347,10 @@ func setMgmtInterface(node *clabtypes.NodeConfig) error {
 }
 
 // ceosPostDeploy runs postdeploy actions which are required for ceos nodes.
-func (n *ceos) ceosPostDeploy(_ context.Context) error {
+func (n *ceos) ceosPostDeploy(ctx context.Context) error {
 	nodeCfg := n.Config()
-	d, err := clabutils.SpawnCLIviaExec("arista_eos", nodeCfg.LongName, n.Runtime.GetName())
-	if err != nil {
-		return err
-	}
-
-	defer d.Close()
-
 	cfgs := []string{
+		"configure terminal",
 		"interface " + nodeCfg.MgmtIntf,
 		"no ip address",
 		"no ipv6 address",
@@ -410,18 +404,40 @@ func (n *ceos) ceosPostDeploy(_ context.Context) error {
 	}
 
 	// add save to startup cmd
-	cfgs = append(cfgs, "wr")
+	cfgs = append(cfgs, "end", "write memory")
 
 	log.Debugf("cEOS PostDeploy configuration for node %s: %v", n.Cfg.ShortName, cfgs)
 
-	resp, err := d.SendConfigs(cfgs)
-	if err != nil {
-		return err
-	} else if resp.Failed != nil {
-		return errors.New("failed CLI configuration")
+	var lastErr error
+	var lastResp *clabexec.ExecResult
+	cliCmd := "Cli -p 15 --abort-on-error -c $'" + strings.Join(cfgs, "\n") + "'"
+
+	for range 60 {
+		execCmd := clabexec.NewExecCmdFromSlice([]string{"/bin/bash", "-lc", cliCmd})
+		resp, err := n.RunExec(ctx, execCmd)
+		if err == nil && resp.GetReturnCode() == 0 {
+			return nil
+		}
+
+		lastErr = err
+		lastResp = resp
+		log.Debugf("%s - Cli not ready (%v, %v) - waiting.", nodeCfg.LongName, err, resp)
+		time.Sleep(2 * time.Second)
 	}
 
-	return err
+	if lastErr != nil {
+		return lastErr
+	}
+	if lastResp != nil {
+		return fmt.Errorf(
+			"failed CLI configuration: rc=%d stdout=%q stderr=%q",
+			lastResp.GetReturnCode(),
+			lastResp.GetStdOutString(),
+			lastResp.GetStdErrString(),
+		)
+	}
+
+	return fmt.Errorf("failed CLI configuration")
 }
 
 // CheckInterfaceName checks if a name of the interface referenced in the topology file correct.
--- a/nodes/ceos/ceos.go
+++ b/nodes/ceos/ceos.go
@@ -419,10 +419,15 @@ func (n *ceos) ceosPostDeploy(ctx context.Context) error {
 
 	var lastErr error
 	var lastResp *clabexec.ExecResult
-	cliCmd := "Cli -p 15 --abort-on-error -c $'" + strings.Join(cfgs, "\n") + "'"
+	cliCmd := []string{
+		"Cli",
+		"-p", "15",
+		"--abort-on-error",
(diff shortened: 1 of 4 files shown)


ankudinov (changes requested):
@orrious Not sure what is this PR about but as it's not in a draft status - please add the docs and tests required before making it ready for review.

orrious:
Addressed the requested follow-up on this PR.

Changes pushed:
- added cEOS postdeploy unit tests
- added cEOS coverage for runtime-assigned Management0 IPv4 in Robot
- added the corresponding docs note
- fixed the unit test endpoint type issue

Validation:
- real x86 runner: `go test ./nodes/ceos` and build passed
- real ARM runner: `go test ./nodes/ceos` and build passed
- GitHub-hosted build/unit/docs jobs also passed on amd64 and arm64; hosted cEOS integration failed only on GHCR image pull

orrious:
@ankudinov Following up on your July 28 review: the requested docs and tests are now added on this branch, along with the cEOS Robot coverage described in the latest update. Please take another look when you have a moment.

orrious:
@ankudinov I rebased this branch onto current `main` today (August 5, 2026), so the requested docs/tests follow-up is now on top of the latest base as well. When you have a moment, please take another look.

nodes/ceos/ceos.go
```diff
@@ -410,18 +413,40 @@ func (n *ceos) ceosPostDeploy(_ context.Context) error {
...
-	if err != nil {
-		return err
-	} else if resp.Failed != nil {
-		return errors.New("failed CLI configuration")
+	var lastErr error
+	var lastResp *clabexec.ExecResult
+	cliCmd := "Cli -p 15 --abort-on-error -c $'" + strings.Join(cfgs, "\n") + "'"
+
+	for range 60 {
+		execCmd := clabexec.NewExecCmdFromSlice([]string{"/bin/bash", "-lc", cliCmd})
```
ankudinov: Should we really involve bash here? Is there a specific use case for that? Is direct Cli call an option?

```golang
clabexec.NewExecCmdFromSlice([]string{
    "Cli",
    "-p", "15",
    "--abort-on-error",
    "-c", strings.Join(cfgs, "\n"),
})
```

nodes/ceos/ceos.go
```diff
@@ -410,18 +413,40 @@ func (n *ceos) ceosPostDeploy(_ context.Context) error {
...
-	resp, err := d.SendConfigs(cfgs)
-	if err != nil {
-		return err
-	} else if resp.Failed != nil {
-		return errors.New("failed CLI configuration")
+	var lastErr error
+	var lastResp *clabexec.ExecResult
+	cliCmd := "Cli -p 15 --abort-on-error -c $'" + strings.Join(cfgs, "\n") + "'"
+
+	for range 60 {
```
ankudinov: If understand this correctly, this simply throws the whatever post deploy command at cEOS 60 times without any possibility to detect syntax errors, unsupported commands, etc. While I'd expect this line to be healthy in general and not have all of that, can we consider an alternative approach?

1) loop 60 times to detect readiness with some simple CLI
2) once CLI is ready - stop looping and send post deploy once
3) Return any configuration error immediately with stderr

nodes/ceos/ceos.go
```diff
@@ -410,18 +413,40 @@ func (n *ceos) ceosPostDeploy(_ context.Context) error {
...
+		execCmd := clabexec.NewExecCmdFromSlice([]string{"/bin/bash", "-lc", cliCmd})
+		resp, err := ceosPostDeployExec(n, ctx, execCmd)
+		if err == nil && resp.GetReturnCode() == 0 {
+			return nil
+		}
+
+		lastErr = err
+		lastResp = resp
+		log.Debugf("%s - Cli not ready (%v, %v) - waiting.", nodeCfg.LongName, err, resp)
+		ceosPostDeploySleep(2 * time.Second)
```
ankudinov: have to test that in a live lab, but I suspect cancellation will not stop this loop. Please check and fix if it's the case. I'll run this fork in my lab during the final review.

ankudinov (changes requested):
@orrious Thank you for adding the tests! This PR deserves a better description explaining the problem and what it's changing. And this description must be correctly reflected in docs. Even after reading the recent doc update it is still not clear what this PR solves.

Here is the AI generated summary at a cost of few tokens burned. If you agree - please rewrite and add this to the doc in a shortened from:

```text
Existing lifecycle:
deploy → cEOS PostDeploy() → configure interfaces/IPs

Before:
Scrapligo → spawn `docker/podman exec -it ... Cli` subprocess
            → interactive EOS CLI session
After:
cEOS PostDeploy() → containerlab Runtime.Exec()
                  → run EOS `Cli -c ...` as one batch command
Why do it?
Use containerlab’s native Docker/Podman runtime connection instead of spawning a host-side CLI binary.
Avoid interactive TTY/Scrapligo behavior.
Capture return code, stdout and stderr properly.
Retry CLI startup up to 60 times, rather than the old unbounded retry.
Make execution more consistent with runtime configuration, particularly Podman.
It also changes the EOS command sequence slightly:
Adds configure terminal.
Uses --abort-on-error.
Ends with end and write memory.
Adds unit and integration tests confirming management/data-interface addresses are applied.
My take: the title is misleading. This is a cEOS-specific transport refactor/robustness change, not a new post-deploy feature. Furthermore, the old implementation already selected either the docker or podman executable, so the PR’s practical motivation is not explained especially well. There is no PR description or linked bug.
```

I'd also explain a bit more about post-deploy in the doc. However this isn't directly related to this PR and can be skipped for now.

nodes/ceos/ceos.go
orrious: Addressed in 97a7cacf0. `Runtime.Exec` now invokes `Cli` directly with an argv slice, with the multiline configuration passed as a single `-c` argument. The unit test verifies the exact argument boundaries.

nodes/ceos/ceos.go
orrious: Addressed in b406f8441. Post-deploy now retries a non-mutating `show version` readiness probe, then submits the configuration exactly once. Runtime failures and nonzero CLI results are returned immediately with return code, stdout, and stderr. Tests cover readiness retries and the single configuration attempt.

nodes/ceos/ceos.go
orrious: Addressed in 72e447ecb. The readiness loop checks the context before each attempt and uses a context-aware timer between attempts. Cancellation now returns immediately, and a regression test verifies that no additional exec attempt is made.

orrious:
@ankudinov The remaining implementation feedback from the August 14 review is now addressed in three focused commits:

- 97a7cacf0 invokes `Cli` directly without a Bash wrapper
- b406f8441 separates CLI readiness probing from the single configuration attempt
- 72e447ecb makes readiness retries honor context cancellation

The PR title, description, and cEOS documentation were also updated in 02d526cc4. Validation: `go test ./nodes/ceos` passed under Linux Go 1.26.5. I replied to each inline discussion with the corresponding details. Please take another look when you have a moment.

PR #3283, opened 2026-07-15
Add targeted node filtering to deploy reconciliation

orrious:
## Goal and use case

Allow part of an already-running lab to be reconciled without disrupting unrelated nodes.

A typical use case is a large lab where one device under test or traffic-generator node must be added or recreated while preserving the control-plane state, container IDs, links, and runtime state of the rest of the topology. The existing `--node-filter` flag already identifies the intended target set; this change applies that selection safely during reconciliation.

## Behavior

- allow `clab deploy --node-filter` and the `clab apply` alias to add or recreate selected nodes
- compute the dependency closure for `wait-for` and `network-mode: container:<provider>`
- preserve nodes, endpoints, and links outside the selected closure
- include links required by selected added or recreated nodes, including links to existing unselected peers
- normalize endpoint ownership for shared network namespaces so only the provider owns discovered interfaces
- checkpoint filtered state without persisting runtime-injected environment drift
- preserve filtered-added nodes for later destroy

The branch has been rebased onto current `main` and now contains only the filtered-apply implementation and its follow-up fixes; unrelated stacked PR commits have been removed.

## Validation

- Linux Go 1.26.5: `go test ./core ./types ./nodes`
- Linux Go 1.26.5 with Podman build tags: `go test ./runtime/podman`
- strict Zensical documentation build: `make docs`
- existing unit and smoke coverage exercises dependency closure, cross-filter links, shared namespaces, filtered state, idempotent dry runs, and later destroy


orrious:
Rebased this branch onto current `main`, resolved the rebase conflicts, and force-pushed the updated head.

Current head: `7606ce097`

During runner validation, the rebase exposed a failing unit test in `TestDiscoverLiveApplyEndpointsRejectsStoppedExternalNode`. I fixed that test-only regression on top of the rebased branch and reran validation.

Final validation completed on the real lab runners:
- x86 runner: `go test ./core ./types ./nodes` passed
- ARM runner: `go test ./core ./types ./nodes` passed

hellt:
@orrious what is the goal of this PR? can you explain the use case where you would need this feature?

FloSch62:
@orrious clab deploy/apply should anyhow only touch nodes that needs a recreate/restart. So the node-filter will only introduce breaking nodes/links. The thing is, some kinds need a recreate to get new or deleted interfaces on the remote link. So I am not sure what the PR will solve. But pls let me know if I miss the point.

jbemmel:
@orrious #3371 may be of interest, it's probably narrower than what you had in mind but it fits the same theme

PR #3284, opened 2026-07-16
Unified and improved error check

kaelemc:
Strengthen the validate command to not just return the 1st error. Instead we can now collect all errors and display them to the user/report them with validate cmd.

Instead to just add this to the validate command, this is all unified to deploy/apply etc. should get the same error checking flow.


kaelemc:
Stale.... will open new PR if I find time to proceed with this.

PR #3286, opened 2026-07-17
nodes/srsim: Add ED25519 key support for SR-SIM 26.7 and higher

vista-:
ED25519 SSH keys are now supported in SR-OS 26.7.R1 (and newer). This PR adds support for automatically adding these type of SSH keys.


vista-:
Tested with SR-SIM 26.7.R1 and 25.10.R1 with both MD-CLI and Classic CLI.

kaelemc (approved):

PR #3300, opened 2026-08-05
fix: make bind resolution deterministic

### LINKED ISSUE #3299: `clab apply` reports false `config drift: Binds` and recreates unchanged nodes when a node has ≥2 binds
## Summary

`containerlab apply` intermittently plans **recreate** for nodes whose topology definition is unchanged, with reason:

```text
config drift: Binds
```

This happens whenever a node has **two or more** bind mounts. The bind *set* is identical between `.state.clab.yaml` and the current topology; only the **slice order** differs across calls. Unrelated nodes (e.g. FRR routers elsewhere in a Clos fabric) get destroyed and recreated on apply, which resets BGP and dataplane state far beyond the node that actually changed.

## Environment

| Item | Value |
|------|--------|
| containerlab | **v0.77.0** |
| Runtime | Docker |
| OS | Linux (also reproduced inside DinD) |
| Command | `containerlab apply -t clab.yml` / `--dry-run` |

## Impact

In labs that use the common pattern of **two binds per linux node** (daemon config + bootstrap script), every `apply` that touches *any* node has a high chance of also recreating a **random subset** of other multi-bind nodes:

```text
│ recreated nodes │ L2-1-2 (config drift: Binds) │
│ recreated nodes │ S3-2 (config drift: Binds)   │
│ recreated nodes │ SS1 (config drift: Binds)    │   ← superspine, unrelated to the edited node
│ recreated nodes │ SS2 (config drift: Binds)    │
```

That makes `apply` unsafe for incremental workflows (device swap, single-node kind/image change, link edits): control-plane sessions on untouched nodes flap.

With **N** nodes that each have **K≥2** binds, false positives are common because bind order is effectively random per `GetNodeBinds` call (see root cause).

### Minimal topology

```yaml
name: binds-drift
prefix: ""

topology:
  nodes:
    n1:
      kind: linux
      image: alpine:3.19
      binds:
        - n1.conf:/etc/n1.conf
        - n1-init.sh:/etc/n1-init.sh
      entrypoint: /bin/sh
      cmd: sleep infinity
    n2:
      kind: linux
      image: alpine:3.19
      binds:
        - n2.conf:/etc/n2.conf
        - n2-init.sh:/etc/n2-init.sh
      entrypoint: /bin/sh
      cmd: sleep infinity
```

Create empty bind sources next to the topology file, deploy once, then repeatedly dry-run apply **without editing the topology**:

```bash
touch n1.conf n1-init.sh n2.conf n2-init.sh
containerlab deploy -t topo.clab.yml
# Repeat; some runs are clean, some recreate n1 and/or n2 for Binds drift:
for i in $(seq 1 20); do
  echo "=== run $i ==="
  containerlab apply -t topo.clab.yml --dry-run
done
```

### Observed (unchanged topo)

Some iterations:

```text
INFO Apply plan
╭────────────┬─────────╮
│   Action   │ Details │
├────────────┼─────────┤
│ no changes │ -       │
╰────────────┴─────────╯
```

Other iterations (same files on disk):

```text
INFO Apply plan
╭─────────────────┬──────────────────────────────╮
│      Action     │            Details           │
├─────────────────┼──────────────────────────────┤
│ recreated nodes │ n1 (config drift: Binds)     │
│ recreated nodes │ n2 (config drift: Binds)     │
╰─────────────────┴──────────────────────────────╯
```

Diffing `.state.clab.yaml` vs `topo.clab.yml` for those nodes shows **identical** `binds:` lists in YAML. The drift is introduced after `GetNodeBinds` turns that list into a slice.

### Contrast: single bind

If each node has **exactly one** bind, repeated `--dry-run` stays at `no changes`. Collapsing two file binds into one directory bind is an effective workaround (what we did in our lab generator).

## Expected behavior

If the multiset of bind mount specs for a node is unchanged (same sources/destinations/options), `apply` must **not** report `config drift: Binds` and must **not** recreate the node.

## Actual behavior

`apply` intermittently treats bind order as a definition change and recreates the node.

## Root cause

Apply reconciliation compares “old” vs “new” node config built via `resolveNodeConfigFromTopology` → `Topology.GetNodeBinds`:

https://github.com/srl-labs/containerlab/blob/v0.77.0/core/topology_reconcile.go (around `resolveNodeConfigFromTopology` / `planNodeReconciliation`)

`GetNodeBinds` correctly dedupes by destination using a `map[string]*Bind`, then **materializes the result by ranging over that map**:

```go
// types/topology.go — GetNodeBinds
binds := map[string]*Bind{}
// ... populate by destination ...
result := make([]string, 0, len(binds))
for _, b := range binds {          // ← Go map iteration order is randomized
    result = append(result, b.String())
}
return result, nil
```

`DefaultNode.ComputeDiff` then compares bind slices with an **order-sensitive** helper:

```go
// nodes/default_node.go — ComputeDiff
if !clabutils.SlicesEqualOrBothEmpty(oldCfg.Binds, newCfg.Binds) {
    diff.Fields = append(diff.Fields, "Binds")
}
```

```go
// utils/diff.go — SlicesEqualOrBothEmpty
for i := range a {
    if a[i] != b[i] {   // ← positional equality
        return false
    }
}
```

So for the same two binds `{A, B}`:

1. Old topo `GetNodeBinds` may return `[A, B]`
2. New topo `GetNodeBinds` may return `[B, A]`
3. Diff → `Binds` → recreate

YAML/`state` order is stable; the non-determinism is entirely in map iteration when building the compare slice. That is why dry-runs flip between “no changes” and “recreate” with **zero** topology edits.

`Binds` is listed as a recreate-triggering field in `types/topo_diff.go`, so this false positive is destructive, not a soft restart.

## Suggested fix

Any of these would work (preferably more than one):

1. **Stable sort** when returning from `GetNodeBinds` (e.g. by destination path, then by full bind string).
2. **Order-insensitive bind compare** in `ComputeDiff` (sort copies, or compare as a set keyed by `Dst()` / normalized bind string).
3. Preserve **authoritative declaration order** from the merged source list instead of map iteration when emitting the slice (map only for override-by-destination).

A unit test that calls `GetNodeBinds` twice (or `ComputeDiff` on two configs built from the same topology with ≥2 binds) in a loop and asserts no `Binds` field would lock this down. Under the Go race/random map iteration, flaky failures today would become deterministic failures until sorted.

## Workaround

Use **at most one bind entry per node** (e.g. bind a whole `config/` directory once and symlink/copy files inside the entrypoint), or avoid `apply` and use full `redeploy` (still painful operationally).

## Related

- `clab apply` node definition reconciliation / `.state.clab.yaml` baseline  
- Docs: changes to `binds` intentionally recreate — but here `binds` did not change in the topology file

Happy to send a PR that sorts `GetNodeBinds` output and adds a regression test if that approach sounds right.

FloSch62:
## What changed

- sort resolved bind mounts by destination path before returning them from `GetNodeBinds`
- add a regression test that repeatedly verifies deterministic multi-bind ordering

## Why

`GetNodeBinds` deduplicates bind mounts in a map and previously returned them in randomized map iteration order. Apply reconciliation compares the resulting slices positionally, so unchanged nodes with multiple binds could be reported as `config drift: Binds` and unnecessarily recreated.

This keeps unchanged multi-bind nodes stable for both `containerlab deploy` and its `apply` alias while preserving detection of real bind changes.

Fixes #3299.

## Validation

- `make test` (race detection and coverage)
- `make build`
- bind regression tests repeated across 100 test runs
- deployed two real Alpine nodes with two bind mounts each
- 100 unchanged dry-runs and 20 unchanged real applies all reported no changes without replacing either container
- a controlled environment change recreated only the intended node; the untouched multi-bind node retained its container ID
- 50 post-change dry-runs all reported no changes

### DIFF
--- a/types/topology.go
+++ b/types/topology.go
@@ -771,17 +771,11 @@ func (t *Topology) GetNodeBinds(nodeName string) ([]string, error) {
 		return nil, nil
 	}
 
-	// Sort by destination path so map iteration cannot introduce config drift.
-	destinations := make([]string, 0, len(binds))
-	for dst := range binds {
-		destinations = append(destinations, dst)
-	}
-	slices.Sort(destinations)
-
 	result := make([]string, 0, len(binds))
-	for _, dst := range destinations {
-		result = append(result, binds[dst].String())
+	for _, b := range binds {
+		result = append(result, b.String())
 	}
+	slices.Sort(result)
 
 	return result, nil
 }
--- a/types/topology.go
+++ b/types/topology.go
@@ -772,9 +772,11 @@ func (t *Topology) GetNodeBinds(nodeName string) ([]string, error) {
 	}
 
 	result := make([]string, 0, len(binds))
+
 	for _, b := range binds {
 		result = append(result, b.String())
 	}
+
 	slices.Sort(result)
 
 	return result, nil
--- a/types/topology.go
+++ b/types/topology.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"slices"
 	"strings"
 
 	"github.com/docker/go-connections/nat"
@@ -770,11 +771,16 @@ func (t *Topology) GetNodeBinds(nodeName string) ([]string, error) {
 		return nil, nil
 	}
 
-	// build the result array with all the entries from binds map
-	result := make([]string, 0, len(binds))
+	// Sort by destination path so map iteration cannot introduce config drift.
+	destinations := make([]string, 0, len(binds))
+	for dst := range binds {
+		destinations = append(destinations, dst)
+	}
+	slices.Sort(destinations)
 
-	for _, b := range binds {
-		result = append(result, b.String())
+	result := make([]string, 0, len(binds))
+	for _, dst := range destinations {
+		result = append(result, binds[dst].String())
 	}
 
 	return result, nil
--- a/types/topology_test.go
+++ b/types/topology_test.go
@@ -896,6 +896,39 @@ func TestGetNodeBinds(t *testing.T) {
 	}
 }
 
+func TestGetNodeBindsDeterministicOrder(t *testing.T) {
+	topology := &Topology{
+		Defaults: &NodeDefinition{},
+		Nodes: map[string]*NodeDefinition{
+			"node1": {
+				Binds: []string{
+					"z-source:/etc/z.conf",
+					"a-source:/etc/a.conf",
+					"m-source:/etc/m.conf",
+				},
+			},
+		},
+	}
+	want := []string{
+		"a-source:/etc/a.conf",
+		"m-source:/etc/m.conf",
+		"z-source:/etc/z.conf",
+	}
+
+	for range 100 {
+		got, err := topology.GetNodeBinds("node1")
+		if err != nil {
+			t.Fatalf("GetNodeBinds() unexpected error: %v", err)
+		}
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Fatalf(
+				"GetNodeBinds() returned binds in a non-deterministic order (-want +got):\n%s",
+				diff,
+			)
+		}
+	}
+}
+
 func TestGetNodeEnv(t *testing.T) {
 	for name, item := range topologyTestSet {
 		t.Logf("%q test item", name)


hellt:
thanks @FloSch62 
I removed the extra slice, which should not impact the scope of the feature

PR #3312, opened 2026-08-11
core/topology_reconcile: Fix deployment and reconciliation of bridges

vista-:
At the moment, deploying any topology with bridges will fail due to incorrectly handling them as containers.

Test topo:
```
name: test
topology:
  nodes:
    leaf1:
      kind: nokia_srlinux
      image: ghcr.io/nokia/srlinux:26.7.1
    br-br100:
      kind: bridge
  links:
    - endpoints: ["leaf1:e1-1", "br-br100:e2"]
    - endpoints: ["leaf1:e1-2", "br-br100:e3"]
```

Created bridge before deployment.

```
➜ clab deploy
14:57:18 INFO Containerlab started version=0.78.0
14:57:18 INFO Parsing & checking topology file=test.clab.yaml

   ERROR

  Externally managed node "br-br100" is NotFound; start it outside containerlab before running apply.
```
  
This PR fixes at least the initial deployment and basic reconciliation of bridges. There is still some wonkyness involved, e.g. deploy the topology listed above, and then change the links like so, and deploy it again:

```
  link:
                                    # was e2
    - endpoints: ["leaf1:e1-1", "br-br100:e1"]
                                    # was e3
    - endpoints: ["leaf1:e1-2", "br-br100:e2"]
```

and this will result in e2 getting deleted and only e1 getting deployed:

```
15:15:12 INFO Apply summary
╭───────────────────┬───────────────────────────╮
│       Action      │          Details          │
├───────────────────┼───────────────────────────┤
│ added links       │ br-br100:e1 -- leaf1:e1-1 │
│ deleted endpoints │ br-br100:e3               │
╰───────────────────┴───────────────────────────╯
➜ ip -br l | grep -P "e(1|2)"
e1@if106         UP             aa:c1:ab:77:c4:97 <BROADCAST,MULTICAST,UP,LOWER_UP>
```

Running a deploy again will make Clab realize that the link is missing ;)

```
15:16:19 INFO Apply summary
╭─────────────┬───────────────────────────╮
│    Action   │          Details          │
├─────────────┼───────────────────────────┤
│ added links │ br-br100:e2 -- leaf1:e1-2 │
╰─────────────┴───────────────────────────╯
```

It might be a more complex fix - I would like to get this one out of the door to allow people with any bridges in their topologies to get productive again.

PR #3314, opened 2026-08-12
fix: reconcile host bridge links in one pass

hellt:
substitutes #3312 

## Problem

Labs containing topology `kind: bridge` nodes could deploy successfully but fail on a subsequent `deploy`/`apply`. Root-namespace bridges were classified as external containers, so reconciliation queried a nonexistent container and rejected the lab. Even after bypassing that failure, endpoint-only reconciliation considered reused interface names live despite their veth peers having changed; deleting a stale endpoint could then remove a peer needed by another desired link.

## Changes

- Model root-namespace resources separately from externally managed containers and avoid container lifecycle/status checks for them.
- Scope interface ownership to the exact logical `(node, interface)` marker, preventing host-namespace bridge/host/OVS nodes from cross-attributing interfaces.
- Keep fresh-deployment collision checks while allowing existing endpoints during reconciliation, with an explicit unowned-interface guard before mutations.
- Discover veth peer and bridge-master metadata and require desired two-ended links to match their live peer pairing before preserving them.
- Reject unsupported `network-mode: container:...` reconciliation explicitly.
- Pass link-verification parameters explicitly through one verification method, keeping fresh deployment and apply policies clear.
- Separate namespace identity from logical ownership identity so interfaces parked during SR-SIM/node restarts remain discoverable and owned by their original node.
- Add a new gating smoke suite that deploys bridge ports `e2/e3`, remaps them to `e1/e2` in one deploy, and verifies the second deploy is a no-op. Existing bridge topologies were not changed.

## Validation

- Baseline bridge regression fails before the fix during reconciliation.
- New `30-bridge-reconcile.robot`: 3/3 passed.
- Existing bridge suites `03*.robot` and `24*.robot`: passed.
- Existing apply suite `29*.robot`: 10/10 passed.
- SRSIM apply suite `11-srsim-apply.robot`: 5/5 passed after the parked-namespace ownership fix.
- `make test`: passed.
- `make lint`: passed.
- Focused follow-up tests: `go test ./core ./links` passed.

### DIFF
--- a/.cursor/rules/git-mutations.mdc
+++ b/.cursor/rules/git-mutations.mdc
@@ -1,0 +1,10 @@
+---
+description: Require explicit user authorization for Git mutations
+alwaysApply: true
+---
+
+# Git mutation policy
+
+- Never automatically commit, amend, push, force-push, or create/update a pull request.
+- Leave ordinary fixes as working-tree changes unless the user explicitly requests the specific Git or GitHub action.
+- A request to implement or fix code alone does not authorize a commit or push.
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -53,6 +53,7 @@ jobs:
           - "27*.robot"
           - "28*.robot"
           - "29*.robot"
+          - "30*.robot"
 
     # allow podman job to fail, since it started to fail on github actions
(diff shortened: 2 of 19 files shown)


hellt:
@vista- can you build from this PR and check if your labs behave correctly?

PR #3319, opened 2026-08-13
fix veth-stitch race handling and return stitch interfaces in `clab events`

kaelemc:


hellt:
## PR analysis

This PR fixes two related veth-stitch problems.

### 1. Deployment-time netlink races

Parallel link deployment and stitch setup can hit transient kernel errors while veth peers, network namespaces, or VXLAN devices are becoming visible. The stitched link could also reach post-deploy before both veth segments were ready, leading to intermittent failures.

The fix:

- Centralizes transient netlink retry handling for `EFAULT`, `ENODEV`, and `ENOENT`, with bounded exponential backoff and context cancellation.
- Reuses that retry path for regular veth deployment, veth-stitch setup, and VXLAN-stitch post-deploy.
- Makes veth-stitch lifecycle handling serialized and idempotent, and stitches immediately once both segments are fully deployed; post-deploy remains a safe fallback.
- Runs link post-deploy work with the configured worker limit instead of serially.
- Configures nodeless veth endpoints in the current namespace rather than trying to move them through a container node.

### 2. Host-side stitch interfaces were not reported consistently

For `veth-stitch`, traffic, impairment, and inspection belong to the host-side interface (`clab-stitch-*`), while the container-side interface is the topology-facing endpoint. `inspect interfaces` already had an overlay path, but the lookup was duplicated and `clab events` did not retain enough node identity to resolve the host-side interface.

The fix:

- Adds one `ToolsInterface` lookup shared by interface inspection, netem targeting, and event reporting.
- Preserves the topology alias while replacing reported name, index, MTU, state, type, and MAC with the host-side stitch interface.
- Carries node and root-node labels through container event reconstruction so multi-component nodes resolve to the correct topology node.
- Mirrors the node-side MAC onto the stitched far end.
- Moves the common TC stitch implementation out of the VXLAN-specific file so both stitch types use the same code.

The added tests cover transient retry behavior, nodeless veth setup, stitch-interface identity guards, MAC propagation, and `clab events --initial-state` reporting. Focused local validation passes with `go test ./links ./core/events`.

PR #518, opened 2021-07-16
fixing cvx lab suffixes

networkop:
fixes #516

PR #3330, opened 2026-08-19
Standardize VM-based node resource configuration on QEMU_SMP/QEMU_MEMORY

### LINKED ISSUE #3318: Standardize setting cpu and memory for cisco xrv9k
For some reason (probably legacy) cisco xrv9k documents setting the cpu and memory for its vrnetlab in a different way than the other vrnetlab nodes.

This leads to the docs for it having a different documented way to set them.

It seems to me that the default behavior is defined in the `common/vrnetlab.py` file in the vrnetlab project, which 
[qemu args setting](https://github.com/srl-labs/vrnetlab/blob/9dbcd465a75f7c2d048bd3342eac10c3537eb00d/common/vrnetlab.py#L399) hopefully xrv9k uses the same common functions.

Then instead of passing vcpu/ram as [cli args](https://github.com/srl-labs/containerlab/blob/cb6828d830204d1ff56638e17719e101ea5fceef/nodes/vr_xrv9k/vr-xrv9k.go#L92) it should be possible to set those same values as `QEMU_MEMORY`, and `QEMU_SMP` env vars. This retains backward compatibility and hopefully makes it easier to drop this patch in the future so that they all use the same method.

hellt:
fix #3318


hellt:
## Rationale and relationship to vrnetlab #518

This PR is the containerlab side of [issue #3318](https://github.com/srl-labs/containerlab/issues/3318). The inconsistency came from some VM kinds using legacy `VCPU`/`RAM` environment variables or passing `--vcpu`/`--ram` to launchers, while vrnetlab’s common `VM` class already provides the standardized `QEMU_*` override path.

The two PRs are intentionally coupled:

1. [vrnetlab #518](https://github.com/srl-labs/vrnetlab/pull/518) makes the base `VM.ram`, `VM.cpu`, and `VM.smp` properties authoritative for `QEMU_MEMORY`, `QEMU_CPU`, and `QEMU_SMP`. It also removes the SR OS property bypass and the duplicate hard-coded SONiC `-smp` arguments.
2. This PR updates the remaining containerlab VM kinds to inject `QEMU_SMP`/`QEMU_MEMORY`, removes containerlab-generated legacy resource CLI arguments for XRv9k, Cat9Kv, and XRd, and aligns the documentation and regression coverage.

User-provided environment values still override kind defaults. The vrnetlab launcher CLI flags remain available for direct-launcher backward compatibility, while containerlab’s generic `cpu`/`memory` topology fields remain container cgroup limits and are not implicitly translated into guest QEMU resources.

For the standardized behavior to be effective in deployed images, the corresponding vrnetlab changes from [#518](https://github.com/srl-labs/vrnetlab/pull/518) need to land in the vrnetlab image build used by containerlab.

PR #3353, opened 2026-08-23
Fix setuid iptables DOCKER-USER rules on legacy backends

jbemmel:
# What is broken

Containerlab can be installed setuid-root for clab_admins. In that mode the process keeps the invoking user’s real UID (ruid) and elevates only the effective UID (euid=0). Privilege work runs as root; later file ownership under $HOME still follows the real user.

For mgmt external-access, if the nftables path isn’t usable (e.g. Docker on iptables-legacy, DOCKER-USER only in legacy iptables), containerlab shells out to iptables / ip6tables to add/remove DOCKER-USER ACCEPT rules for the mgmt bridge.

From iptables 1.8.8, libxtables refuses to run when getuid() != geteuid(). It loads match/target plugins via dlopen and trusts env like LD_*, so running under a setuid parent is treated as unsafe. It exits 111 with little/no message. Containerlab logged iptables check error: exit status 111 and skipped the rules — so external access failed for setuid users. sudo containerlab worked because both UIDs are 0.

# What this PR changes

Only the iptables CLI spawn path:

Matching root UIDs in the child — via SysProcAttr.Credential so the child gets Uid: 0 (Linux setuid(0) as euid-root sets real/effective/saved). Parent keeps ruid=user.
Strip LD_* from the child’s env so loader overrides don’t ride along once UIDs match.
Nftables is untouched


jbemmel:
This can probably be simplified to not pass any env variables at all

PR #3364, opened 2026-08-27
fix podman client context when copying config to container

WhateverAWS:
image

I have found this bug, when running containerlab with podman. 

The configuration failed to be copied to the container due to missing context connection (`ctx`) on `CopyToContainer` function in podman runtime which caused another function (`CopyFromArchiveWithOptions`) invoked inside to fail with `Client not set in context`.


hellt:
thanks @WhateverAWS

PR #3371, opened 2026-08-31
Feat/preserve links on node destroy

jbemmel:
## Summary

Add `destroy --keep-links` support for filtered node replacement and preserve dataplane veths across subsequent `apply` operations, including swaps where the replacement node uses different interface names.

## Motivation

Replacing one node previously recreated its dataplane veth pairs. This removed the surviving peer interfaces and discarded kernel-only state such as IP addresses, neighbor state, and other runtime configuration.

For example, swapping a DUT from Nokia SR Linux to Arista cEOS changes interface names from `e1-1`/`e1-2` to `et1`/`et2`. The identity of a preserved connection is the veth pair—not the old interface name. Recreating that pair breaks peers such as BIRD unless their link initialization is rerun, defeating the purpose of keeping links.

## Workflow
```
destroy --keep-links --node-filter dut
  └─ validate options
  └─ load requested topology
  └─ ParkEndpoints(dut)
       └─ discover owned DUT endpoints
       └─ move each endpoint into clab-park-dut
  └─ remove DUT container
  └─ leave peer endpoints untouched

containerlab apply
  └─ build reconciliation plan
       └─ detect missing DUT + parking namespace
       └─ classify DUT as added + parked
       └─ discover parked and live endpoints
       └─ match links by veth Index/PeerIndex
       └─ suppress stale endpoint removal
       └─ decide whether links need deployment
  └─ create replacement DUT container
  └─ RestoreEndpoints(dut)
       └─ match parked endpoint through surviving peer
       └─ rename endpoint to desired interface name
       └─ update ownership altname
       └─ move endpoint into replacement container
       └─ set link mode default and bring interface up
  └─ deploy only genuinely missing links
  └─ run node post-deploy
```

## Improvements

- Add `--keep-links` to filtered `destroy`.
- Move selected node endpoints into a deterministic parking network namespace before removing the container.
- Preserve the peer side of each veth, including kernel addresses and runtime state.
- Discover parked endpoints during apply even when the original container no longer exists.
- Match renamed endpoints by veth `Index`/`PeerIndex` identity rather than requiring an exact interface-name match.
- Rename parked endpoints to the replacement topology’s desired interface names before restoring them.
- Update ownership altnames after an interface rename.
- Prevent parked endpoints from being treated as stale solely because their desired names changed.
- Handle stale runtime entries where Docker reports the original container as `NotFound`.
- Use the explicitly requested topology during filtered destroy when an immutable topology label references a deleted temporary apply file.
- Restore interfaces using the default Linux link mode before bringing them up, clearing dormant mode left by network operating systems such as cEOS.

## Expected behavior

A sequence such as:

```text
SR Linux DUT → cEOS DUT → FRR DUT
```

preserves the original dataplane veth pairs throughout. Apply may log interface renames, but it does not log new link creation for the kept adjacencies.

Peer interfaces retain addresses that exist only in the kernel, so peer link-initialization does not need to run again.

## Testing

Added coverage for:

- CLI and destroy-option handling.
- Parking links during filtered destroy.
- Matching parked endpoints across interface renames.
- Avoiding unnecessary link deployment when a parked endpoint remains paired with the live peer.
- Stale runtime entries with a valid parking namespace.
- Explicit topology fallback during filtered destroy.
- Restoring interfaces in default link mode.

Validated with:

```bash
go test ./core ./links
```

https://github.com/user-attachments/assets/16869204-506a-4046-a1e4-4a529884125a

### DIFF
--- a/cmd/destroy.go
+++ b/cmd/destroy.go
@@ -69,6 +69,12 @@ func destroyCmd(o *Options) (*cobra.Command, error) {
 		o.Destroy.KeepManagementNetwork,
 		"do not remove the management network",
 	)
+	c.Flags().BoolVar(
+		&o.Destroy.KeepLinks,
+		"keep-links",
+		o.Destroy.KeepLinks,
+		"preserve data-plane links when destroying nodes selected with --node-filter",
+	)
 	c.Flags().StringSliceVarP(
 		&o.Filter.NodeFilter,
 		"node-filter",
@@ -81,6 +87,14 @@ func destroyCmd(o *Options) (*cobra.Command, error) {
 }
 
 func destroyFn(cobraCmd *cobra.Command, o *Options) error {
+	if o.Destroy.KeepLinks && len(o.Filter.NodeFilter) == 0 {
+		return fmt.Errorf("keep-links requires node-filter")
+	}
+
+	if o.Destroy.KeepLinks && o.Destroy.Cleanup {
+		return fmt.Errorf("keep-links cannot be used with cleanup")
+	}
+
 	if o.Destroy.Cleanup && len(o.Filter.NodeFilter) != 0 {
 		return fmt.Errorf("cleanup cannot be used with node-filter")
 	}
--- a/cmd/destroy_test.go
+++ b/cmd/destroy_test.go
@@ -1,0 +1,44 @@
+package cmd
+
+import (
+	"context"
+	"strings"
+	"testing"
+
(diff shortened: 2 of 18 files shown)


jbemmel:
Without this PR, what fails is the ip address on BIRD's uplink. It can be fixed of course, but preserving the peer links is way more elegant and preferable. Most of the namespace parking logic is already in place, this PR simply applies it

PR #3382, opened 2026-09-03
feat(nodes): add FRRouting (frr) kind

mwinter-osr:
Adding a FRRouting (more native) kind. 
The change is based on a new docker container specifically for containerlab which is based on the normal frr container. The new docker container is maintained by the frrouting community and will be updated on each FRR release. First version is the current 10.7.1 release.

Disclaimer: I'm a FRRouting maintainer, but my containerlab and go foo is limited. As such, I had to use quite a bit of AI
to help with this PR. I've reviewed it on my own and it all verifies ok and looks ok. But I suggest a good review and let
me know if you need changes.

FRR could already be run as a linux node, but every topology had to hand-write bind mounts for /etc/frr/frr.conf and /etc/frr/daemons, and the resulting containers were not reachable over SSH.

The frr kind writes frr.conf, daemons and vtysh.conf into the node's lab directory and mounts them over the container's /etc/frr. All three are always written: the official image ships only daemons, and vtysh refuses to start without the other two. frr.conf is rendered through GenerateConfig, so startup-config, enforce-startup-config and suppress-startup-config behave as they do for other kinds. SaveConfig writes "show running-config" back to the lab directory.

Which daemons run is set with extras.frr.daemons. Omitting it starts every daemon, so a lab works without the user having to know which daemons their config needs; naming daemons starts exactly those. The daemons file is generated from the one FRR 10.7.1 ships, keeping the per-daemon *_options lines that carry the VTY bind addresses.

The public keys containerlab collects from the host are written to the root user's authorized_keys. This needs an image with an SSH server, which the official quay.io/frrouting/frr image is not; ghcr.io/srl-labs/frr adds one on top of it (see srl-labs/frr-container).

The frr01 lab example moves to the new kind, dropping nine bind lines and three copies of the daemons file. Its documentation page was also stale: it referenced frr v7.5.1 and a run.sh that no longer exists.


hellt:
thanks, @mwinter-osr 
let me check what are the gaps on the docs side

hellt:
@mwinter-osr I have massaged the docs a little bit and added the frr tests to our public CI.

mwinter-osr:
> @mwinter-osr I have massaged the docs a little bit and added the frr tests to our public CI.

Thanks a lot. Let me know if it needs anything from my side.

PR #3383, opened 2026-09-04
apply: support network-mode: container: nodes

steiler:
## What
`apply` previously blanket-rejected any node with `network-mode: container:<target>` via `checkUnsupportedApplyNodes()`. That check was added defensively in an unrelated bridge PR and turns out to be unnecessary for the common case of a sidecar container with no clab-managed links sharing a target's netns (e.g. `network-mode: container:worker1`).

## Changes
- Delete `checkUnsupportedApplyNodes()` and its call site.
- `DeployNodes` now waits for a node's `network-mode: container:` target to be running from a per-node goroutine decoupled from the worker pool, so a dependent node queued ahead of its target (added/recreated in the same `apply`) can't deadlock a constrained worker pool (e.g. `--max-workers 1`). Mirrors fresh deploy's existing `waitForExternalNodeDependencies` for targets not managed by this topology.
- `planApply` now forces a node into `recreatedNodeSet` (via a fixed-point pass, so chains cascade) whenever its `network-mode: container:` target is itself recreated — Docker binds `NetworkMode` to the target's container ID at creation time, so a plain restart of the dependent can't rebind it after the target gets a new container ID.
- `planApply` rejects an apply that deletes a `network-mode: container:` target while its dependent node stays desired, naming both nodes in the error, to avoid silently leaving the dependent attached to a defunct network namespace.

## Testing
- New unit tests in `core/apply_test.go` / `core/deploy_test.go` cover the cascade (including a transitive A→B→C chain and excluding newly-added nodes), the deleted-target rejection, and the `DeployNodes` ordering fix (including a regression test that reproduces the deadlock against a deliberately-reverted naive implementation).
- New Robot Framework smoke suite `tests/01-smoke/31-apply-network-mode.robot` exercises against real Docker containers:
  1. adding a no-link sidecar to an already-running target
  2. target recreation cascading a sidecar recreate
  3. adding a target + its sidecar together in one `apply --max-workers 1` call
  4. `apply` rejecting removal of a target while its sidecar remains desired
- `make build`, `go vet`, `go test -race ./...`, and the existing `tests/01-smoke/29-apply.robot` suite all pass with no regressions.

### DIFF
--- a/core/apply.go
+++ b/core/apply.go
@@ -3,7 +3,6 @@ package core
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	clablinks "github.com/srl-labs/containerlab/links"
 )
@@ -123,10 +122,6 @@ func (c *CLab) apply(
 		}
 
 		return result, nil
-	}
-
-	if err := c.checkUnsupportedApplyNodes(); err != nil {
-		return nil, err
 	}
 
 	if err := c.setMgmtBridgeFromRuntime(currentNodes); err != nil {
@@ -258,21 +253,6 @@ func (c *CLab) checkApplyTopologyDefinition(ctx context.Context) error {
 	}
 
 	return c.verifyDuplicateAddresses()
-}
-
-func (c *CLab) checkUnsupportedApplyNodes() error {
-	for _, nodeName := range sortedNodeNames(c.Nodes) {
-		cfg := c.Nodes[nodeName].Config()
-		if cfg != nil && strings.HasPrefix(cfg.NetworkMode, "container:") {
-			return fmt.Errorf(
-				"apply does not support nodes with %q; use redeploy or "+
-					"deploy --reconfigure",
-				cfg.NetworkMode,
-			)
-		}
-	}
-
-	return nil
 }
 
 func (c *CLab) prepareApply(
--- a/core/apply.go
+++ b/core/apply.go
@@ -186,15 +186,11 @@ func (c *CLab) apply(
 		return nil, err
 	}
 
-	if err := c.DeployNodes(ctx, deployNodeNames, options.maxWorkers); err != nil {
+	if err := c.deployApplyNodes(ctx, plan, options.maxWorkers); err != nil {
 		return nil, err
 	}
 
 	if err := c.restoreRecreatedNodes(ctx, plan); err != nil {
-		return nil, err
-	}
-
-	if err := c.startStoppedNodes(ctx, plan); err != nil {
 		return nil, err
 	}
 
--- a/core/apply_test.go
+++ b/core/apply_test.go
@@ -809,6 +809,178 @@ func TestResolveNodeConfigFromTopologyRuntimeOptions(t *testing.T) {
 	}
 }
 
+func TestNetworkModeContainerTarget(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		networkMode string
+		want        string
+	}{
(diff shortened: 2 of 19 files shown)


hellt:
Added six follow-up commits, each with regression coverage:

- [66369b90e](https://github.com/srl-labs/containerlab/commit/66369b90e923421c759e2d0e35817ae5dd4e8bf2) — **Clean up every smoke-test node.** Teardown now renders the topology with vars that include all incrementally added nodes and asserts that no lab containers remain. Previously, teardown could report success while leaving three containers running because it rendered only the initial node set.

- [57d2b222f](https://github.com/srl-labs/containerlab/commit/57d2b222fb946b4cf483fd89c7e1345654480803) — **Cascade recreation after link planning.** Namespace dependents are marked for recreation after all link-triggered recreation decisions, with parking and start sets updated accordingly. Previously, adding a link to a target with `link-apply-mode: recreate` could recreate only the target and leave its sidecar attached to the deleted container's namespace. Planner and Docker tests cover this, including transitive dependents.

- [61c317408](https://github.com/srl-labs/containerlab/commit/61c317408c6c33c21ba54520b9d3ab27f415899e) — **Restart dependents after their targets.** Restart requirements propagate through namespace dependencies, and targets restart before their sidecars. A restart can replace the target's network namespace even though its container ID stays the same, so leaving the sidecar running breaks namespace sharing. Tests check ordering, handling of newly added/recreated dependents, and namespace sharing while preserving container IDs.

- [5f05ee502](https://github.com/srl-labs/containerlab/commit/5f05ee502e1d368c56f5c9786c13219fcbe189d9) — **Schedule stopped targets alongside node creation.** Starts and creates now participate in the same deployment operation. Previously, a new sidecar waited for its stopped target inside `DeployNodes`, while the target's start was scheduled after `DeployNodes` returned. Tests cover one worker, adding a sidecar to a stopped target, starting both containers, and rebinding an already-running sidecar when its target starts.

- [e70bad7e8](https://github.com/srl-labs/containerlab/commit/e70bad7e80f2a2afc1782ccd7dd348d3af08c7b4) — **Signal readiness instead of polling internal tasks.** Nodes created or started in the same operation notify their dependents directly, including on failure. Existing/external targets receive an immediate status check. This removes the unconditional three-second startup delay and avoids waiting for the 15-minute dependency timeout after a target fails. Tests cover immediate checks, chains with one worker, failure propagation, cancellation, and cycle rejection.

- [859daa6db](https://github.com/srl-labs/containerlab/commit/859daa6db71f381ba98ed29d421ddf50f8791a45) — **Make recreation propagation linear.** A reverse dependency map and queue replace repeated sorting and whole-topology scans; parsing the target also avoids allocating a split slice. This changes cascade propagation from worst-case `O(V² log V)` to `O(V + E)`. A regression test bounds configuration reads, and a benchmark covers long chains. For the synthetic 5,000-node chain, single-iteration measurements improved from **3.06 s / 811 MB cumulatively allocated** to **2.4 ms / 2.35 MB**.

Validation: the binary builds, `go test ./...` passes on the host, lint and formatting checks pass, and all **11 Docker smoke tests** pass. The isolated race run also passes with `go test -race ./... -skip '^TestVerifyLinks$'`; `TestVerifyLinks` needs Docker access and passed in the full host suite.

hellt:
ty @steiler check the use cases that you have against the tip of this branch. If all good I will merge it in

PR #3388, opened 2026-09-05
fix(iol): support management IP and SSH bootstrap for L2 images (#3384, #3385)

### LINKED ISSUE #3384: [cisco_iol] Management IP configuration fails on L2 IOL images due to missing no switchport
### Description
When deploying a Cisco IOL node using an **L2 (Switch) image**, Containerlab fails to configure the management IP and routing on `Ethernet0/0`. 

This happens because `Ethernet0/0` on L2 IOL images defaults to a Layer 2 switchport, which does not accept an IP address without the `no switchport` command. As a result, users cannot access the node via SSH/Telnet out of the box unless they manually inject a startup-config or attach to the console.

### Steps to Reproduce & Evidence
When attaching to the L2 IOL console right after `containerlab deploy`, we can see the following configuration errors in the Cisco IOS-XE logs:

```text
Switch(config)#interface Ethernet0/0
Switch(config-if)#ip address 172.20.20.5 255.255.255.0
                       ^
% Invalid input detected at '^' marker.

Switch(config-if)#no ipv6 address
                          ^
% Invalid input detected at '^' marker.

Switch(config-if)#exit
Switch(config)#ip route vrf clab-mgmt 0.0.0.0 0.0.0.0 Ethernet0/0 172.20.20.1
% Must specify a L3 port as the next hop interface
```

### Root Cause
Containerlab generates a bootstrap configuration for `cisco_iol` nodes assuming they are all Layer 3 routers. It misses executing `no switchport` under `interface Ethernet0/0` for L2 variants.

### Proposed Solution
We should update the `cisco_iol` node definition/template to include `no switchport` for the management interface. 

Since executing `no switchport` on an already-L3 interface (like L3 IOL routers) is harmless and simply ignored by Cisco IOS, we could either:
1. Unconditionally add `no switchport` to `interface Ethernet0/0` in the default bootstrap config for `cisco_iol`.
2. Or, condition it based on the node's `type` field (e.g., if `type: l2`).

I would love to contribute and open a Pull Request to fix this if the maintainers agree with this approach! Let me know what you think.

### LINKED ISSUE #3385: [cisco_iol] SSH login fails with "Permission denied" due to missing AAA / VTY local authentication bootstrap
### Description
When deploying a Cisco IOL node, Containerlab provisions management user credentials, but SSH login results in a `Permission denied` error due to missing local authentication or AAA configuration on VTY lines.

### Steps to Reproduce & Evidence
1. Deploy a `cisco_iol` node with configured credentials.
2. Attempt to SSH into the node, which fails even with correct credentials. Console checks reveal `aaa new-model` and VTY `login` configurations are missing.

### Proposed Solution
Update the bootstrap command sequence in `UpdateMgmtIntf` within `nodes/iol/iol.go` to include standard AAA/VTY local authentication commands (e.g., `aaa new-model`, `line vty` local login configurations).

(For the full detailed description and command examples, please refer to the referenced web document)

naoya-oyama:
### Description
Fixes #3384
Fixes #3385

This PR completely fixes the management interface configuration and secure SSH access bootstrap logic for Cisco IOL nodes, specifically addressing behavior with Layer 2 (Switch) images and modern IOS-XE releases (17.16+).

### Core Fixes & Enhancements
1. **Dynamic L2 Port Mode Handling:** Added case-insensitive matching (`strings.Contains`) via `n.Cfg.Image`. If an L2 variant is detected, it cleanly executes `no switchport` on `Ethernet0/0` before assigning the management IP, preventing syntax rejections.
2. **Credential Synchronization:** Extracted custom `USERNAME` and `PASSWORD` definitions dynamically mapped out of the topology configuration environment context, perfectly aligning with other `vrnetlab`-based nodes.
3. **Modern Cipher Compliance:** Explicitly provisioned `ip ssh server algorithm hostkey rsa-sha2-256 rsa-sha2-512` into the setup chain. This resolves immediate user auth rejections introduced by legacy cipher strictness on IOS-XE 17.16+.
4. **Boot Wait Logic:** Introduced a robust `time.After(15 * time.Second)` delay combined with explicit sequential carriage returns (`\r\r\r\r`) to guarantee that the slow-booting IOL processing tables (such as internal VRF generation) are completely steady and available before injecting commands. This eliminates terminal input hijacking and accidental IP address duplication.

Verified working perfectly in my local lab environment!

### DIFF
--- a/nodes/iol/iol.go
+++ b/nodes/iol/iol.go
@@ -439,17 +439,48 @@ func (n *iol) CheckInterfaceName() error {
 }
 
 func (n *iol) UpdateMgmtIntf(ctx context.Context) error {
-	mgmt_str := fmt.Sprintf(
-		"\renable\rconfig terminal\rinterface Ethernet0/0\rip address %s %s\rno ipv6 address\ripv6 address %s/%d\rexit\rip route vrf clab-mgmt 0.0.0.0 0.0.0.0 Ethernet0/0 %s\ripv6 route vrf clab-mgmt ::/0 Ethernet0/0 %s\rend\rwr\r",
-		n.Cfg.MgmtIPv4Address,
-		clabutils.CIDRToDDN(n.Cfg.MgmtIPv4PrefixLength),
-		n.Cfg.MgmtIPv6Address,
-		n.Cfg.MgmtIPv6PrefixLength,
-		n.Cfg.MgmtIPv4Gateway,
-		n.Cfg.MgmtIPv6Gateway,
-	)
+        // L2 IOL images default to switchport mode, which rejects IP addresses.
+        // We inject "no switchport" if the image name contains "l2" (case-insensitive).
+        switchportCmd := ""
+        if strings.Contains(strings.ToLower(n.Cfg.Image), "l2") {
+                switchportCmd = "no switchport\r"
+        }
 
-	return n.Runtime.WriteToStdinNoWait(ctx, n.Cfg.ContainerID, []byte(mgmt_str))
+        // Align with other vrnetlab-based nodes by honoring topology env credentials,
+        // falling back to standard "admin/admin" if omitted.
+        username := "admin"
+        if u, ok := n.Cfg.Env["USERNAME"]; ok && u != "" {
+                username = u
+        }
+        password := "admin"
+        if p, ok := n.Cfg.Env["PASSWORD"]; ok && p != "" {
+                password = p
+        }
+
+        // IOS-XE (17.16+) boots slowly. We must wait for the VRF initialization
+        // to complete, otherwise IP configuration will be applied to the global table.
+        select {
+        case <-time.After(15 * time.Second):
+        case <-ctx.Done():
+                return ctx.Err()
+        }
+
+        // We prepend multiple carriage returns (\r\r\r\r) to explicitly wake up the console,
+        // clear any stale prompt states, and ensure it sits at a clean EXEC prompt before config entry.
+        mgmt_str := fmt.Sprintf(
+                "\r\r\r\renable\rconfig terminal\rinterface Ethernet0/0\r%sip address %s %s\rno ipv6 address\ripv6 address %s/%d\rexit\rusername %s privilege 15 password %s\rip ssh server algorithm hostkey rsa-sha2-256 rsa-sha2-512\rline vty 0 4\rlogin local\rtransport input ssh\rexit\rip route vrf clab-mgmt 0.0.0.0 0.0.0.0 Ethernet0/0 %s\ripv6 route vrf clab-mgmt ::/0 Ethernet0/0 %s\rend\rwr\r",
+                switchportCmd,
+                n.Cfg.MgmtIPv4Address,
+                clabutils.CIDRToDDN(n.Cfg.MgmtIPv4PrefixLength),
+                n.Cfg.MgmtIPv6Address,
+                n.Cfg.MgmtIPv6PrefixLength,
+                username,
+                password,
+                n.Cfg.MgmtIPv4Gateway,
+                n.Cfg.MgmtIPv6Gateway,
+        )
+
+        return n.Runtime.WriteToStdinNoWait(ctx, n.Cfg.ContainerID, []byte(mgmt_str))
 }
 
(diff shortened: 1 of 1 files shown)


hellt:
Hi @naoya-oyama 
in #3392 @steiler did attempt to fix the slow boot and premature console access. Would you mind given the current `main` a try? If that works for you, you can change your PR to only adds the bits related to the USR/PWD combo, ciphers, and L2 image detection

naoya-oyama:
Hi @helt
Understood. I will test it with the latest main branch. If it works fine, I'll update this PR as suggested.

naoya-oyama:
Hi @hellt 
Tested with the current main branch, and the boot/config logic works perfectly without any errors! 

I also verified that the SSH session connects and authenticates successfully (`%SEC_LOGIN-5-LOGIN_SUCCESS: Login Success`).

As you suggested, I will update this PR to remove the boot-related bits and keep it focused only on the USR/PWD combo, ciphers, and L2 image detection. I'll let you know once the PR is updated.

naoya-oyama:
Hi @hellt,

While testing this with the CML refplat IOS-XE L2 image, I noticed another edge case regarding the `startup-config`.

**The Problem:**
When a user specifies a custom `startup-config` in the topology, Containerlab still executes the SSH bootstrap logic (including `crypto key generate rsa`). After that, the `startup-config` is applied. This often overwrites or corrupts the freshly generated crypto key state (especially if the startup-config contains old/different key strings or lacks them entirely), causing subsequent SSH logins to fail with "Permission denied".

**Proposed Solution:**
I believe we should skip the automatic SSH key generation (or the entire SSH bootstrap logic) by Containerlab if a `startup-config` is explicitly provided in the topology config, assuming the user will handle the SSH/crypto configuration within their own startup script.

What do you think about this approach? If it sounds good, I can look into adding this logic as well.

hellt:
@naoya-oyama is this behaviour only appearing when the startup config itself contains some key related config knobs?

What I'd like to understand is what causes the corrupt ssh key config in the ssh server, because ideally I would want to have the keys added to the startup config without messing the config.
but if that is not possible, we may look into skipping the keys injection. I just wonder why this was never brought up before, hence I think maybe it is a specific startup config that interferes with the key injection?

naoya-oyama:
Hi @hellt,

Thank you for the quick feedback. Here is a brief summary regarding Cisco IOS/IOS-XE crypto key behavior with startup-config and proposed solutions (Option A: swap execution order, or Option B: skip key generation). Option A is recommended. Please let me know your preference!

kaelemc:
I am still a bit confused. The `UpdateMgmtIntf()` is supposed to run only on the subsequent boots. So *in theory* `no switchport` should already be in the config of an L2 IOL.

Furthermore. I think the SSH cipher config, is definitely warranted (and good) but it should not be in the `UpdateMgmtIntf()`, that should strictly just update the management IP since the method we use is quite weak (via stdio). We should add this cipher config into the bootstrap config template instead.

I also don't know about the 15 second wait, maybe we can use the log streaming function we have in the codebase to listen to the stdio (logs) and wait for the prompt, then start spamming the carriage returns to get the prompt and update the management interfaces.

Is it at possible you have booted an L3 IOL, changed the type and reused the same startup config? or you are using a startup config without `no switchport` ? as nobody reported this type of issue before it would be nice to collect more info so we can make a more targeted/scoped change.

naoya-oyama:
@kaelemc 
Sorry for the late reply, I have been very busy with work lately.

Regarding your points:

- SSH cipher config: I agree with you. It makes more sense to move this into the bootstrap config template rather than keeping it in UpdateMgmtIntf(). I will look into refactoring this.
- 15-second wait time: I think switching to the log streaming function to listen to the stdio and wait for the prompt is a great idea. I will explore this approach as well.

To clarify the switchport issue, I am not mixing up L3 and L2 IOL configs. In L2 IOL (IOS-XE), all interfaces operate as switchport and no shutdown by default on their very first boot. Since this is the native specification for L2 switches, the management interface can retain this default state. This is why explicitly handling it becomes necessary.

I will work on these updates and update the PR soon. Thank you for the valuable feedback!

naoya-oyama:
Closing this in favor of #3416 to reflect the refactored code based on feedback.

PR #3390, opened 2026-09-07
sonic-vs: mark the wire interfaces ARP-off and IPv6-off so only the port answers

roc-ops:
A `sonic-vs` node ends up with two kernel interfaces per link: the veth containerlab creates
(`ethN`) and the tap `syncd` derives from it - `eth1` becomes `Ethernet0`, `eth2` becomes
`Ethernet4`, and so on. SONiC configures the port; the `ethN` interface is only the wire that
carries the port's frames.

Left alone, the kernel treats that wire as an ordinary interface. It answers ARP for the port's
address with the wire's MAC, and it brings up an IPv6 link-local address there. A neighbour can
therefore cache the wrong MAC for the port, and can discover a device on the link that isn't the
port at all. Both interfaces see the same ARP request, so which reply the neighbour keeps is a
race - the entry is unreliable rather than reliably wrong, which makes it easy to miss.

SONiC's own virtual-switch test harness turns ARP off on these interfaces for the same reason:
[`sonic-swss/tests/conftest.py#L245-L246`](https://github.com/sonic-net/sonic-swss/blob/8b183d0cf1c1ebe3f1309099141676adbf9c63bf/tests/conftest.py#L245-L246),
class `VirtualServer` - *"disable arp, so no neigh on physical interfaces"*.

This change does the same for every data endpoint of the node (`eth0`, the management interface,
is skipped), running inside the container:

```
ip link set arp off dev <ifname>
sysctl -w net.ipv6.conf.<ifname>.disable_ipv6=1
```

A failure is logged as a warning and the deploy continues - this is a hardening fixup, not
something a lab should die on. `PostDeploy` applies it before the SONiC agents start, and
`PostDeployEndpoints` applies it again for links added to an already running node -
`clab tools veth create -t <topo>`, and `clab apply` where the node is not recreated
(`link-apply-mode: live` or `restart`), which is the path `deployLinks` uses. `Start` re-applies it
after `clab stop`/`clab start`, because moving a netdev between network namespaces resets its IPv6
sysctls. The commands are idempotent, so running them more than once is harmless.

Scope: this affects only the `sonic-vs` kind. `dell_sonic`, `plvision_sonic` and `sonic-vm` are
VM-based kinds and do not have this container/tap duality. The one way an existing user could
notice a difference: a lab that addresses the `ethN` interface directly instead of the
`Ethernet` port would lose IPv4 and IPv6 on that interface. The documented workflow for this kind
configures the port (`config interface ip add Ethernet0 ...`), which is unaffected.

The kind documentation gains a short note explaining the `ethN`/`Ethernet` relationship, and
there is a unit test for the command list.

Verified on a two-node `sonic-vs` lab (`s1:eth1 <-> s2:eth1`, `Ethernet0` addressed
`10.1.1.1/30` + `2001:db8::1/64` and `10.1.1.2/30` + `2001:db8::2/64`):

- **IPv4:** `eth1` shows `NOARP`; `ping -c3 10.1.1.2` is 0% loss and `ip neigh show 10.1.1.2`
  resolves to s2's `Ethernet0` MAC, not s2's `eth1` MAC.
- **IPv6:** `net.ipv6.conf.eth1.disable_ipv6 = 1` and `ip -6 addr show eth1` is empty, while
  `Ethernet0` keeps its own link-local; `ping6 -c3 2001:db8::2` is 0% loss and
  `ip -6 neigh show 2001:db8::2` resolves to s2's `Ethernet0` MAC.
- **Later-added link:** `clab tools veth create -t <topo> -a s1:eth2 -b s2:eth2` on the running
  lab leaves `eth2` with `NOARP` and `disable_ipv6=1` on both nodes.
- **Stop/start:** without the `Start` override, `clab stop` + `clab start` leaves `eth1` at
  `net.ipv6.conf.eth1.disable_ipv6 = 0` with a fresh `fe80::` address (NOARP survives); with it,
  `eth1` comes back `disable_ipv6 = 1`, `NOARP`, and `ip -6 addr show eth1` empty.
- `eth0` is untouched throughout: no `NOARP`, IPv6 still enabled.
- **Control on stock 0.78.2:** the same test resolves `10.1.1.2` to s2's `eth1` (wire) MAC on 2 of
  3 flush-and-ping rounds and to the port MAC on 1 of 3 - the race, observed.

### DIFF
--- a/docs/manual/kinds/sonic-vs.md
+++ b/docs/manual/kinds/sonic-vs.md
@@ -67,6 +67,10 @@ sonic-vs container uses the following mapping for its linux interfaces:
 
 When containerlab launches sonic-vs node, it will assign IPv4/6 address to the `eth0` interface. Data interface `eth1` mapped to `Ethernet0` port and needs to be configured with IP addressing manually. See Lab examples for exact configurations.
 
+/// note
+The `ethN` data interfaces are only the wire behind the SONiC ports (`eth1` behind `Ethernet0`, `eth2` behind `Ethernet4`, and so on); SONiC configures the port, while the `ethN` interface carries its frames. Containerlab therefore marks every `ethN` data interface ARP-off and IPv6-off once the links exist, so that only the port answers on the link and a neighbour cannot cache the wire's MAC.
+///
+
 ## Lab examples
 
 The following labs feature sonic-vs node:
--- a/nodes/sonic/sonic.go
+++ b/nodes/sonic/sonic.go
@@ -20,6 +20,13 @@ const (
 	generateable     = true
 	generateIfFormat = "eth%d"
 
+	// mgmtIfName is the management interface of a sonic-vs node. It is created
+	// by the runtime rather than by a link, so it is not an endpoint of the node
+	// today - but a node with `network-mode: none` may declare one, and the
+	// management interface is not a wire behind a SONiC port, so it is filtered
+	// out explicitly.
+	mgmtIfName = "eth0"
+
 	scrapliPlatformName = "sonic"
 )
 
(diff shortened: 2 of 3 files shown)


hellt:
Hi @roc-ops 

you write:

> A sonic-vs node ends up with two kernel interfaces per link: the veth containerlab creates
(ethN) and the tap syncd derives from it - eth1 becomes Ethernet0, eth2 becomes
Ethernet4, and so on. SONiC configures the port; the ethN interface is only the wire that
carries the port's frames.

Would you have the lab deployed where this is evident? I though sonic renames the interfaces `ethN` to `ethernetN`. But I haven't ran any labs recently with sonic

roc-ops:
Yes — I have it up right now, stock `docker-sonic-vs:latest`, two nodes, one link:

```yaml
name: sonicarp
topology:
  nodes:
    s1: { kind: sonic-vs, image: docker-sonic-vs:latest }
    s2: { kind: sonic-vs, image: docker-sonic-vs:latest }
  links:
    - endpoints: ["s1:eth1", "s2:eth1"]
```

`Ethernet0` is not `eth1` renamed. `start.sh` reads the `ethN` netdevs at boot to build `lanemap.ini`, and syncd's VS SAI then creates a **tap** for each mapped port. Both netdevs stay, with different types and different MACs (clab 0.78.2, unpatched, `s2`):

```
$ docker exec clab-sonicarp-s2 cat /usr/share/sonic/hwsku/lanemap.ini
eth1:25,26,27,28

$ docker exec clab-sonicarp-s2 ip -d link show eth1
6365: eth1@if6364: <BROADCAST,MULTICAST,PROMISC,UP,LOWER_UP> mtu 9122 ...
    link/ether aa:c1:ab:bc:40:cc brd ff:ff:ff:ff:ff:ff link-netnsid 1 ...
    veth addrgenmode eui64 ...

$ docker exec clab-sonicarp-s2 ip -d link show Ethernet0
6: Ethernet0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9100 ...
    link/ether ca:ba:5b:a2:66:37 brd ff:ff:ff:ff:ff:ff ...
    tun type tap pi off vnet_hdr off ...
```

veth `aa:c1:ab:bc:40:cc` and tap `ca:ba:5b:a2:66:37`, one link.

That is what this PR is about: with `10.0.0.2/30` configured on `s2`'s `Ethernet0`, both of them answer for it. From `s1`, unpatched:

```
$ docker exec clab-sonicarp-s1 arping -c 4 -I Ethernet0 10.0.0.2
42 bytes from aa:c1:ab:bc:40:cc (10.0.0.2): index=0 time=156.561 usec
42 bytes from ca:ba:5b:a2:66:37 (10.0.0.2): index=1 time=239.858 usec
42 bytes from aa:c1:ab:bc:40:cc (10.0.0.2): index=2 time=93.510 usec
42 bytes from ca:ba:5b:a2:66:37 (10.0.0.2): index=3 time=141.063 usec
...
4 packets transmitted, 8 packets received,   0% unanswered (4 extra)

$ docker exec clab-sonicarp-s1 ping6 -c 3 -I Ethernet0 ff02::1
64 bytes from fe80::34bf:2ff:fec4:acb%Ethernet0: icmp_seq=1 ttl=64 time=0.041 ms   <- s1's own port
64 bytes from fe80::a8c1:abff:febc:40cc%Ethernet0: icmp_seq=1 ttl=64 time=0.356 ms <- s2's wire (aa:c1:ab:bc:40:cc)
64 bytes from fe80::c8ba:5bff:fea2:6637%Ethernet0: icmp_seq=1 ttl=64 time=0.431 ms <- s2's port (ca:ba:5b:a2:66:37)
```

Whichever ARP reply lands last wins, so the neighbour can end up sending the port's traffic to the wire's MAC; and a link with two nodes on it looks like a link with three.

Same lab, same commands, on a build of this PR (MACs differ, it is a fresh deploy — wire `aa:c1:ab:7d:3e:1c`, port `7a:73:b4:bd:9d:c3`):

```
$ docker exec clab-sonicarp-s2 ip -br link | grep -E '^(eth[0-9]|Ethernet0 )'
eth0@if6368   UP        7a:73:b4:bd:9d:c3 <BROADCAST,MULTICAST,UP,LOWER_UP>
Ethernet0     UNKNOWN   7a:73:b4:bd:9d:c3 <BROADCAST,MULTICAST,UP,LOWER_UP>
eth1@if6370   UP        aa:c1:ab:7d:3e:1c <BROADCAST,MULTICAST,NOARP,PROMISC,UP,LOWER_UP>

$ docker exec clab-sonicarp-s1 arping -c 4 -I Ethernet0 10.0.0.2
42 bytes from 7a:73:b4:bd:9d:c3 (10.0.0.2): index=0 time=317.508 usec
42 bytes from 7a:73:b4:bd:9d:c3 (10.0.0.2): index=1 time=181.259 usec
42 bytes from 7a:73:b4:bd:9d:c3 (10.0.0.2): index=2 time=158.079 usec
42 bytes from 7a:73:b4:bd:9d:c3 (10.0.0.2): index=3 time=153.434 usec
4 packets transmitted, 4 packets received,   0% unanswered (0 extra)

$ docker exec clab-sonicarp-s1 ping6 -c 3 -I Ethernet0 ff02::1
64 bytes from fe80::fc47:19ff:fea5:d132%Ethernet0: icmp_seq=1 ttl=64 time=0.043 ms   <- s1's own port
64 bytes from fe80::7873:b4ff:febd:9dc3%Ethernet0: icmp_seq=1 ttl=64 time=0.432 ms   <- s2's port
```

`eth1` keeps its name and still carries the frames — it is just `NOARP` and IPv6-off, so only the port answers.

I have also pushed a commit that says this in the kind docs and the code comment, since "eth1 becomes Ethernet0" was the phrasing that read as a rename.

PR #3399, opened 2026-09-12
MACVLAN mgmt-net driver and clab manged IPAM

kaelemc:
Continuation of #2732


.github/workflows/smoke-tests.yml
```diff
@@ -103,6 +105,14 @@ jobs:
 
       - name: Run smoke tests
         run: |
+          if [[ '${{ matrix.test-suite }}' == '32*.robot' ]]; then
```
hellt: lets add a comment here why this is needed

kaelemc: ah infact that can go. was used in debugging since the ARP was not behaving as expected in the runner.

Issue #3401, opened 2026-09-13
cisco_iol: mgmt interface statically set and cannot be changed


ccieninja:
## Description

Every IOL node's `Ethernet0/0` is claimed for management and that interface cannot be used in topology. This is a vital interface used commonly in academic settings and labs making this a very inconvenient placement for management. 

https://github.com/srl-labs/containerlab/blob/d1d767e526a058eb35326551b2663233f5a65f02/nodes/iol/iol.cfg.tmpl#L21-L30

Even if the startup configuration is changed, the interface is still bound to the container networking behind the scenes and cannot be used.

Turning off the networking with `network-mode: none` removes the ability to pass a startup config entirely. 

## Proposed Solution

Introduce an ENV var to allow users to define what port they wish to bind for management.

```yaml
topology:
  kinds:
    cisco_iol:
      env:
        CLAB_IOL_MGMT_INTF: Ethernet3/3
```

Happy to draft PR. Will submit one shortly hopefully get a fix started.

PR #3404, opened 2026-09-13
cisco_iol: allow replacing or disabling the bootstrap config via CLAB_IOL_BOOTSTRAP_CONFIG

### LINKED ISSUE #3403: cisco_iol: bootstrap configuration statically set and cannot be changed or disabled
## Description

Every IOL node boots with containerlab's built-in bootstrap configuration that sets up management networking and SSH:

https://github.com/srl-labs/containerlab/blob/d1d767e526a058eb35326551b2663233f5a65f02/nodes/iol/iol.cfg.tmpl#L1-L54

This baseline is rather static and does not allow much control over startup state. A full startup-config can replace it, but the user would have to give up partial layering entirely: there is no way to define a custom baseline for all IOL nodes of a topology and keep small per-node partials (this would be great to have). Turning off the networking with `network-mode: none` removes the ability to pass a startup config entirely.

## Proposed Solution

Introduce an environment variable to allow replacing or disabling the bootstrap configuration:

```yaml
topology:
  kinds:
    cisco_iol:
      env:
        CLAB_IOL_BOOTSTRAP_CONFIG: my-baseline.cfg.tmpl   # replaces the built-in bootstrap
  nodes:
    r1:
      startup-config: r1.partial.cfg                 # layered on my-baseline.cfg
    r2:
      env:
        CLAB_IOL_BOOTSTRAP_CONFIG: none              # no bootstrap: factory-default boot
```

Happy to draft PR. Hoping to submit one shortly to advance a fix.

ccieninja:
## Description

Fixes #3403

**BLUF**: Complementary PR for #3401. Introduce ability to replace or disable the default bootstrap configuration via `CLAB_IOL_BOOTSTRAP_CONFIG` environment variable. Also includes a fix for template rendering for partial startup-configs (currently not working).

Containerlab ships a bootstrap config for IOL that sets up networking and SSH. 
https://github.com/srl-labs/containerlab/blob/d1d767e526a058eb35326551b2663233f5a65f02/nodes/iol/iol.cfg.tmpl#L22-L30

This configuration is rather static and not very customizable. In its current design, it is always included when providing partial configs and is inconvenient to manage or control without disabling other desired features like the ability to provide a startup config altogether (setting `network-mode: none`). 

This solution introduces an optional `CLAB_IOL_BOOTSTRAP_CONFIG` environment variable which accepts a replacement Go template via `file-path` or can be set to `none` which disables the bootstrap so only a user startup-config (if any) is applied. Leaving this variable unset/empty will continue to operate as it does today making this a non-intrusive change to the workflow.

Included is also a fix for template rendering in partial configs (which is currently non functional). Templates in partial startup-configs are now also rendered correctly, so template variables work in partials the way they already do in full startup-configs.

This also includes an update to the documentation for IOL startup configuration.

## Usage 
```yaml
---
topology:
  kinds:
    cisco_iol:
      env:
        CLAB_IOL_BOOTSTRAP_CONFIG: my-baseline.cfg.tmpl   # replaces the built-in bootstrap
```

## Validation
Co-developed and cross-validated with AI. Confirmed operational with the following test files

### Files

**Topology**
```yaml
---
name: bootstrap-test
topology:
  defaults:
    kind: cisco_iol
  kinds:
    cisco_iol:
      image: vrnetlab/cisco_iol:17.12.01
      env:
        CLAB_IOL_BOOTSTRAP_CONFIG: my-baseline.cfg.tmpl   # replaces the built-in bootstrap
  nodes:
    r1:
      startup-config: r1.partial.cfg                 # layered on my-baseline.cfg.tmpl
    r2:
      env:
        CLAB_IOL_BOOTSTRAP_CONFIG: none              # no bootstrap: factory-default boot
```

**my-baseline.cfg.tmpl**
```
hostname {{ .Hostname }}
ip domain name customlab
no ip domain lookup
username admin privilege 15 secret admin
vrf definition clab-mgmt
 address-family ipv4
 !
!
interface Ethernet0/0
 vrf forwarding clab-mgmt
 ip address {{ .MgmtIPv4Addr }} {{ .MgmtIPv4SubnetMask }}
 no shutdown
!
ip route vrf clab-mgmt 0.0.0.0 0.0.0.0 Ethernet0/0 {{ .MgmtIPv4GW }}
ip ssh version 2
crypto key generate rsa modulus 2048
!
line vty 0 4
 login local
 transport input ssh
!
{{ .PartialCfg }}
end
```

**r1.partial.cfg**
```
interface Loopback10
 ip address 10.99.1.1 255.255.255.255
snmp-server location {{ .Hostname }}-loc
```

### Results

**R1**
```
r1#show running-config | section interface Ethernet0/0|interface Loopback10|line vty|snmp-server
interface Loopback10
 ip address 10.99.1.1 255.255.255.255
interface Ethernet0/0
 vrf forwarding clab-mgmt
 ip address 172.20.20.10 255.255.255.0
snmp-server location r1-loc
line vty 0 4
 login local
 transport input ssh
```

**R2**
```
Would you like to enter the initial configuration dialog? [yes/no]:
```

Unit tests are on a separate branch, left out to keep the change minimal: https://github.com/ccieninja/containerlab/commit/e206fd7be


kaelemc:
I think the [full startup config](https://containerlab.dev/manual/kinds/cisco_iol/#full-startup-configuration ) is the one to use here.

ccieninja:
The full startup configuration is certainly possible but its an all or nothing configuration today. Enabling control of the bootstrap config with this change allows customization or omission of the static containerlab config that sets up management **_enabling customized management while retaining the use of partials with it._** It would be the start of cool templating dynamic where a `config-for-all` can be layered with a `config-for-this-node`. The `none` provides nodes with the override ability to opt-out of `config-for-all`.

Customizable management would be a powerful feature to have across the board for all nodes. I would like to start paving the path to mature the feature-set. This is a feature that would make my life a whole lot easier.

### Example use case for me

Here is a practical example where this would be great for me. I am studying for the CCIE and the several different high level courses provide labs and configs to set up the environment to practice the topics. Porting these configs over into container lab may involve renaming/shifting an interface (because eth0/0 is unusable), but other than that, its good go. 

Now because this is containerlab, I am imagining taking these labs to a new level by plugging in suziq or grafana/syslog stack or ansible/terraform because it comes supported. Now I need to customize the management access for these static course configs I have to wire things up. 

I could:

1. Go through another round of edits for all the course configs and further diminish their original state
2. Utilize layering provided with this change to keep configs clean and add a customized management bootstrap for all. 

### Future thoughts and use for others

Like myself I'd like to think that others out there are switching to containerlab because it represents the future of labbing. Staff running labs at universities, professors teaching a course, or specialized enthusiasts integrating devops. In all the uses cases I just mentioned, having an improved templating system that facilitates porting configs over is a huge win.

PR #3411, opened 2026-09-17
Update vrnetlab.md - UUID info

sdargoeuves:
add the information about the UUID variable. Related to https://github.com/srl-labs/vrnetlab/pull/529


hellt:
@sdargoeuves I've left a comment in the vrnetlab counterpart. When you confirm it is good on your end I will merge both

